PR #8, opened 2014-11-12
Ensure we don't override a group membership.

iamcarrico:
Reset the keys on the display, so that "- Private group -" doesn't override a value in the array.

Currently, since private groups are pulled out, the key for the elements will not be in order, so the count is giving a number within the array. This will override any groups that should be displayed with the "- Private group -" text.

### DIFF
--- a/og_ui/og_ui.module
+++ b/og_ui/og_ui.module
@@ -763,9 +763,8 @@ function og_ui_field_formatter_view($entity_type, $entity, $field, $instance, $l
         }
       }
       if ($private) {
-        $count = count($element);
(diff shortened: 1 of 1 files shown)


amitaibu:
Hi! Any chance for a simpleTest to prevent future regressions? :)

amitaibu:
@iamcarrico or even without a test -- can you describe the steps to reproduce the problem on a clean installation.

iamcarrico:
I can... and will try and write a test... once my demo is over tomorrow.

amitaibu:
@iamcarrico is it still in your radar? :)

amitaibu:
Merged :)

PR #10, opened 2014-11-16
Adding handling of group privacy change using batch API.

shushu:


og.module
```diff
@@ -2353,6 +2396,54 @@ function og_get_entity_groups($entity_type = 'user', $entity = NULL, $states = a
...
+ *   The group object or group ID.
+ * @param $group_type
+ *   (optional) The group type. If empty, assumed 'node'.
+ *
+ * @return
+ *  An array with the group content's entity type as the key, and array - keyed by
+ *  the OG membership ID and the entity ID as the value. If nothing found,
+ *  then an empty array.
+ */
+function c4m_og_get_group_content($group, $group_type = 'node') {
```
amitaibu: you probably won't need it as is -- you'll just need an EFQ to get all nodes with `->range(0, 50)`. Also, function name seems wrong :P

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+ * @param $entity_type
+ *   The group type.
+ */
+function og_handle_group_privacy_change($entity, $entity_type) {
+  if (!og_is_group($entity_type, $entity)) {
+    return;
+  }
+
+  // Check whether group privacy change batch processing is needed.
+  if (!variable_get(OG_PRIVACY_CHANGE_BATCH_PROCESSING, TRUE)) {
```
amitaibu: move as the first check.

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+    return;
+  }
+  // Check for privacy change.
+  $original_entity = $entity->original;
+  $wrapper = entity_metadata_wrapper($entity_type, $entity);
+  $original_wrapper = entity_metadata_wrapper($entity_type, $original_entity);
+
+  $context = array(
+    'change_detected' => FALSE,
+    'updated_group' => $wrapper,
```
amitaibu: pass the entity itself ($wrapper->value())

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+  );
+  batch_set($batch);
+
+  // Since batch_set() is called not from a submit handler, batch_process()
+  // should be called.
+  batch_process();
+}
+
+/**
+ * Implements hook_og_update_group_content_privacy_alter().
```
amitaibu: add line break.

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+    $max = $query
+      ->entityCondition('entity_type', 'og_membership')
+      ->propertyCondition('group_type', $entity_type, '=')
+      ->propertyCondition('entity_type', 'node', '=')
+      ->propertyCondition('gid', $entity_id, '=')
+      ->count()->execute();
+
+    $context['sandbox']['progress'] = 0;
+    $context['sandbox']['last_id'] = 0;
+    $context['sandbox']['max'] = $max;
```
amitaibu: max => total

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+  $orig_og_access = $context['original_group']->{OG_ACCESS_FIELD}->value();
+  if ($og_access !== $orig_og_access) {
+    $context['change_detected'] = TRUE;
+  }
+}
+
+/**
+ * Batch API group content privacy change callback.
+ */
+function og_update_group_content_privacy($entity_type, $entity_id, &$context) {
```
amitaibu: $entity_type => $group_type
$entity_id => $group_id

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+    ->propertyCondition('group_type', $entity_type, '=')
+    ->propertyCondition('entity_type', 'node', '=')
+    ->range(0, $limit)
+    ->propertyCondition('etid', $context['sandbox']['last_id'], '>')
+    ->propertyCondition('gid', $entity_id, '=')
+    ->execute();
+
+  if (isset($result['og_membership'])) {
+    foreach (array_keys($result['og_membership']) as $id) {
+      if (!$og_membership = entity_load_single('og_membership', $id)) {
```
amitaibu: why would og_membership not load?

og.module
```diff
@@ -894,6 +900,118 @@ function og_entity_update($entity, $entity_type) {
...
+  $result = $query
+    ->entityCondition('entity_type', 'og_membership')
+    ->propertyCondition('group_type', $entity_type, '=')
+    ->propertyCondition('entity_type', 'node', '=')
+    ->range(0, $limit)
+    ->propertyCondition('etid', $context['sandbox']['last_id'], '>')
+    ->propertyCondition('gid', $entity_id, '=')
+    ->execute();
+
+  if (isset($result['og_membership'])) {
```
amitaibu: Return early:

``` php
if (empty($result['og_membership'])) {
  $context['finished'] = 1;
  return;
}
```

og.module
```diff
@@ -894,6 +900,121 @@ function og_entity_update($entity, $entity_type) {
...
+  $orig_og_access = $original_wrapper->{OG_ACCESS_FIELD}->value();
+  if ($og_access !== $orig_og_access) {
+    $context['change_detected'] = TRUE;
+  }
+}
+
+/**
+ * Batch API group content privacy change callback.
+ */
+function og_update_group_content_privacy($group_type, $group_id, &$context) {
```
amitaibu: Maybe better `og_invoke_node_access_acquire_grants`?

shushu:
selection_394

shushu:
Tests fails on Travis, but I can't find a reason. Output seems similar to earlier OG travis executions. Needs to guess which code makes the failure.

og.module
```diff
@@ -75,6 +75,12 @@ define('OG_MEMBERSHIP_TYPE_DEFAULT', 'og_membership_type_default');
 define('OG_MEMBERSHIP_REQUEST_FIELD', 'og_membership_request');
 
 /**
+ * The default variable name that controls abtch processing on
```
amitaibu: All this code should move under `og_access`

og_access/og_access.module
```diff
@@ -255,4 +255,21 @@ function _og_access_verify_access_field_existence($node) {
...
\ No newline at end of file
+}
+
+/**
+ * Implements hook_og_invoke_node_access_acquire_grants_alter().
+ *
+ * Simple check whether OG_ACCESS_FIELD values was changed.
+ */
+function og_access_og_invoke_node_access_acquire_grants_alter(&$context) {
+  return;
```
amitaibu: looks wrong :)

og.module
```diff
@@ -894,6 +900,105 @@ function og_entity_update($entity, $entity_type) {
...
+    ->entityCondition('entity_type', 'og_membership')
+    ->propertyCondition('group_type', $group_type, '=')
+    ->propertyCondition('entity_type', 'node', '=')
+    ->range(0, $limit)
+    ->propertyCondition('etid', $context['sandbox']['last_id'], '>')
+    ->propertyCondition('gid', $group_id, '=')
+    ->execute();
+
+  if (!isset($result['og_membership'])) {
+    $context['finished'] = 1;
```
amitaibu: Add comment.

shushu:
Update: failure seems to happen in "OG UI upgrade path". This is what I get running locally.
selection_395

amitaibu:
@shushu try to run in on a completely clean installation without any other non-related/dev modules present. Simpletest is sometimes delicate..

shushu:
Clean Drupal 7 + OG from this branch doesn't seems to reproduce the same problem. I try to execute all relevant OG tests to try and reproduce the Travis failures.

amitaibu:
Don't be discouraged -- those tests are hard to get right... :)

shushu:
I think I got it :-)
It seems that simpletest can't bare batch API calls during its execution.
From what I understand, the problem is with the batch_process() call, which is done by us because we don't create the batch during a submit handler.

shushu:
As I see it, calling `og_handle_group_privacy_change` should not be executed in case the entity is created/updated programatically. I think it is a decent limitation, and it will solve the test failures and any other failures that can happen from other places.
The only question is how we can check whether the `og_entity_update` is triggered from a programmatic call or from form submission.

amitaibu:
Triggering `og_handle_group_privacy_change` should happen on the `form_submit()` instead of update hooks.

amitaibu:
@shushu is this ready for review?

amitaibu:
@shushu https://github.com/Gizra/og/pull/10/files#r20425215

shushu:
@amitaibu I think it is now. Tests seems to pass both in og and in c4m that uses this branch.

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+  }
+}
+
+/**
+ * Implements hook_field_attach_form().
+ */
+function og_access_field_attach_form($entity_type, $entity, &$form, &$form_state, $langcode) {
+  list(,, $bundle) = entity_extract_ids($entity_type, $entity);
+
+  if ($entity_type == 'user' || !og_is_group_type($entity_type, $bundle)) {
```
amitaibu: is the `$entity_type == 'user'` really needed?

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+    return;
+  }
+  $form['#submit'][] = 'og_access_handle_group_privacy_change';
+}
+
+/**
+ * Submit handler; In case of group privacy change, create batch operation to update group content.
+ *
+ * When group privacy (OS_ACCESS_FIELD) is changed, its group content should be
```
amitaibu: OS_ACCESS_FIELD => (e.g. OG_ACCESS_FIELD)

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+ * changed as well. Since a group might have big amount of content, it is being
+ * handled using Batch API.
+ *
+ * @param $form
```
amitaibu: remove

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+    return;
+  }
+
+  // Check for privacy change.
+  $context = array(
+    'change_detected' => FALSE,
+    'entity' => $entity,
+    'entity_type' => $entity_type,
+  );
+  drupal_alter('og_access_invoke_node_access_acquire_grants', $context);
```
amitaibu, line 350: this should be documented in `og_access.api.php`

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+  if (!$context['change_detected']) {
+    return;
```
amitaibu, line 353: Add comment why return early.

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+  $wrapper = entity_metadata_wrapper($entity_type, $entity);
+  $batch = array(
+    'title' => t('Handle group privacy change'),
+    'operations' => array(
+      array('og_access_invoke_node_access_acquire_grants', array($entity_type, $wrapper->getIdentifier())),
+    ),
+  );
+  batch_set($batch);
+
+  // Since batch_set() is called not from a submit handler, batch_process()
```
amitaibu: is this still the case?

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+  // should be called.
+  batch_process();
+}
+
+/**
+ * Batch API group content privacy change callback.
+ */
+function og_access_invoke_node_access_acquire_grants($group_type, $group_id, &$context) {
```
amitaibu, line 376: missing `@param`

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function og_access_form_og_ui_admin_settings_alter(&$form, &$form_state) {
+  $form[OG_ACCESS_PRIVACY_CHANGE_BATCH_PROCESSING] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Update group content privacy'),
+    '#description' => t("Upon group privacy change, create batch operation to update group content."),
```
amitaibu: use `'` instead of `"`

shushu:
It seems that triggering og_handle_group_privacy_change in form_submit() cause the entity not to have the `original` value, which is needed for comparing.
This was found only in manual re-testing, and clarifies (for me, again) how limit is the automatic tests even when you invest a lot in them.
I try to both catch it on an automatic test to avoid future degradation, and to find a proper solution (calling `entity_load_unchanged` during the hook does not solve this).

og_access/og_access.module
```diff
@@ -255,4 +260,142 @@ function _og_access_verify_access_field_existence($node) {
...
+/**
+ * Implements hook_field_attach_form().
+ */
+function og_access_field_attach_form($entity_type, $entity, &$form, &$form_state, $langcode) {
+  list(,, $bundle) = entity_extract_ids($entity_type, $entity);
+
+  if ($entity_type == 'user' || !og_is_group_type($entity_type, $bundle)) {
+    return;
+  }
+  $form['#submit'][] = 'og_access_handle_group_privacy_change';
```
amitaibu: move this to after_build

``` php
$form['#after_build'][] = 'og_access...after_build';`

function og_access...after_build($form) {
  $form['#submit'][] = 'og_access_handle_group_privacy_change';
  return $form;
}
```

shushu: Moving to `#after_build` does not help, since `$entity->original` is created during `node_save()`.
I think the way to handle this is as you suggested before - leave some temporary attribute in the entity during the submit handler, and grab it during `hook_entity_update`.

shushu:
This patch effects only groups that are created using forms. Current OG tests done programatically, hence we fail to test this feature.

og_access/og_access.module
```diff
@@ -255,4 +260,182 @@ function _og_access_verify_access_field_existence($node) {
...
+    return;
+  }
+  $form['#after_build'][] = 'og_access_handle_group_privacy_change_after_build';
+}
+
+/**
+ * After_build handler. Adds a submit handler, verifying it will be called last.
+ */
+function og_access_handle_group_privacy_change_after_build($form, &$form_state) {
+  $form['#submit'][] = 'og_access_determine_group_privacy_change';
```
amitaibu, line 290: do we still need this?

og_access/og_access.module
```diff
@@ -255,4 +260,182 @@ function _og_access_verify_access_field_existence($node) {
...
+  $entity->_og_access_trigger_access_change = TRUE;
+}
+
+/**
+ * Implements hook_entity_update().
+ * In case the group entity was marked with group privacy change,
+ * create batch operation to update group content.
+ */
+function og_access_entity_update($entity, $entity_type) {
+  if (!isset($entity->_og_access_trigger_access_change) || $entity->_og_access_trigger_access_change == FALSE) {
```
amitaibu: probably enough just to do `!empty($entity->_og_access_trigger_access_change)`

og_access/og_access.module
```diff
@@ -255,4 +260,174 @@ function _og_access_verify_access_field_existence($node) {
...
+    $node = node_load($og_membership->etid);
+    node_access_acquire_grants($node);
+    $context['sandbox']['progress']++;
+    $context['sandbox']['last_id'] = $id;
+  }
+  $context['finished'] = $context['sandbox']['progress'] / $context['sandbox']['total'];
+}
+
+/**
+ * Implements hook_form_alter().
```
amitaibu: => `Implements hook_form_FORM_ID_alter().`

og_access/og_access.module
```diff
@@ -255,4 +260,174 @@ function _og_access_verify_access_field_existence($node) {
...
+    ->propertyCondition('gid', $group_id, '=')
+    ->execute();
+
+  // Mark "finished" if there are no more results.
+  if (!isset($result['og_membership'])) {
+    $context['finished'] = 1;
+    return;
+  }
+
+  foreach (array_keys($result['og_membership']) as $id) {
```
amitaibu: `foreach (entity_load('og_membership', array_keys($result['og_membership'])) as $og_membership)`

og_access/og_access.module
```diff
@@ -255,4 +260,174 @@ function _og_access_verify_access_field_existence($node) {
...
+    $og_membership = entity_load_single('og_membership', $id);
+
+    // Load the node group content and "refresh" its grants.
+    $node = node_load($og_membership->etid);
+    node_access_acquire_grants($node);
```
amitaibu, line 416: `// Rebuild the permissions for the node.`

amitaibu:
PR looks good, I've added minor comments

amitaibu:
Merged, thanks.

PR #18, opened 2014-12-13
Fixes #17: og_field_access making the OgMigrateMembership update fail

### LINKED ISSUE #17: of_field_access makes OgMigrateMembership fail
### Problem/Motivation

During a migration of OG Memberships from 7.x-1.x to 7.x-2.x, the process tries to determine whether the user has access to all potential group audience fields. That calls the `og_field_access_field_access()` function which calls `og_user_access_entity()` which tries to load all the groups related to the given entity. The problem is that most of these groups are not fully migrated yet so their `group_type` is empty so when OG tries to create an `entity_metadata_wrapper('', $gid)`, an Exception is fired.
### Proposed resolution

Use the last `og_get_best_group_audience_field()` parameter to avoid that `field_access` test during the migration process.

DuaelFr:
See #17


amitaibu:
Merci :)

PR #22, opened 2015-01-09
Fix og_field_access 1.x to 2.x migration

DuaelFr:
During an upgrade from OG 1.x to 2.x if you used to use OG Field Access submodule, your custom permissions are lost because OgMigrateRoles::preImport() is run while the new audience fields are not set on the group content bundles yet.
This change just declare OgMigrateMembership as dependency of OgMigrateRoles to ensure that these fields have been added to the bundles before we try to check them.
As OgMigrateMembership does not rely on OG user roles or permissions to run, that fix might not introduce any side effect.

### DIFF
--- a/includes/migrate/7200/og_roles.migrate.inc
+++ b/includes/migrate/7200/og_roles.migrate.inc
@@ -13,6 +13,8 @@ class OgMigrateRoles extends OgEntityMigration {
 
   public $tableName = 'og_role';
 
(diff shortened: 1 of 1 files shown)


DuaelFr:
This issue has been cross-posted to drupal.org
https://www.drupal.org/node/2404701

amitaibu:
Thanks :)

PR #9, opened 2014-11-13
Check if user has access on returned group from og_context()

amitaibu:

PR #23, opened 2015-01-13
Check if user has access on returned group from og_context()

HelenaEksler:
#9


HelenaEksler:
@amitaibu Made changes in other PR, because I don't have permissions to push into Gizra/og.

PR #25, opened 2015-01-15
Do not hardcode field name.

HelenaEksler:
#24


HelenaEksler:
@amitaibu Please, check this.

amitaibu:
OG core shouldn't "know" about pluggable_node_access module. It's pluggable_node_access that should implement OG's hooks if needed.

PR #26, opened 2015-01-15
Do not check twice if group access has changed

HelenaEksler:
We don't need to check whether privacy was changed twice. We already checking it using drupal_alter. Just need to check if this is actually a group and that batch processing is needed.

### DIFF
--- a/og_access/og_access.api.php
+++ b/og_access/og_access.api.php
@@ -58,7 +58,7 @@ function hook_og_access_invoke_node_access_acquire_grants_alter(&$result, $conte
   $og_access = $wrapper->{OG_ACCESS_FIELD}->value();
   $original_og_access = $original_wrapper->{OG_ACCESS_FIELD}->value();
   if ($og_access !== $original_og_access) {
-    $result[] = TRUE;
+    $result['my_module'] = TRUE;
   }
 }
 
--- a/og_access/og_access.api.php
+++ b/og_access/og_access.api.php
@@ -14,23 +14,23 @@
 /**
  * Allow modules to mark group privacy change.
  *
- * @param &$context
- *   Array including 3 keys:
- *   'entity' - the submitted group entity.
- *   'entity_type' - String. The type of the entity that was submitted.
- *   'change_detected' - Boolean. Marks whether the group privacy changed or not.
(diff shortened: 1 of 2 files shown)


HelenaEksler:
@amitaibu 
I think, this is good solution. Check, please.

og_access/og_access.api.php
```diff
@@ -14,23 +14,23 @@
...
- *   'change_detected' - Boolean. Marks whether the group privacy changed or not.
- *                       FALSE by default, and changeable be each hook
- *                       implementation.
+ * @param array $context
+ *   Array with the following keys:
+ *   'entity': the group entity.
+ *   'entity_type':  The group type.
  */
-function hook_og_access_invoke_node_access_acquire_grants_alter(&$context) {
+function hook_og_access_invoke_node_access_acquire_grants($context) {
```
amitaibu, line 25: you need to also document the `hook_og_access_invoke_node_access_acquire_grants_alter(&$result, $context)`

HelenaEksler:
@amitaibu PR is ready.

og_access/og_access.api.php
```diff
@@ -14,22 +14,51 @@
...
+ *
+ * @param $result
+ *   Array of results of the hooks called before.
+ * @param array $context
+ *   Array with the following keys:
+ *   'entity': the group entity.
+ *   'entity_type':  The group type.
+ */
+function hook_og_access_invoke_node_access_acquire_grants_alter(&$result, $context) {
+  $wrapper = entity_metadata_wrapper($context['entity_type'], $context['entity']);
```
amitaibu, line 50: the alter is mostly for changing existing results. Also i think the `$result` array is keyed by the module name the provided the response., so the example could probably be:

``` php
$result['my_module'] = FALSE;
```

HelenaEksler, line 50: I think example should be:

``` php
if ($access !== $original_access) {
  $result['my_module'] = TRUE; 
}
```

Because after altering all functions we are checking that we have in results at least one `TRUE`, that means that access has changed and we go on.

amitaibu, line 50: ok.

amitaibu:
Merged, thanks.

PR #29, opened 2015-01-28
Do not load menu item just before determining context handler to prevent...

jummonk:
... premature access denial due to empty og context.


amitaibu:
Thanks!

amitaibu:
Sorry, this PR was reverted -- see more info in #35

amitaibu:
Here's a solution to using OG context from within an access callback - https://github.com/Gizra/og_context_access_callback

PR #31, opened 2015-02-09
Fix checking whether user has access to the group in og_context()

HelenaEksler:
1. Check if we already have context only if we don't have to check user's permissions to access the current group.
2. Check whether user has access to the group if it's necessary.

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -215,9 +215,8 @@ function og_context($group_type = 'node', $group = NULL, $account = NULL, $check
 
   $context = &drupal_static(__FUNCTION__, array());
 
-  if (empty($context[$cache_key])) {
+  if (!array_key_exists($cache_key, $context)) {
     // Mark that this context hasn't been checked yet.
-
     $context[$cache_key] = FALSE;
   }
 
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -199,19 +199,21 @@ function og_context($group_type = 'node', $group = NULL, $account = NULL, $check
   global $user;
   // User account is sent.
   $account = $account ? $account : user_load($user->uid);
+  if (!$check_access) {
+    // Check if we already have context only if we don't have to check user's
+    // permissions to access the current group.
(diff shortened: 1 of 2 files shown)


og_context/og_context.module
```diff
@@ -200,24 +200,43 @@ function og_context($group_type = 'node', $group = NULL, $account = NULL, $check
...
+    $group_type,
+    $id,
+    $account->uid,
+    $check_access,
+  );
+  $cache_key = implode(':', $cache_keys);
+
+  $context = &drupal_static(__FUNCTION__, array());
+
+  if (empty($context[$cache_key])) {
```
amitaibu: this looks wrong - if we called `og_context()` and it was empty, then it will enter here again in the second call.

amitaibu:
Merged, thanks

og_context/og_context.module
```diff
@@ -199,27 +199,44 @@ function og_context($group_type = 'node', $group = NULL, $account = NULL, $check
...
-        return $context;
-      }
+  $id = FALSE;
+  if ($group) {
+    list($id) = entity_extract_ids($group_type, $group);
+  }
+
+  $cache_keys = array(
+    $group_type,
+    $id,
```
pianomansam, line 210: This right here is the cause of not being able to retrieve the [current active groups](https://www.drupal.org/node/2720337). It makes $id required from $group, even if we just want a list of the active groups. What happens on og_context set is that $cache_key becomes something like this - `node:123:0:TRUE`

Then when trying to retrieve a list of active groups with `og_context()`, the $cache_key becomes `node::0:TRUE` which **never** matches anything, thus always returning false.

I wonder if we need separate functions for setting, getting, and getting active og contexts. Perhaps something like this:
- `function og_context_set($group_type, $group)`
- `function og_context_get($group_type = 'node', $group = NULL, $account = NULL, $check_access = TRUE)`
- `function og_context_get_all($group_type = 'node')`

amitaibu:
@pianomansam Thanks. worth opening a new issue/ PR

PR #32, opened 2015-02-13
Added relationship handler alterations to OgBehaviorHandler.class.php

mdance:


amitaibu:
Hi, Can you explain a bit more what your trying to do here?

mdance:
I had an event content type which contained an og_group_ref entity reference field.

I created a view to list the events and I was trying to display a category taxonomy field associated with the group node.

The SQL query being generated was failing once I tried to create the group relationship for the bridged og_group_ref entity without this patch.

amitaibu:
Thanks

PR #33, opened 2015-02-20
Getting referencable entities only works for nodes.

sheise:
When checking the user's edit permisssion in buildEntityFieldQuery, things break when using non-node groups. Use $group_type instead.


amitaibu:
Thanks - any chance for a simpleTest? :)

sheise:
Thanks for merging. Unfortunately I don't have much time for a simple test right now.

Issue #34, opened 2015-02-25
Node context negotiation fails


kaaresl:
### Problem / Motivation

Issue #28 and commit 66d3b723 causes Node context negotiation to fail due to comparison between menu router path (pattern) and path instance:

``` php
        if (strpos(current_path(), $path) === 0) {
```

For nodes this result in comparison between e.g. `node/1107` and `node/%`.
### Proposed resolution

One of:
1. Revert
2. As explained in #28, the call to `menu_get_item()` may result in a new/nested call to og_context() due to menu router item object loading and translation. But one cannot compare a menu item path with a live path due to the above comparison failure, so you have to retrieve the menu router item from the current live path somehow. One solution is to implement a truncated version of `menu_get_item()` that halts processing the router item before item translation and object loading. It's only used to compare string. Leave it to the negotiation implementations to further check access control.

amitaibu:
> Revert or find a better method to match menu paths vs actual paths.

Yeah, lets revert

Issue #28, opened 2015-01-28
Do not load menu item just before determining context handler to prevent premature access denial due to empty og context.


jummonk:
Issue in og_context.module, function og_context_determine_context().

When determining the context, don't by default load the menu item and just test on current path. Because if during the custom menu access check og_context is called (as can be the case in a views access plugin or ctools access plugin where og_context would be called), then the statically cached initialized context will seem empty. 
Solution is not loading the menu item before determining the context, thus preventing a custom access being executed which could in turn execute og_context() to establish a context, resulting in denial of access as the context is initialized alrready, but empty.

![og_context-access-issue](https://cloud.githubusercontent.com/assets/3871797/5936575/365a126c-a6e6-11e4-98a7-29b50a19a5c6.png)

PR #35, opened 2015-02-25
Issue #34: Trimmed version of menu_get_path() as comparison.

kaaresl:
This has the same functionality as pre #28, only using a really basic version of menu_get_item.
### Remaining tasks
- Tests needed (I'm not the guy here)
- ~~Maybe extract this code to an internal function.~~
- ~~Add static cache of found menu router items?~~
- ~~Rename function to `_og_context_menu_get_(unprocessed|untranslated|raw)_item()`~~ (Went for `_og_context_menu_get_untranslated_item()` to better match the fact that `_menu_translate` isn't called).

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -386,9 +386,7 @@ function og_context_determine_context($group_type, $item = NULL, $check_access =
 
   $providers = og_context_negotiation_info();
   if (empty($item)) {
-    if (!$item = _og_context_menu_get_untranslated_item(current_path())) {
-      return;
-    }
+    $item = _og_context_menu_get_untranslated_item(current_path());
   }
 
   foreach ($enabled_providers as $name => $ignore) {
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -200,24 +200,42 @@ function og_context($group_type = 'node', $group = NULL, $account = NULL, $check
   // User account is sent.
   $account = $account ? $account : user_load($user->uid);
 
-  $context = &drupal_static(__FUNCTION__, FALSE);
+  $id = FALSE;
+  if ($group) {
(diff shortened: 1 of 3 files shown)


kaaresl:
Other than missing tests, I think this is review-worthy.

amitaibu:
@HelenaEksler can you please do a first review. Bonus points for a test

amitaibu:
I've reverted 66d3b723b6c305006acf5204fca27d7d2f30bc26 - @kaaresl can you please re-roll

kaaresl:
Did a `git rebase git rebase upstream/7.x-2.x` which caused some code leakage from #29, but otherwise, `og_context_determine_context()` is the same as prior to the revert of #29. So all in all, this PR is an alternative solution of 29 :-)

kaaresl:
Oh, this patch isn't ready. The parameter `$item` in `og_context_determine_context()` is ignored.

kaaresl:
That "code leakage" was not at all code leakage. It was removed in #29, hence the `$item` parameter got ignored. It's back and things should hopefully work with menu items given as parameter without calling `_og_context_menu_get_untranslated_item()`. Only `$item['path']` is ever used and this isn't touched in `_menu_translate()`.

kaaresl:
This PR and commit stack got a bit messy. A clean diff using `git diff -u --stat upstream/7.x-2.x` is available here: https://gist.github.com/kaaresl/dd03248fb8c02633c1fc

kaaresl:
Updated gist to match latest commit. Needs review.

amitaibu:
PR still seems to have non-related code. It should show only your own diff (you can always use `push -f` if you must )

kaaresl:
Forget this PR and head over to #40.

PR #39, opened 2015-03-06
Fixed issue with entity metadata wrapper not returning correct group value in og_is_group().

bdlangton:
This is a strange issue, and I'm not sure if it's actually an issue with the entity API. When I am saving a new node and select it to be in a certain group (the group is a node), then it says "The referenced group (node: 4) is invalid." and doesn't save the node. If I get "$wrapper->value()" and inspect the entity object, the group value is set to 1 (which it should be). However, "$wrapper->{OG_GROUP_FIELD}->value()" returns blank.


amitaibu:
Thanks, but this doesn't seem related to OG - are you able to reproduce this in a clean installation?

Anyway, this PR won't go in as is, since it's well - worked fine until now :)

PR #43, opened 2015-04-12
Checking user access for og_get_groups_by_user() and og_get_entity_groups().

HelenaEksler:
Using functions og_get_groups_by_user() and og_get_entity_groups() the user could get groups he doesn't have access to, e.g. unpublished ones.
So I added the check_access parameter to have an opportunity to get only groups user has access to.

### DIFF
--- a/og.module
+++ b/og.module
@@ -2294,9 +2294,9 @@ function og_user_access_entity($perm, $entity_type, $entity, $account = NULL) {
  */
 function og_get_entity_groups($entity_type = 'user', $entity = NULL, $states = array(OG_STATE_ACTIVE), $field_name = NULL, $access_check = FALSE) {
   $cache = &drupal_static(__FUNCTION__, array());
+  global $user;
 
   if ($entity_type == 'user' && empty($entity)) {
-    global $user;
     $entity = clone $user;
   }
   if (is_object($entity)) {
@@ -2349,15 +2349,14 @@ function og_get_entity_groups($entity_type = 'user', $entity = NULL, $states = a
     ->execute()
     ->fetchAll();
 
-  $account = user_load($id);
+  $account = $entity_type == 'user' ? user_load($id) : NULL;
   foreach ($result as $row) {
-    foreach (entity_load($row->group_type, array($row->gid)) as $group) {
-      if ($access_check && !entity_access('view', $row->group_type, $group, $account)) {
-        // User doesn't have access to the group.
-        continue;
-      }
-      $cache[$identifier][$row->group_type][$row->id] = $row->gid;
+    $group = entity_load_single($row->group_type, $row->gid);
+    if ($access_check && $account && !entity_access('view', $row->group_type, $group, $account)) {
+      // User doesn't have access to the group.
+      continue;
     }
+    $cache[$identifier][$row->group_type][$row->id] = $row->gid;
   }
 
   return $cache[$identifier];
--- a/og.module
+++ b/og.module
@@ -2349,12 +2349,15 @@ function og_get_entity_groups($entity_type = 'user', $entity = NULL, $states = a
     ->execute()
     ->fetchAll();
 
+  $account = user_load($id);
   foreach ($result as $row) {
-    if ($access_check && !entity_access('view', $row->group_type, entity_load_single($row->group_type, $row->gid), user_load($id))) {
-      // User doesn't have access to the group.
-      continue;
+    foreach (entity_load($row->group_type, array($row->gid)) as $group) {
+      if ($access_check && !entity_access('view', $row->group_type, $group, $account)) {
+        // User doesn't have access to the group.
+        continue;
+      }
+      $cache[$identifier][$row->group_type][$row->id] = $row->gid;
     }
-    $cache[$identifier][$row->group_type][$row->id] = $row->gid;
   }
 
   return $cache[$identifier];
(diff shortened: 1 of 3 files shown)


og.module
```diff
@@ -2347,6 +2350,10 @@ function og_get_entity_groups($entity_type = 'user', $entity = NULL, $states = a
     ->fetchAll();
 
   foreach ($result as $row) {
+    if ($access_check && !entity_access('view', $row->group_type, entity_load_single($row->group_type, $row->gid), user_load($id))) {
```
amitaibu: `entity_load_single()` should become `entity_load()`

HelenaEksler:
Using `entity_load()` causes problems. Will leave `entity_load_single()`. Anyway, I can't have more than one entity with the same entity type and entity ID, can I?

amitaibu:
> Will leave entity_load_single().

sorry, not possible - we can't have a performance hit on such a low level API function. please investigate why this causes an issue.

HelenaEksler:
@amitaibu Ready. Could you review this, please?

amitaibu:
Thanks. This requires a test, as it's dealing with access

HelenaEksler:
@amitaibu I know. That's why I've already added test. https://github.com/Gizra/og/pull/43/files#diff-165969fe289d202d15143228836ccb73R339

amitaibu:
Sorry, somehow missed them.

og_access/og_access.test
```diff
@@ -344,6 +344,32 @@ class OgAccessTestCase extends DrupalWebTestCase {
...
 
     $this->assertFalse(node_access('view', $node, $user2), 'Group member can not view unpublished node.');
+
+    $groups1 = og_get_entity_groups('user', $user2, array(OG_STATE_ACTIVE), NULL, TRUE);
+    $groups2 = og_get_groups_by_user($user2, 'node', TRUE);
+    $groups1 = $groups1 ? $groups1 : array('node' => array());
+    $groups2 = $groups2 ? $groups2 : array();
+
+
+    $this->assertTrue(node_access('view', $group_node1, $user2), 'Group member can view published group.');
```
amitaibu: why use node_access() and also entity_access()?

og_access/og_access.test
```diff
@@ -344,6 +344,32 @@ class OgAccessTestCase extends DrupalWebTestCase {
     drupal_static_reset('node_access');
 
     $this->assertFalse(node_access('view', $node, $user2), 'Group member can not view unpublished node.');
+
+    $groups1 = og_get_entity_groups('user', $user2, array(OG_STATE_ACTIVE), NULL, TRUE);
```
amitaibu: maybe move this whole code to it own test method?

amitaibu: Also a comment or two would be useful.

og_access/og_access.test
```diff
@@ -344,6 +344,32 @@ class OgAccessTestCase extends DrupalWebTestCase {
...
+    $groups2 = og_get_groups_by_user($user2, 'node', TRUE);
+    $groups1 = $groups1 ? $groups1 : array('node' => array());
```
amitaibu: I think it would be more readable to have the assert deal with the raw result, instead of adding a default value.

og_access/og_access.test
```diff
@@ -344,6 +344,32 @@ class OgAccessTestCase extends DrupalWebTestCase {
...
+    node_save($group_node1);
+    drupal_static_reset('node_access');
+    drupal_static_reset('og_get_entity_groups');
+
+    $groups1 = og_get_entity_groups('user', $user2, array(OG_STATE_ACTIVE), NULL, TRUE);
+    $groups2 = og_get_groups_by_user($user2, 'node', TRUE);
+    $groups1 = $groups1 ? $groups1 : array('node' => array());
+    $groups2 = $groups2 ? $groups2 : array();
+
+    $this->assertFalse(node_access('view', $group_node1, $user2), 'Group member can\'t view unpublished node.');
```
amitaibu: instead of `can\'t` you can use `"` to make the sentence more readable for devs

og_access/og_access.test
```diff
@@ -345,6 +345,59 @@ class OgAccessTestCase extends DrupalWebTestCase {
...
+    $group_node->status = NODE_NOT_PUBLISHED;
+    node_save($group_node);
+    drupal_static_reset('node_access');
+
+    $this->assertFalse(entity_access('view', 'node', $group_node, $user2), "Group member can't view unpublished group.");
+
+    // Clear the cache because permissions were changed.
+    drupal_static_reset('og_get_entity_groups');
+
+    // Get user's groups he has access to using og_get_entity_groups().
```
amitaibu: `groups he has ..` => `groups they have ..` (here and below)

because they can be a girl to -- like in your example :)

og_access/og_access.test
```diff
@@ -345,6 +345,59 @@ class OgAccessTestCase extends DrupalWebTestCase {
 
     $this->assertFalse(node_access('view', $node, $user2), 'Group member can not view unpublished node.');
   }
+
+  /**
+   * Test unpublished node.
+   */
+  function testUnpublishedGroupAccess() {
```
amitaibu: this test should move under `og.test`. `og_access.test` should test only functionality of OG access module.

rbran100:
@HelenaEksler Thanks for this, error from tests:

```
OgException: Cannot set visibility of node ID 2 as the node [error]
group of type mktfk304 does not have the "Group visibility" field
attached to it. in _og_access_verify_access_field_existence() (line
259 of /home/travis/build/Gizra/og/og_access/og_access.module
```

I will take a look at the error if I get a chance.

I also wanted weigh in here, that some of the internal calls in og should probably call with the access check on.  og_context_get_context makes sense to grab the first published node.

claudiu-cristea:

PR #48, opened 2015-05-21
Create og_context_init function to determine that og_context is already running.

HelenaEksler:


og_context/og_context.module
```diff
@@ -177,10 +177,29 @@ function og_context_og_invalidate_cache($gids = array()) {
   }
 }
 
+/**
+ * Determines if OG context has been already called.
+ * og_context function should not be called again until it is still running.
```
amitaibu: Line break missing.

amitaibu: actually this line isn't needed.

og_context/og_context.module
```diff
@@ -177,10 +177,29 @@ function og_context_og_invalidate_cache($gids = array()) {
...
+ *
+ * @param bool $called
+ *  Determines if og_context function is still running.
```
amitaibu: `Determines if og_context() function has already been executed. Defaults to FALSE.`

og_context/og_context.module
```diff
@@ -177,10 +177,29 @@ function og_context_og_invalidate_cache($gids = array()) {
...
+ *
+ * @return bool
+ *  Returns TRUE if function is still running.
```
amitaibu: `still running` isn't correct. You mean `was already executed`

og_context/og_context.module
```diff
@@ -177,10 +177,29 @@ function og_context_og_invalidate_cache($gids = array()) {
   }
 }
 
+/**
+ * Determines if OG context has been already called.
+ * og_context() function should not be called again if it was already executed.
```
amitaibu, line 182: minor: there should be a line break after the first line.

amitaibu:
I've fixed the minor comment and merged, thanks.

pianomansam:
What is the purpose of og_context_init()? It never seems to do anything!

https://www.drupal.org/node/2713039

PR #49, opened 2015-06-03
"OG: user role in group" access plugin

Leksat:
Hi!

Can we have "OG: user role in group" access plugin in the OG core? It would be very handy in some cases.

### DIFF
--- a/og.module
+++ b/og.module
@@ -2630,6 +2630,21 @@ function og_role_permissions($roles = array()) {
 }
 
 /**
+ * Returns an array of all roles.
+ *
+ * @return array
+ *   An associative array with the role id as the key and the role name as
+ *   value.
+ */
+function og_get_all_roles() {
+  return db_select('og_role', 'ogr')
+    ->fields('ogr', array('rid', 'name'))
+    ->orderBy('rid', 'ASC')
+    ->execute()
+    ->fetchAllkeyed();
+}
+
+/**
  * Retrieve an array of roles matching specified conditions.
  *
  * @param $group_type
--- a/plugins/access/og_role.inc
+++ b/plugins/access/og_role.inc
@@ -1,0 +1,86 @@
+<?php
+
+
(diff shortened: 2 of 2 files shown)


og.module
```diff
@@ -2630,6 +2630,21 @@ function og_role_permissions($roles = array()) {
...
 
 /**
+ * Returns an array of all roles.
+ *
+ * @return array
+ *   An associative array with the role id as the key and the role name as
+ *   value.
+ */
+function og_get_all_roles() {
+  return db_select('og_role', 'ogr')
```
amitaibu, line 2640: this won't scale if we have multiple roles for each group (i.e. group that have their own roles)

plugins/access/og_role.inc
```diff
@@ -0,0 +1,86 @@
...
+
+
+/**
+ * @file
+ * Plugin to provide access control based on user roles in a group.
+ */
+
+$plugin = array(
+  'title' => t('OG: user role in group'),
+  'description' => t('Control access by group role.'),
```
amitaibu, line 11: `t('Control access by group role. This applies only to node type groups.'),`

plugins/access/og_role.inc
```diff
@@ -0,0 +1,86 @@
...
+function og_role_ctools_access_check($conf, $context) {
+  list($user_context, $node_context) = $context;
+  if (empty($user_context->data) || empty($node_context->data)) {
+    return FALSE;
+  }
+  $account = $user_context->data;
+  $node = $node_context->data;
+
+  $user_og_roles = og_get_user_roles('node', $node->nid, $account->uid, FALSE);
+  return count(array_intersect($conf['og_rids'], array_keys($user_og_roles))) > 0;
```
amitaibu, line 59: should be enough to just `return array_intersect($conf['og_rids'], array_keys($user_og_roles));`

PR #50, opened 2015-06-04
Allow hook_og_ui_get_group_admin_alter() to change the sort order of group admin items

DavidRothstein:
The current code prevents doing that, but I don't see any good reason that it should.


amitaibu:
Thanks!

PR #60, opened 2015-07-28
Consider to use LionsAd/drupal_ti

fabianx-ai:
Hi,

I have been working on making testing of modules on travis easier.

This will also enable easy PHP 5.3 testing on the container infrastructure.

In addition to that adding e.g. behat tests or unit tests in the future is as simple as assing a new DRUPAL_TI_RUNNER.

This PR is here mainly for checking if travis passes.


MPParsley:
Hi @LionsAd, I like the idea of having a package to manage Travis setup.

I can imagine we could deprecate Travis CI now that GitHub has its own CI with GitHub actions, is LionsAd/drupal_ti still supported?

fabianx-ai:
It’s in maintenance mode with bug fixes only - I would say. It still seems to be used even though I am right now not actively using it myself.

PR #63, opened 2015-08-24
Replace isGroup logic with bundle entity based storage

damiankloip:
This is still very much a WIP, trying out using the bundle config entity as storage for OG group flags.

### DIFF
--- a/src/Og.php
+++ b/src/Og.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\og\Og.
+ */
+
+namespace Drupal\og;
+
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * A static helper class for OG.
+ */
+class Og {
+
+  /**
+   * A static cache of entity type group mappings.
+   *
+   * @var array
+   */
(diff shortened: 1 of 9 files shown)


src/Og.php
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * Returns whether an entity type instance is a group type.
+   *
+   * @param \Drupal\Core\Config\Entity\ConfigEntityInterface $entity_type
+   *
+   * @return bool
+   */
+  public static function isGroupEntityType(ConfigEntityInterface $entity_type) {
```
damiankloip, line 70: Stuff like this could make this a no go. To use this storage we need to assume that bundle entities will be using config storage interface. We could also check `ContentEntityTypeInterface`, but we would still need config storage to provide this.

amitaibu:
Thanks. Any chance to hold the PRs until the beginning of September when I'm back?

I think it would make more sense and productive to tackle this after I'll go over the existing code with you and talk about design decisions.

damiankloip:
@amitaibu sure, we can hold off. As mentioned in the description this is very much a WIP, just wanted to posted this as I had been playing around with it locally. It's not a problem to wait until you return. Is that next week?

amitaibu:
Yes, I'm back on the 1st. We can schedule a Skype via email :)

damiankloip:
Closing in favour of https://github.com/amitaibu/og/pull/2 for now

Issue #66, opened 2015-09-06
[D8] Create a proper entity reference widget.


RoySegall:
@damiankloip I looked in your repo that you started to do some work over there. I'll see if this can be merged in.

PR #67, opened 2015-09-06
WIP: Port OG group reference to D8.

RoySegall:
#66

I took @damiankloip work. From a first look it works like in D7: The field storage in the DB is not populated and membership is created.

### DIFF
--- a/src/Controller/OG.php
+++ b/src/Controller/OG.php
@@ -10,8 +10,6 @@ use Drupal\field\Entity\FieldConfig;
 use Drupal\field\Entity\FieldStorageConfig;
 use Drupal\og\Entity\OgMembership;
 use Drupal\og\OgFieldBase;
-use Symfony\Component\HttpFoundation\JsonResponse;
-use Symfony\Component\HttpFoundation\Request;
 
 class OG {
 
@@ -159,4 +157,5 @@ class OG {
   public static function MembershipDefault() {
     return array('type' => 'og_membership_type_default');
   }
+
 }
--- a/src/Controller/OG.php
+++ b/src/Controller/OG.php
@@ -145,6 +145,7 @@ class OG {
    *   The matched labels as json.
    */
   public static function handleAutocomplete(Request $request, $type, $field_name, $entity_type, $bundle_name, $entity_id) {
+    // todo (maybe) pass this logic to a to the field handler.
     $definitions = \Drupal::entityManager()->getFieldDefinitions($entity_type, $bundle_name);
 
     if (!isset($definitions[$field_name])) {
@@ -173,7 +174,7 @@ class OG {
       ->execute();
 
     /** @var EntityInterface[] $entities */
-    $entities = entity_load_multiple($entity_type, $results);
+    $entities = \Drupal::entityManager()->getStorage($entity_type)->loadMultiple($results);
 
     $matches = array();
     foreach ($entities as $entity) {
--- a/src/Controller/OG.php
+++ b/src/Controller/OG.php
@@ -12,8 +12,6 @@ use Drupal\og\Entity\OgMembership;
 use Drupal\og\OgFieldBase;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
-use Drupal\Component\Utility\Tags;
 
 class OG {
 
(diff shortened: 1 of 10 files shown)


src/Plugin/Field/FieldType/OgMembershipItem.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /** @var \Drupal\Core\Entity\EntityInterface $parent */
+    $parent_entity = $this->getEntity();
+    $membership = OG::MembershipStorage()->create(OG::MembershipDefault());
+
+    $membership->setFieldName($this->getName())
+      ->setEntityType($parent_entity->getEntityTypeId())
+      ->setEntityId($parent_entity->id())
+      ->setGroupType($this->entity->getEntityTypeId())
+      ->setGid($this->entity->id())
```
amitaibu: lets change this to `setGroupId` (can be in a follow up)

src/Plugin/Field/FieldType/OgMembershipItem.php
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function postSave($update) {
+    parent::postSave($update);
+
+    /** @var \Drupal\Core\Entity\EntityInterface $parent */
+    $parent_entity = $this->getEntity();
```
amitaibu: `$parent_entity` -> `$group_content_entity`

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,109 @@
...
+ * @file
+ * Contains \Drupal\og\Plugin\Field\FieldType\OgMembershipItemList.
+ */
+
+namespace Drupal\og\Plugin\Field\FieldType;
+
+use Drupal\Core\Field\EntityReferenceFieldItemList;
+
+/**
+ * Defines a item list class for OG membership fields.
```
amitaibu, line 16: a item => an item

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Populate the list property with groups based on related memberships.
+   */
+  protected function populateGroupsFromMembershipEntities() {
+    // Make sure list is clear.
+    $this->list = [];
+    $entity = $this->getEntity();
+    $group_type = $this->getFieldDefinition()->getSetting('target_type');
+
+    $membership_ids = \Drupal::entityQuery('og_membership')
```
amitaibu, line 162: we might need to check the state of the membership here.

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,109 @@
...
+
+    return isset($this->list[$index]) ? $this->list[$index] : NULL;
+  }
+
+  /**
+   * @param $property_name
+   *
+   * @return mixed
+   */
+  public function __get($property_name) {
```
amitaibu, line 54: this part is not clear to me.

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,109 @@
...
+
+    // Fetch the list of memberships for this field.
+    if (!$this->fetched) {
+      $this->populateGroupsFromMembershipEntities();
+      $this->fetched = TRUE;
+    }
+
+    // Automatically create the first item for computed fields.
+    if ($index == 0 && !isset($this->list[0])) {
+      $this->list[0] = $this->createItem(0);
```
amitaibu, line 43: can you clarify what's going on here? :)

amitaibu:
great start guys!

src/Controller/OG.php
```diff
@@ -145,6 +145,7 @@ public static function invalidateCache($gids = array()) {
    *   The matched labels as json.
    */
   public static function handleAutocomplete(Request $request, $type, $field_name, $entity_type, $bundle_name, $entity_id) {
+    // todo (maybe) pass this logic to a to the field handler.
```
RoySegall: @amitaibu What do you think?

amitaibu: you mean the ER field? Sounds right

damiankloip:
Thanks for opening, and reviewing. That was still very much a work in progress though :) It's kind of half complete. It does work OK though for an initial effort.

damiankloip:
I think it would be good to leave all widget related stuff for another PR, too much in one go makes things harder to review etc... Using a selection handler etc.. is definitely the way to go though.

src/Plugin/Field/FieldType/OgMembershipItemList.php
damiankloip, line 162: Yes, for sure.

amitaibu:
> It does work OK though for an initial effort.

Which is totally great :smile: 

> I think it would be good to leave all widget related stuff for another PR

Agreed

src/Plugin/Field/FieldType/OgMembershipItemList.php
damiankloip, line 43: This makes the same behaviour as other computed/list field items.

src/Plugin/Field/FieldType/OgMembershipItem.php
damiankloip: Agree. :+1:

src/Plugin/EntityReferenceSelection/OgSelection.php
```diff
@@ -0,0 +1,45 @@
...
+   *   (Optional) The operation the matching should be done with. Defaults
+   *   to "CONTAINS".
+   *
+   * @return \Drupal\Core\Entity\Query\QueryInterface
+   *   The EntityQuery object with the basic conditions and sorting applied to
+   *   it.
+   */
+  protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS') {
+    $query = parent::buildEntityQuery($match, $match_operator);
+    $query->condition(OG_GROUP_FIELD, 1);
```
damiankloip, line 41: @amitaibu we need to talk about whether https://github.com/Gizra/og/pull/63 is feasible, as it would change how this works.

amitaibu, line 41: Sure, are you available for a quick call?

damiankloip, line 41: I am still out of office at the moment. Not back until tomorrow afternoon! so could do then, or tuesday morning some time?

RoySegall:
OK. So what I got left is to fix the widget if the autocomplete. If i have 2 groups that the content belong two and i have this code:

```
  /**
   * {@inheritdoc}
   */
  public static function getWidgetState(array $parents, $field_name, FormStateInterface $form_state) {
    $widgetState = parent::getWidgetState($parents, $field_name, $form_state);
    $widgetState['items_count'] = 2;
    return $widgetState;
  }
```

Then everething OK. I can see all the groups i'm reference to. The problem is to know how to do the query since i don't have any info on the entity in that method. I'll check.

damiankloip:
@RoySegall let's move widget stuff to a separate PR. I think work in my branch should be one PR and your new work should be another. In other words, I think we should close this and open two new ones :)

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,152 @@
...
+   * @return mixed
+   */
+  public function __get($property_name) {
+    // For empty fields, $entity->field->property is NULL.
+    if ($item = $this->first()) {
+      return $item->__get($property_name);
+    }
+  }
+
+  public function postSave($update) {
```
damiankloip, line 61: How come this moved? To get all the groups?

src/Plugin/Field/FieldType/OgMembershipItemList.php
```diff
@@ -0,0 +1,152 @@
...
+    parent::postSave($update);
+
+    $new_groups = [];
+    foreach ($this->list as $item) {
+      $new_groups[] = $item->getValue()['target_id'];
```
damiankloip, line 66: This is not new groups, just groups I think.

src/Plugin/Field/FieldWidget/OgComplex.php
```diff
@@ -79,4 +97,26 @@ public function elementValidate($element, FormStateInterface $form_state, $form)
...
+   * {@inheritdoc}
+   */
+  public static function getWidgetState(array $parents, $field_name, FormStateInterface $form_state) {
+    $widgetState = parent::getWidgetState($parents, $field_name, $form_state);
+
+    if (empty(self::$info)) {
+      return $widgetState;
+    }
+    $info = self::$info;
+    $results = \Drupal::entityQuery('og_membership')
```
damiankloip, line 111: This will just run every time. As self::$info is not assigned to anything, and if `empty(self::$info)` is confusing. Is the static caching needed? This is just for the widget, so this should not be running multiple times per request per widget anyway?

damiankloip, line 111: Also, doesn't the work I started to do on loading make this less of a problem? The widget should get any values from the field item (as it does now) and use that. Rather than any funky stuff in the widget itself?

damiankloip, line 111: Oh. Or do you mean the query for the autocomplete entity reference selection handler?

amitaibu:
@RoySegall @damiankloip I just had an idea (which I'm not sure why I haven't thought of for OG7) - instead of blocking the storage, and then creating the OG membership on `post` hooks, we could write our own storage implementation which will CRUD the OG membership.

This will provide a more natural flow of OG, as the field's data pipes into the field storage itself.

damiankloip:
we should close this now, as we got https://github.com/amitaibu/og/pull/1 merged over there. @RoySegall You should be able to get any widget work from here easy enough and open a new PR against https://github.com/amitaibu/og

RoySegall:
If so - I'll close this PR and we will keep the work over there.

PR #91, opened 2015-10-23
Port tests to 8.x

jhedstrom:
This is a WIP, but could be merged anytime since thus far the tests are still in the old 7.x format. Currently these all load on the test overview, but will all fatal error when run.


amitaibu:
Thanks. Better do the work against amitaibu/og , as porting is currently
done there.

jhedstrom:
I think this is outdated.

PR #98, opened 2015-12-30
removed og_membership from hook_schema 

matslats:
Og_membership entity defines its own schema and the one in hook_schema contradicts it and is out of date.


amitaibu:
PR should be against `amitaibu/og` where developing is happening :)

PR #101, opened 2016-01-27
"OG: user role in group" access plugin

Zekvyrin:
A follow up for:

https://github.com/Gizra/og/pull/49

I did the changes mentioned by amitaibu plus made some other changes. 

I wanted to use it in a custom page made by page manager (panels), which didn't have context, so now node_context is optional.

My use case is that I wanted to create something like a "dashboard" for all "Group admins" in which each admin will be able to see whatever he needs to see from his own group.

Any feedback is welcomed :)

### DIFF
--- a/og.module
+++ b/og.module
@@ -2637,6 +2637,49 @@ function og_role_permissions($roles = array()) {
 }
 
 /**
+ * Returns an array of all roles.
+ *
+ * @param $include_all
+ *   (optional) If TRUE then the anonymous and authenticated default roles will
+ *   be included.
+ *
+ * @return array
+ *   An associative array with the role id as the key and the role name as
+ *   value.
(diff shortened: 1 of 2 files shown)


amitaibu:
We still have [this](https://github.com/Gizra/og/pull/49#discussion_r31600737) problem. I believe this code can easily live in contrib, and not OG core.

amitaibu:
(but thanks for contributing :wink: )

Zekvyrin:
I included that change mentioned [here](https://github.com/Gizra/og/pull/49#discussion_r31600737). Was there something more to be done?

Well.. I think it could be included in core, or at least the a og_get_all_roles() function would be helpful. 
For the plugin itself.. yes it could live in another module.

PR #104, opened 2016-01-28
Changing .travis.yml to use composer rather than PEAR for drush install.

grasmash:


amitaibu:
Thanks. closed in favor of #187

Issue #106, opened 2016-02-11
Only a subset of the group nodes get a node_access_acquire_grants() when the group access changes.


zero2one:
When the group access type changes, a batch is run to recalculate the grants for all nodes belonging to that group.

The batch processes 50 items stores the last id and triggers a next set.
The next set is loaded by the last etid that was processed.

The problem is that loading a set of 50 items is not ordered in any way so we get the records ordered by the order they were created.
### Solution

Order the og_memberships by the etid.

amitaibu:
Thanks!

zero2one:
you're welcome!

PR #107, opened 2016-02-11
Issue #106: Order by the entity ID.

zero2one:
Order by the entity ID as the memberships are not ordered by this by default.
This results in only a subset of the content that gets a node_access_acquire_grants().


amitaibu:
Hi!

Seems that the Travis config needs some love. Anyway, thanks for the PR - merged

zero2one:
See #106

PR #109, opened 2016-02-17
Use theme('username') for usernames in the admin interface.

zero2one:
Using the $account->name as the value on the screen limits the possibility to format the username using the theming layer.

Replacing it by 

```
theme('username', array('account' => $account));
```

This gives modules, site builders and themers the power to replace the name with another value (eg. Full name specified in the user profile).


zero2one:
See also: https://www.drupal.org/node/2670416

claudiu-cristea:

PR #111, opened 2016-03-12
Support for uri entity ID

idimopoulos:
Actually this pull request is mostly about supporting more characters in IDs.
For my case, a URI can contain many of the symbols not supported by the '.' symbol in preg_match. So the "other_groups" fail to identify items when there are special characters in the ID.
Still, this should be better handled in the future because it still breaks if for some reason parenthesis symbols are entered in the string ID.


amitaibu:
Oh my, it should probably be against `amitaibu/og` :wink:

idimopoulos:
pff... annoying mistakes I'm making... indeed this was not intended for this. sorry. Closing.

amitaibu:
meh, all the counts is that you did a PR! :)

Issue #121, opened 2016-05-06
Add access true/false option in og_context_plugin_argument_default_group_context after og_context() rewrite in 7.x-2.8


alesrebec:
> Copied from https://www.drupal.org/node/2720403

Since the big og_context() changes made in OG 7.x-2.8 anonymous users, anonymous users get FALSE on og_context() call. Prior to that version, the return was an array with og gid in context.
My understanding is that this is (at least a minor) breakage of an OG 7.x-2.x API.

To keep the required behaviour you have to rewrite all og_context() function calls and include "$check_access = FALSE" so the access check is skipped and og_context() function acts like it was until 7.x-2.8.

This was added:

```
 * ...
 * @param bool $check_access
 *   Determines if access to the group should be done. Defaults to TRUE.
 ...
 */
function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = TRUE) {
...
}

```

I had this bug somewhere deep in a complex site. I wouldn't even noticed it myself. I got a support ticket created that there's something wrong on that page.
There are thousands of sites using OG and only question is how many sites are broken because of this change.

alesrebec:
Closed.
https://www.drupal.org/node/2720403#comment-11186941

alesrebec:
Copied from: https://www.drupal.org/node/2720403#comment-11195599

> @amitaibu, I did changed all function calls to og_context() with $access_check set to FALSE (as suggested).
> The problem that appeared after that was in og_context_plugin_argument_default_group_context.inc views handler where an option to select between Check access (YES or NO) was missing in views contextual filter handler for og context.
> 
> I had two options. Create an own handler which would be a copy of this one with added option to select Check access or patch this one, the same way og_context() function was patched with $check_access parameter which is set to TRUE by default.
> Actually this should come in with that og_context() patch by Helena.
> If we allow to call og_context() with $check_access set to FALSE, we need to allow this in views handler too.

amitaibu:
Thanks. Can you please close the d.o. issue and open a PR

kyletaylored:
I went ahead and closed a D.O ticket that was related to this patch as well.

[Adding non-Required OG Relationship to View Causes View to Display no Results for Anonymous](https://www.drupal.org/node/2339953#comment-12163432)

alesrebec:
Pull request is still unmerged in #122 
Should be released in version 7.x-2.10.

We can keep a list of related issues in one place.
[...]

PR #122, opened 2016-05-06
#121 - og_context() default value set to FALSE.

alesrebec:
og_context() default value set to FALSE to keep the same og_context() return for all users as it was up to version 7.x-2.7
https://github.com/Gizra/og/issues/121

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -216,7 +216,7 @@ function og_context_is_init($called = FALSE) {
  *   Array keyed by the group type and group ID, or FALSE if no context
  *   was found.
  */
-function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = TRUE) {
+function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = FALSE) {
   global $user;
   // User account is sent.
   $account = $account ? $account : user_load($user->uid);


alesrebec:
@amitaibu drush is messing up with travis here so the test cannot be run.

og_context/og_context.module
```diff
@@ -216,7 +216,7 @@ function og_context_is_init($called = FALSE) {
  *   Array keyed by the group type and group ID, or FALSE if no context
  *   was found.
  */
-function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = TRUE) {
+function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = FALSE) {
```
amitaibu, line 219: I'd be against that change, since I'd think that OG should take a restrictive approach.

If my og_context() returns an unpublished node, I'd argue that 99% of the time, this is not the context you'd want to show. What do you think?

alesrebec: I will agree!

Restrictive approach over breaking one function.
I think we already agreed on that when this function call was amended in og_context views handler https://github.com/Gizra/og/pull/131 https://github.com/Gizra/og/pull/131/files

All we need to do is let people know they have to amend their og_context() function call to make it work "the old way" and close all related issues as "Works as designed"
https://github.com/Gizra/og/issues/121

amitaibu: 👍 

You can also add a change record, for bonus points 😄

alesrebec:
Closing this PR #122 together with issue #121 in favour of more restrictive approach by default.

PR #124, opened 2016-05-22
Fixed #120: New member state is pending, should be active when added by admin

### LINKED ISSUE #120: New member state is pending, should be active when Admin adds user via profile widget
**The Bug:**
When site administrator adds a user to a group (ours are invite only) using the widget on the user's profile page, the user's state is set to _pending_ instead of _active_.

**Context:**
1. Version: og 7.x-2.9
2. Likely caused by security update of og 2.9 in file "OgBehaviorHandler.class.php" on line 127 located in og/plugins/entityreference/behavior/ possibly because membership state provided by the widget is empty.
3. What's the best fix for this: Make the widget provide user's state? Or check if actor has administer group permission and set state to _active_ if they do?

**Culprit Code:**

```
    // Prepare an array with the membership state, if it was provided in the widget.
        ...
      // Must provide correct state in the event that approval is required.
      if (empty($item['state']) && $entity_type == 'user' && !og_user_access($group_type, $gid, 'subscribe without approval', $entity)) {
        $item['state'] = OG_STATE_PENDING;
      }...

```

**Slow Work around:**
Site administrator must add user to groups one by one from each group's add user page so that user's states are correctly set to _active_.

alex-bukach:


jcnventura:
Agree, admin should be able to promote to active directly.

mmcintosh (commented):
found this via the drupal issue queue. Just wanted to take a look and perhaps test as I have this problem on a site.

https://www.drupal.org/node/2744405

mmcintosh (commented):
I tested this code on OG 7.2.9 where I had the same issue as described in this issue:

https://www.drupal.org/node/2744405

I found the cope work as expected and now an administrator can add a user to a group and have it be active instead of pending. This was the behavior that was previously in place.

amitaibu:
Seems there's a patch in https://www.drupal.org/node/2744405, so closing to prevent duplication.

alex-bukach:
The patch posted there on d.o 3 days ago is the exactly the same I posted in this pull request in May? Why shouldn't it be committed?

alex-bukach:
If "Development and issues are handled on GitHub" (as mentioned at https://www.drupal.org/project/og), why to close the pull request on Githuub in favour of the d.o issue?

amitaibu:
Sorry, the reason was that currently there's the #133 which is blocking. you are right, I'll close the d.o. one.

The reason it's not committed is - https://www.drupal.org/node/2744405#comment-11735053

geek-merlin:
> The reason it's not committed is - https://www.drupal.org/node/2744405#comment-11735053

So the [test](https://www.drupal.org/pift-ci-job/504491) failed, saying:

> fail: [Other] Line 461 of sites/all/modules/og/og_access/og_access.test:
User membership is pending.

I did not go deeper into this, but it looks like the test must be changed too.
(maybe discerning admin role.)

geek-merlin:
CI sais:
> The command "pear channel-discover pear.drush.org" failed and exited with 1 during .

That's not a problem of the patch...

amitaibu:
Thanks!

jackaponte:
> I don't see this commit in og 7.x-2.10 on drupal.org. Has it been fixed elsewhere now or was it just omitted?

I ran into the same question myself today while applying updates. This diff of the og-7.x-2.9 file in question against the 7.x-2.10 version seems to indicate that this commit/patch is no longer needed because the code affected by the patch is completely gone:

```
diff --git b/sites/all/modules/og/plugins/entityreference/behavior/OgBehaviorHandler.class.php a/sites/all/modules/og/plugins/entityreference/behavior/OgBehaviorHandler.class.php
index dc7373f7..d17a8e26 100644
--- b/sites/all/modules/og/plugins/entityreference/behavior/OgBehaviorHandler.class.php
+++ a/sites/all/modules/og/plugins/entityreference/behavior/OgBehaviorHandler.class.php
@@ -122,11 +122,7 @@ class OgBehaviorHandler extends EntityReference_BehaviorHandler_Abstract {
     foreach ($items as $item) {
       $gid = $item['target_id'];
 
-      // Must provide correct state in the event that approval is required.
-      if (empty($item['state']) && $entity_type == 'user' && !og_user_access($group_type, $gid, 'subscribe without approval', $entity) && !og_user_access($group_type, $gid, 'administer group')) {
-        $item['state'] = OG_STATE_PENDING;
-      }
-      elseif (empty($item['state']) || !in_array($gid, $diff['insert'])) {
+      if (empty($item['state']) || !in_array($gid, $diff['insert'])) {
         // State isn't provided, or not an "insert" operation.
         continue;
       }
```

(cross-posting to https://www.drupal.org/node/2744405)

ghost:
@jackalope: I think you're right. It's been a month now and I'm seeing no problems.

PR #126, opened 2016-05-22
Fixed #125: Behavior for og_membership_delete_action

### LINKED ISSUE #125: Behavior for og_membership_delete_action
(Duplicate of https://www.drupal.org/node/2730291)
### Problem/motivation

OG actions do not set `behavior` parameter key used by VBO (see e.g. https://www.drupal.org/node/2052067). In this case VBO uses default behavior value `array('changes_property')`. 

However when an action has `changes_property` behavior, VBO re-saves the entity after executing the action. 

I.e. for `og_membership_delete_action` action VBO tries to re-save a membership entity after having deleted it. Though it doesn't re-creates the membership entity, it causes executing `hook_og_membership_update()` which is confusing.
### Proposed solution

Set behavior for `og_membership_delete_action` explicitly to `array('deletes_property')`.

alex-bukach:


jacobembree:
The Drupal.org issue for this is [Behavior for og_membership_delete_action](https://www.drupal.org/project/og/issues/2730291).

Issue #133, opened 2016-07-04
Fix failing travis for D7


amitaibu:
We can probably install higher PHP as-well

@nirgn975 this one has your name on it :)

krueschi:
When tinkering around with travis-ci testing I found that installing drush with composer (instead of pear) brings up no error message. So I changed

```
- pear channel-discover pear.drush.org
- pear install drush/drush-5.8.0
```

and put instead

```
- export PATH="$HOME/.composer/vendor/bin:$PATH"
- composer global require drush/drush:6.*
```

BUT: I have not yet figured out how to install the drupal testing environment using the drush site-install command correctly, so my travis-ci tests fail.

Maybe it helps nonetheless.

Credits for this go to https://www.drupal.org/u/dsnopek at https://www.drupal.org/node/2195445.

AronNovak:
Now the Travis itself is fixed, the test execution is there, the issue is that the tests seem to be broken right now:
![test](https://cloud.githubusercontent.com/assets/114076/25116073/b49397d6-240a-11e7-885e-61e01d7372a0.png)

AronNovak:
Of course that"s something to validate against a local execution

AronNovak:
I have almost the same failures at localhost, so it"s not Travis specific that it fails:
![localfail](https://cloud.githubusercontent.com/assets/114076/25116780/224cdf0e-240f-11e7-9b46-91c3d25611c4.png)
@amitaibu should we proceed in the same PR by fixing the tests itself in addition to Travis?

AronNovak:
We need to restore the internal webserver and re-evaluate the above findings.
```
Pass      Browser    og_ui.test         309 OgUiManagePeopleTestCase->testOgUiD
    GET http://localhost/group/entity_test/1/admin/people/delete-membership/2
    returned 404 (250 bytes).
Pass      Browser    og_ui.test         309 OgUiManagePeopleTestCase->testOgUiD
    Valid HTML found on
    "http://localhost/group/entity_test/1/admin/people/delete-membership/2"
Fail      Other      og_ui.test         309 OgUiManagePeopleTestCase->testOgUiD
    Found the Remove button
```

AronNovak:
Almost all the tests pass, but:
```
[...]

PR #134, opened 2016-07-04
Fix failing travis for D7

nirgn975:
#133

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,4 +36,8 @@ before_script:
   - drush runserver 127.0.0.1:8080 &
   - until netstat -an 2>/dev/null | grep '8080.*LISTEN'; do true; done
 
-script: ~/.composer/vendor/bin/drush.php test-run "Organic groups","Organic groups access","Organic groups context","Organic groups field access","Organic groups UI" --uri=http://127.0.0.1:8080
+# script: ~/.composer/vendor/bin/drush.php test-run "Organic groups","Organic groups access","Organic groups context","Organic groups field access","Organic groups UI" --uri=http://127.0.0.1:8080
+script:
+  - cd $TRAVIS_BUILD_DIR/../drupal
+  - php ./scripts/run-tests.sh --php $(which php) --concurrency 4 --verbose --color --uri=http://127.0.0.1:8080 og 2>&1 | tee /tmp/simpletest-result.txt
+  - exit 0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,9 @@ before_script:
   - cd ../..
 
   # install drush
+  - pear upgrade –force Console_Getopt
```
amitaibu: use composer to get Drush, like we have in other projects

nirgn975:
@amitaibu Now we have drush (installed with composer) (in `d1539c1` I print `drush --help` to make sure it's installed globally).

But I get a new error: 
build__183_-_gizra_og_-_travis_ci

Any ideas?

amitaibu:
@ordavidil might know

amitaibu:
Try copying config from Message - https://github.com/Gizra/message/blob/7.x-1.x/.travis.yml

ordavidil:
10x for covering up ...

nirgn975:
@amitaibu I looked at this, but couldn't figure it out. There is no `scripts/run-tests.sh` there.. What file is it run?

amitaibu:
It runs Drupal core's test-runner. See also [RESTful as example](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/.travis.yml#L88).

Better copy and just do the adjustments. `php ./scripts/run-tests.sh` from Drupal's root should work on your local as-well.

amitaibu:
This is a blocking issue - would love t get it fixed.

nirgn975:
@amitaibu The tests are working now, but couple of them are fail:
build__196_-_gizra_og_-_travis_ci

I don't think we should fix them in this PR, am I right?

amitaibu:
We actually need to fix it here. Thanks,  I'll give it a look.

amitaibu:
@nirgn975 can you continue looking at this please, and try to figure why it fails. Does it fail on your local as-well?

amitaibu:
What's the status here?

nirgn975:
@amitaibu Tried to run the tests on my local - without success. And I don't have any idea to why it fails.

amitaibu:
I'm able to run the failing tests on my local via the UI. Can you try with [this](https://github.com/Gizra/message/blob/7.x-1.x/.travis.yml#L49) command

amitaibu:
closed in favor of #187

PR #137, opened 2016-07-16
Cannot save group reference field programmatically

### LINKED ISSUE #136: Cannot save group reference field programmatically
Please see https://www.drupal.org/node/2766637

glekli:
Copied from https://www.drupal.org/node/2766637

OG's entity reference handler class, OgBehaviorHandler, checks field access before it saves the group memberships, and skips saving if the currently logged in user does not have permission to edit the group reference field. As a result, if the current user does not have permission to edit the group reference field, the field is going to be ignored when the entity is saved programmatically. This can cause problems in two situations:
- The current user is irrelevant to the content being updated. E.g. during drush run
- The user may not have access to directly edit the group reference field, but custom code may want to be able to change the group memberships based on some other action taken by the user.

When content is saved programmatically, (e.g. using node_save()), the values that get saved into the database should not depend on what user is logged in. Actually, I think the entity reference handler class should not be concerned with user access when saving the group memberships. This is low level functionality, and when we get to the point where the entity needs to be saved, the field values should already have been validated.

This PR removes the access check from the membership saving logic. Please let me know if I have taken the right direction with this.

### DIFF
--- a/og_context/includes/views/handlers/og_context_plugin_access_og_perm.inc
+++ b/og_context/includes/views/handlers/og_context_plugin_access_og_perm.inc
@@ -59,7 +59,7 @@ class og_context_plugin_access_og_perm extends views_plugin_access {
         '#title' => t('Argument position for group ID'),
         '#default_value' => $this->options['group_id_arg'],
         '#options' => array(NULL => t('None')) + range(0, 9),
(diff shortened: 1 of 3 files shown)


claudiu-cristea:

PR #140, opened 2016-08-10
Remove reliance on current path in Views access check

glekli:
The access callback of the Views access plugin directly calls arg() to get the group id from the path. This makes the access check inaccurate when access is checked on any page other than the view's actual path. Such checks may occur when:
- The view is embedded on another page
- Any page needs to print a link to the view and checks access (e.g. using drupal_valid_path()) before doing so

This PR updates the access callback so that one of the arguments passed to _og_context_views_page_access () will be a path component number, which will be replaced with the View's argument at run time. This allows _og_context_views_page_access() to receive the View's argument instead of the path position number entered by the user.


claudiu-cristea:

PR #142, opened 2016-08-17
Inject EntityTypeManager in the MembershipManager

pfrenssen:
Now that we have the `MembershipManager` service, let's inject `EntityTypeManager` into it so we don't have to fall back to the static methods in `\Drupal` and are fully decoupled.


amitaibu:
Thanks

PR #143, opened 2016-08-17
Inject the MembershipManager in OgDeleteOrphansBase

pfrenssen:
Now that `MembershipManager` is a proper service, we can use dependency injection to add it to `OgDeleteOrphansBase`.


amitaibu:
Cleanups are the best :)

PR #147, opened 2016-08-23
WIP: prevent nonmember role when creating new memberships

dakala:
PR for issue #293. Marking as WIP as it'd be nice to add a unit test for this scenario.

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -320,6 +320,7 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
     }
 
     foreach ($this->getRoles() as $role) {
+      /** @var \Drupal\og\Entity\OgRole $role */
       if ($role->getName() == OgRoleInterface::ANONYMOUS) {
         throw new \LogicException('Cannot save an OgMembership with reference to a non-member role.');
       }
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -191,10 +191,7 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
    * {@inheritdoc}
    */
   public function getRoles() {
-    // Add the member role.
-    $roles[] = Og::getRole($this->getGroupEntityType(), $this->getGroup()->bundle(), OgRoleInterface::AUTHENTICATED);
-    $roles = array_merge($roles, $this->get('roles')->referencedEntities());
-    return $roles;
+    return $this->get('roles')->referencedEntities();
   }
 
   /**
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -297,8 +297,8 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
     // This will watch actual empty values and '0'.
     if (!$this->get('uid')->target_id) {
       // Throw a generic logic exception as this will likely get caught in
-      // \Drupal\Core\Entity\Sql\SqlContentEntityStorage::save and turned in an
-      // EntityStorageException anyway.
+      // \Drupal\Core\Entity\Sql\SqlContentEntityStorage::save and turned into
+      // an EntityStorageException anyway.
       throw new \LogicException('OG membership can not be created for an empty or anonymous user.');
     }
 
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -323,6 +323,9 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
       if ($role->getName() == OgRoleInterface::ANONYMOUS) {
         throw new \LogicException('Cannot save an OgMembership with reference to a non-member role.');
       }
+      if ($role->getName() == OgRoleInterface::AUTHENTICATED) {
+        throw new \LogicException('Cannot save an OgMembership with reference to a member role.');
+      }
     }
 
     parent::preSave($storage);
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -319,6 +319,12 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
       throw new \LogicException(sprintf('Entity type %s with ID %s is not an OG group.', $entity_type_id, $group->id()));
     }
 
+    foreach ($this->getRoles() as $role) {
+      if ($role->getName() == OgRoleInterface::ANONYMOUS) {
+        throw new \LogicException('A group member role is required for creating a membership.');
+      }
+    }
+
     parent::preSave($storage);
   }
 
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -321,7 +321,7 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
 
     foreach ($this->getRoles() as $role) {
       if ($role->getName() == OgRoleInterface::ANONYMOUS) {
-        throw new \LogicException('A group member role is required for creating a membership.');
+        throw new \LogicException('Cannot save an OgMembership with reference to a non-member role.');
       }
     }
 
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -323,7 +323,7 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
       if ($role->getName() == OgRoleInterface::ANONYMOUS) {
         throw new \LogicException('Cannot save an OgMembership with reference to a non-member role.');
       }
-      if ($role->getName() == OgRoleInterface::AUTHENTICATED) {
+      elseif ($role->getName() == OgRoleInterface::AUTHENTICATED) {
         throw new \LogicException('The OgMembership should not be saved along with the "member" role, as the role will be automatically retrieved when asking for the membership\'s roles. Having an OgMembership with an active or pending state already implies the user is a group member.');
       }
     }
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -191,7 +191,9 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
    * {@inheritdoc}
    */
   public function getRoles() {
-    return $this->get('roles')->referencedEntities();
+    $roles[] = Og::getRole($this->getGroupEntityType(), $this->getGroup()->bundle(), OgRoleInterface::AUTHENTICATED);
+    $roles = array_merge($roles, $this->get('roles')->referencedEntities());
+    return $roles;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/Entity/OgMembership.php
```diff
@@ -319,6 +319,12 @@ public function preSave(EntityStorageInterface $storage) {
       throw new \LogicException(sprintf('Entity type %s with ID %s is not an OG group.', $entity_type_id, $group->id()));
     }
 
+    foreach ($this->getRoles() as $role) {
+      if ($role->getName() == OgRoleInterface::ANONYMOUS) {
+        throw new \LogicException('A group member role is required for creating a membership.');
```
amitaibu: Maybe change the name to something more explicit like `Cannot save an OgMembership with reference to a non-member role.`

amitaibu:
@dakala thanks for starting this. Seems you PR also finds some error in a test. Would be indeed great adding a Unit test for this - would you like to take a stab at it? Unit tests looks scary at first, but they are fun

amitaibu:
([Here's](https://github.com/Gizra/og/blob/c42ca5f5e114b75fd03f0456de4c6e2ba72a3a57/tests/src/Kernel/Access/WikiTest.php#L131) the place in the test that causes the exception)

dakala:
@amitaibu: I'll take a look at the broken test and adding a unit test for this scenario too.

amitaibu:
Thinking of it, also the member role should be prevented as well. It is
like in drupal core - you don't save the authenticated role per user.

dakala:
@amitaibu: On second thoughts after the last commit, I don't get why OG shouldn't differ, no matter how slight, from core. A logged in user is automatically recognised as belonging to the AUTHENTCATED role. In OG, if we can't assign a user the MEMBER role, how do we identify basic members of a group? How do we then differentiate between NON-MEMBER and MEMBER roles for group members? It's quite possible I'm completely off the mark! 

Re: [WikiTest](https://github.com/Gizra/og/blob/c42ca5f5e114b75fd03f0456de4c6e2ba72a3a57/tests/src/Kernel/Access/WikiTest.php#L131). I guess the roles being loaded didn't exist so I've created them in the hope that's what we're after.

amitaibu:
> if we can't assign a user the MEMBER role, how do we identify basic members of a group? 

By the fact they _have_ an `OgMembership`. Your question makes me think that maybe throwing an Exception when saved with the member role is a bit too strict, but maybe we can improve the exception text. Something like:

`The OgMembership should not be saved along with the "member" role, as the role will be automatically retrieved when asking for the membership's roles. Having an OgMembership with an active or pending state already implies the user is a group member.`

dakala:
That's much clearer now, I change the text accordingly.

By the way how did you find out that line 131 was the culprit in WikiTest? Thanks.

amitaibu:
> That's much clearer now

Awesome. It's always great to have fresh new pair of eyes.

> By the way how did you find out that line 131 was the culprit in WikiTest?

I saw on Travis where it [failed](https://travis-ci.org/Gizra/og/jobs/154350753#L595)

tests/src/Kernel/Access/WikiTest.php
```diff
@@ -129,12 +130,14 @@ public function setUp() {
...
-      $role = $this->container->get('entity_type.manager')
-        ->getStorage('og_role')
-        ->load($role_id);
-      $role->grantPermission('edit any group_content content');
+      $role = OgRole::create();
+      $role
+        ->setGroupType('block_content')
+        ->setGroupBundle('group')
+        ->setName($role_name)
+        ->setLabel("Block content for group. Role: $role_name")
```
amitaibu: This can be random.

tests/src/Kernel/Access/WikiTest.php
```diff
@@ -129,12 +130,14 @@ public function setUp() {
 
     // Grant both members and non-members permission to edit any group content.
     foreach ([OgRoleInterface::AUTHENTICATED, OgRoleInterface::ANONYMOUS] as $role_name) {
```
amitaibu: Checking we throw an exception can be added into the existing Kernel tests (we could have done it in Unit, but it's already written as Kernel). See [this](https://github.com/Gizra/og/blob/438869c2928e6eed4dc0dc005c92dcca87cb5b92/tests/src/Kernel/Entity/OgMembershipTest.php#L126) for example, as a handy way to tell that an exception is expected.

dakala:
Not sure we can use @expectedException in setUp().

Checked the logs and there's a bunch of tests (Functional and Kernel) where the "you can't add member role" logic exception is getting thrown. Shouldn't those tests be revisited in addition to adding @expectedException annotations? 

Thanks for your help!

amitaibu:
Those tests should be adapted to no longer include the member/non-member role. No need to check for exception in multiple places - each test is responsible to assert its own functionality.

> Thanks for your help!

Thank you for working on this :)

src/Entity/OgMembership.php
```diff
@@ -319,6 +319,15 @@ public function preSave(EntityStorageInterface $storage) {
       throw new \LogicException(sprintf('Entity type %s with ID %s is not an OG group.', $entity_type_id, $group->id()));
     }
 
+    foreach ($this->getRoles() as $role) {
+      if ($role->getName() == OgRoleInterface::ANONYMOUS) {
+        throw new \LogicException('Cannot save an OgMembership with reference to a non-member role.');
+      }
+      if ($role->getName() == OgRoleInterface::AUTHENTICATED) {
```
amitaibu: => elseif

tests/src/Kernel/Access/WikiTest.php
```diff
@@ -60,6 +61,7 @@ class WikiTest extends KernelTestBase {
 
   /**
    * {@inheritdoc}
+   * @expectedException \LogicException
```
amitaibu: lets dedicate own method (e.g. `testDefaultRolesSave`) to assert the exception, and keep the setup() error free

dakala:
- [ ] Add testDefaultRolesSave to assert LogicException

amitaibu:
@dakala are you still on it?

dakala:
@amitaibu: oops! Thought I was done. I'm on it. Will push later. Sorry

dakala:
@amitaibu: I replaced testDefaultRoles with 2 tests - one for saving OgMembership with non-member role and the other with member role and assert the LogicException for both.

I don't fully get what we're checking with the 2 membership states, though, as we're only interested in OgRole and OgMembership regardless of other properties of the membership.

Thanks for helping!

tests/src/Kernel/Access/WikiTest.php
```diff
@@ -184,6 +161,66 @@ public function testEntityOperationAccess($user, array $expected_results) {
...
+      ->setGroupType('block_content')
+      ->setGroupBundle('group')
+      ->setName(OgRoleInterface::AUTHENTICATED)
+      ->setLabel($this->randomString())
+      ->grantPermission('edit any group_content content');
+    $role->save();
+
+    // Subscribe the normal member and the blocked member types to the group.
+    foreach (['member', 'blocked'] as $membership_type) {
+      $state = $membership_type === 'member' ? OgMembershipInterface::STATE_ACTIVE : OgMembershipInterface::STATE_BLOCKED;
```
amitaibu: the state isn't important here - the only thing we care about is to make sure you cannot `addRole()` with the member/non-meber roles

amitaibu:
Thanks for picking this back. There are now a few failing tests (due to the fact we didn't impose this restriction). Could you check them please?

dakala:
Yeah,just going through the Travis logs. I need to check my fork as I haven't got some tests from the main Gizra repo.

tests/src/Kernel/Access/WikiTest.php
dakala: I thought as much.

dakala:
@amitaibu: A couple of tests are wrongly namespaced e.g. tests/src/Kernel/Access/OgAccessHookTest.php so other branches with such tests shouldn't be passing. Which of your branches have all tests passing please? I'd like to compare with mine. I suspect a good number of tests may be broken in other branches.

amitaibu:
https://github.com/Gizra/og branch `8.x-1.x` is the canonical one.

src/Entity/OgMembership.php
```diff
@@ -191,10 +191,7 @@ public function revokeRole(OgRole $role) {
    * {@inheritdoc}
    */
   public function getRoles() {
-    // Add the member role.
```
amitaibu, line 194: we need this for the getting the permissions. So, maybe another option, is making sure in OgMembership::preSave to strip out the auth role?

dakala:
@amitaibu: How about this? 

Since we're disallowing the creation of memberships with non-member role, I'm not sure how the wiki functionality's going to be implemented. So I've left out the test for the ability of non-members to edit any content in a Wiki-style group. Any help would be appreciated :-)

amitaibu:
Thanks, I'll have a look

amitaibu:
@dakala Can you remind me why we don't throw an exception on Member. I remember there was a reason, but I don't remember (which means it should probably be added in the comments)

dakala:
Hi @amitaibu. I'm not sure we need this branch as we'll be working on your repo from now on.

amitaibu:
Sure, you can re-open a new PR.

dakala:
We had this as the exception message before: 

_The OgMembership should not be saved along with the "member" role, as the role will be automatically retrieved when asking for the membership's roles. Having an OgMembership with an active or pending state already implies the user is a group member._

It should probably be what we need to add to the comments.

amitaibu:
Oh Right, it's because `OgMembership::getRoles` returns it, so if you resave, we don't want to get an exception.

dakala:
That's right. I only found out the connection when I removed the code that retrieves the "member" role in OgMembership::getRoles  and you pointed it out to me.

dakala:
Do we have issue priorities? Maybe I can pick up something new from https://github.com/amitaibu/og/issues

amitaibu:
Can you open a new PR please with the branch, so we won't forget it.

Yes, you can tackle the ones in https://github.com/amitaibu/og/issues (which anyway should be moved here at a certain point, so they get more visibility). If you don't find anything, let me know - I can always come up with a nice task :)

oh, btw - I just published [this post](http://www.gizra.com/content/og8-development-mindset/) you might find interesting...

dakala:
Nice article. In the graphic, not sure about why we've got 2 "Group" circles. I thought "Membership" is related to "Group" the same way as "Group content". I'm sure the more I understand the code, the architecture'll be much clearer.

amitaibu:
In og7 membership is related to group and group content (and user is
created kind of a group content)

In og8 we change it, and membership is just to users (as users are really a
special case, they are not content. By doing so we simplified the
architecture quite a bit

dakala:
I see. You had 2 different "Group" entities in your diagram. Is something like this a fair representation?

og8

Maybe we should start using the Wiki for jotting down ideas for documentation?

amitaibu:
Yes, exactly. I thought (and maybe I was wrong), that separating to two would be clearer.

In terms of docs - I think we could start with the README. What do you think?

dakala:
Yeah, I think having a "Group" with both "Membership" and "Group Content" makes it clearer. That's with the "Membership" entity serving as the bridge to the "User"

Agreed. README is a good start.

PR #149, opened 2016-08-24
Group tab

amitaibu:
Here we'll be working on the group tab functionality, and continue work started in https://github.com/amitaibu/og/pull/218

members___site-install


amitaibu:
Update: Currently I'm mostly moving files around, and starting to improve the API, so nothing should be working yet. But it will get there :)

amitaibu:
Update: We went a little back in terms of functionality, but I've removed some of the flexibility to get a more simplified and easier to understand flow. This is starting to get into better shape, but still more work need to be done..

amitaibu:
Things that I plan to do next:
1. Remove the Plugins in favor of `EventSubscriber`. We don't really use the plugins as plugins, they are like a `hook_info()` used for whenever we `alterRoutes()`
2. Re-add the menu links, so the `Group` tab will appear once again
3. Use `SystemController::overview` to show all the routes

amitaibu:
Update: Plugins are out EventSubscriber are in, and we're down to less files and less complexity.

@pfrenssen I have a routing question you might know the answer to:
1. We have an [OG permission check](https://github.com/Gizra/og/blob/438869c2928e6eed4dc0dc005c92dcca87cb5b92/og.routing.yml#L28) we can use in routing.yml files, that is expecting `{entity_type_id}/{entity_id}` in the route's path. Here's the [GroupCheck](https://github.com/Gizra/og/blob/438869c2928e6eed4dc0dc005c92dcca87cb5b92/src/Access/GroupCheck.php#L61) class.
2. In this PR we define `$entity_type->setLinkTemplate('og-group-admin-pages', "/$entity_type_id/{{$entity_type_id}}/group");` so the route's name will be for example `/node/{node}/group`

My question is - can we re-map the path to the access callback, so it will know `node == {entity_type_id}` and `{node} == {entity_type}`?

src/Controller/OgAdminRoutesController.php
```diff
@@ -46,7 +48,15 @@ public static function create(ContainerInterface $container) {
    * @return mixed
    *   List of available admin routes for the current group.
    */
-  public function overview($entity_type_id) {
+  public function overview(RouteMatchInterface $route_match) {
+
+    $parameter_name = $route_match->getRouteObject()->getOption('_og_entity_type_id');
```
amitaibu, line 66: I was expecting that the [{node}](https://github.com/Gizra/og/pull/149/commits/1a7be5529f8fa8a0b4b224fe15880d0ce06656f0#diff-94902c9924afa2fe352c6170db324642R88) will be injected into controller, but for some reason we need to do this hack.

@weitzman do you happen to know the reason why we need this hack? (I copied this pattern from Devel module.)

amitaibu:
Update: The routes/ local tasks wiring is now in - so next is re-adding a simplified membership views, cleanups and tests.

amitaibu:
Update: A simple View is in place:

members___site-install

We'll need also to make sure that `GroupManagerType::addGroup` will request a rebuild cache - as the routes will need to be added

RudyMartin:
Awesome!

amitaibu:
@pfrenssen never mind the routing question from above, I found a solution 😄 https://github.com/Gizra/og/pull/149/commits/31b6fc2cfea4148864b39c2c842e26caae4817f2

amitaibu:
I'm still polishing and going to add automatic tests, but can already be early-reviewed.

amitaibu:
Last commit https://github.com/Gizra/og/pull/149/commits/d22b0d1affc89a05f2bc270151bf9069985c9c7f has a failing Unit test. I've opened a core [issue](https://www.drupal.org/node/2791071), to see how we can mock `Url` in unit tests.

amitaibu:
Unit test were added (We're no longer using `Url::access()`, but rather the `accessManager` directly). I love those unit tests!

Few more tests to go... 😄

amitaibu:
Update: Functional test added to assert Group tab existance.

amitaibu:
Next I'm going to add a Kernel test to assert the `members` table

amitaibu:
Ready for review! 😄

amitaibu:
Thanks @RoySegall for the great initial work, this is now in.

PR #163, opened 2016-09-04
Og membership state cache context

amitaibu:
illum_wisi___site-install

### DIFF
--- a/src/Cache/Context/OgMembershipStateCacheContext.php
+++ b/src/Cache/Context/OgMembershipStateCacheContext.php
@@ -47,8 +47,6 @@ class OgMembershipStateCacheContext implements CacheContextInterface {
    */
   protected $user;
 
-
-
   /**
    * Constructs a new UserCacheContextBase class.
    *
--- a/src/Cache/Context/OgMembershipStateCacheContext.php
+++ b/src/Cache/Context/OgMembershipStateCacheContext.php
@@ -9,6 +9,7 @@ use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\og\GroupTypeManager;
 use Drupal\og\MembershipManagerInterface;
+use Drupal\og\OgMembershipInterface;
 
 /**
  * Defines a cache context service, for "membership state" caching.
@@ -76,14 +77,35 @@ class OgMembershipStateCacheContext implements CacheContextInterface {
    * {@inheritdoc}
    */
   public function getContext() {
-    if (!$route_object = $this->routeMatch->getRouteObject()) {
+    if (!$route_contexts = array_keys($this->routeMatch->getRouteObject()->getOption('parameters'))) {
+      // No "parameters" defined in the route.
       return 'none';
     }
 
-    if (!$route_contexts = $route_object->getOption('parameters')) {
+    if ($entity_type_ids = array_keys($this->groupTypeManager->getAllGroupBundles())) {
+      // No group entities.
       return 'none';
     }
 
+    if (!$entity_type_ids = array_intersect($route_contexts, $entity_type_ids)) {
+      // No parameters that match the group entities.
+      return 'none';
+    }
+
+    // Take just the first entity type ID.
+    $entity_type_id = reset($entity_type_ids);
+
+
+    $group = $this->routeMatch->getParameter($entity_type_id);
+    $states = [
+      OgMembershipInterface::STATE_ACTIVE,
+      OgMembershipInterface::STATE_PENDING,
+      OgMembershipInterface::STATE_BLOCKED,
+    ];
+
+    /** @var OgMembershipInterface $memebrship */
+    $memebrship = $this->membershipManager->getMembership($group, $this->user, $states);
+    return $memebrship ? $memebrship->getState() : 'none';
   }
 
   /**
--- a/src/Cache/Context/OgMembershipStateCacheContext.php
+++ b/src/Cache/Context/OgMembershipStateCacheContext.php
@@ -75,17 +75,17 @@ class OgMembershipStateCacheContext implements CacheContextInterface {
    * {@inheritdoc}
    */
   public function getContext() {
-    if (!$route_contexts = array_keys($this->routeMatch->getRouteObject()->getOption('parameters'))) {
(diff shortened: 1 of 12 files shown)


amitaibu:
Update: I'm still seeing `X-Drupal-Dynamic-Cache:MISS`, so need to figure out what is the cause (not necessarily OG in this case)

amitaibu:
Things I need to figure (but pointers would be appreciated of course):
1. I see a cache miss also when same auth user viewes a non-Og node
2. I don't see the return value under the cache tags. I would expect to see `og_membership_state:none` or `og_membership_state:active`
3. How can I see which cache context caused the invaldiation

pfrenssen:
Following this with great interest. I'm going to try to get some time assigned to review this.

amitaibu:
@pfrenssen hi! In the mean time @DavidHernandez and me were able to add unit tests :)

src/Cache/Context/OgMembershipStateCacheContext.php
```diff
@@ -0,0 +1,117 @@
...
+    return t('OG membership state');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContext() {
+    if (!$route_contexts = $this->routeMatch->getRouteObject()->getOption('parameters')) {
+      // No "parameters" defined in the route.
+      return 'none';
```
pfrenssen: Can we use a constant instead of hardcoding `none`?

src/Cache/Context/OgMembershipStateCacheContext.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (!$entity_type_ids = array_intersect(array_keys($route_contexts), $entity_type_ids)) {
+      // No parameters that match the group entities.
+      return 'none';
+    }
+
+    // Take just the first entity type ID.
+    $entity_type_id = reset($entity_type_ids);
```
pfrenssen, line 101: This won't work if there are groups using multiple entity types. In our project for example we have groups of type `node`, but also groups of type `rdf_entity`.

src/Cache/Context/OgMembershipStateCacheContext.php
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContext() {
+    if (!$route_contexts = $this->routeMatch->getRouteObject()->getOption('parameters')) {
+      // No "parameters" defined in the route.
+      return 'none';
+    }
```
pfrenssen, line 88: Depending directly on the route parameters severely limits the usefulness of this cache context. It will for example not be possible to use this on group content pages, because they won't match the route of the group to which they belong.

An example: in our project all group content has a header that displays the group to which the content belongs. In this header we have some local actions the user can take (for example join the group if they are not a member yet). So we need to vary by membership state cache context on our group content pages.

I am wondering if we can add the current group to the generic route context.

pfrenssen, line 88: It is possible to add a 'runtimeContext' to the route (ref `ContextProviderInterface::getRuntimeContexts()`) - here we can dynamically set the right group that matches the current route. It should be possible to do this plugin based like was already being worked on.

We should use this runtimeContext to determine the cache context.

amitaibu: Thanks for the pointer on `:: getRuntimeContexts`, I'll check it out

> It should be possible to do this plugin based like was already being worked on.

My recent thought about OG context is, for the sake of simplicity not port along with Plugins, but rather have a single ContextProvider.

So maybe, In your case, instead of the plugins you could extend the cache context and add your logic. How does it sound?

src/Cache/Context/OgMembershipStateCacheContext.php
amitaibu: It takes the first entity type ID that is a group (see the `array_intersect` above)

src/Cache/Context/OgMembershipStateCacheContext.php
pfrenssen: Well now it only works on a very specific subset of routes that follow very specific rules (like being formatted as `node/%`). By leveraging the OG context it would work out of the box on all routes that have a group associated with them. That sounds more useful to me. It would of course also work on `node/%`).

I will have to look into this though, I start to understand caching a bit better lately but I am still unsure about some aspects of it.

amitaibu, line 88: Yeah, OG context might be needed after all. /cc @RoySegall

RoySegall: Should I get back to the original PR and try to close it? I'm a bit busy with the DrupalCamp site at the past week but let me know what's the PRs priority.

amitaibu: Actually I think the case is simpler. The cache context is added only in the formater of a field attached directly to a group (subscribe to group). So getting the group in this case is  tightly coupled with the route, thus og context would not be needed.

amitaibu: (Although in general,  and not related to this pr, og context might still be needed - not sure)

amitaibu: @pfrenssen ^^

pfrenssen: This field-specific cache context would work fine for this use case, but it would be obsolete when we get the generic cache context that uses OgContext, since that will also cover its use case. So this implementation is temporary at best.

In our project we now need the generic cache context for a group header that shows some information about the group. This information is different depending on the membership status so we need this cache context.

I can work on this now. I'll have a look at @RoySegall 's ticket, maybe I can help there to move things forward.

amitaibu: > but it would be obsolete when we get the generic cache context that uses OgContext,

Indeed, I think we can have it as a follow up, once we have `OgContextProvider` in place

amitaibu:
I've disabled development mode on my local, and was able to see a `cache: HIT`.

amitaibu:
Oh, I thought I got it in already. Fixed the sniffer mistakes. Once green, I'll merge.

PR #157, opened 2016-09-01
Prevent duplicate OgMembership

dakala:
https://github.com/amitaibu/og/issues/259

Marked as WIP as no tests yet.

PR #164, opened 2016-09-07
Prevent duplicate OgMembership

amitaibu:
https://github.com/amitaibu/og/issues/259
#157

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -320,22 +320,24 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
     }
 
     // Check for an existing membership.
+    $query = \Drupal::entityQuery('og_membership');
+    $query
+      ->condition('uid', $this->get('uid')->target_id)
+      ->condition('entity_id', $this->get('entity_id')->value)
+      ->condition('entity_type', $this->get('entity_type')->value);
+
     if (!$this->isNew()) {
+      // Filter out this membership.
+      $query->condition('id', $this->id(), '<>');
+    }
 
-      $query = \Drupal::entityQuery('og_membership');
-      $count = $query
-        ->condition('uid', $this->get('uid')->target_id)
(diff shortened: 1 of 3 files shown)


src/Entity/OgMembership.php
```diff
@@ -319,6 +319,25 @@ public function preSave(EntityStorageInterface $storage) {
...
+    // Check for an existing membership.
+    if (!$this->isNew()) {
+
+      $query = \Drupal::entityQuery('og_membership');
+      $count = $query
+        ->condition('uid', $this->get('uid')->target_id)
+        ->condition('entity_id', $this->get('entity_id')->value)
+        ->condition('entity_type', $this->get('entity_type')->value)
+        // Filter out this membership.
+        ->condition('id', $this->id(), '<>')
```
amitaibu: @dakala fyi, here's a better solution - we do a `COUNT` query, and make sure to filter out the `$this->id()` (it was missing in your previous logic)

amitaibu:
Waiting for Travis, this should be good to go.

amitaibu:
Actually, sorry, I seem to have missed something - checking..

src/Entity/OgMembership.php
```diff
@@ -319,6 +319,27 @@ public function preSave(EntityStorageInterface $storage) {
...
     }
 
+    // Check for an existing membership.
+    $query = \Drupal::entityQuery('og_membership');
+    $query
+      ->condition('uid', $this->get('uid')->target_id)
+      ->condition('entity_id', $this->get('entity_id')->value)
+      ->condition('entity_type', $this->get('entity_type')->value);
+
+    if (!$this->isNew()) {
```
amitaibu, line 329: @dakala ok, sorry - no the query is right. Good we have tests in place :)

amitaibu:
Exception caught a test that tried to create a 2nd permission. Hopefully this should be green now.

amitaibu:
Thanks @dakala !

PR #166, opened 2016-09-08
Add test to check that membership has member role

dakala:


tests/src/Kernel/Entity/OgMembershipTest.php
```diff
@@ -218,6 +219,24 @@ public function testGetSetState($state, $method) {
   }
 
   /**
+   * Tests that membership has "member" role when roles are retrieved.
+   *
+   * @covers ::getRoles
+   */
+  public function testGetRoles() {
```
amitaibu: maybe better `testMemberRole()`

amitaibu:
Nice one. Minor comment, and I'll merge

amitaibu:
Thanks!

PR #173, opened 2016-09-10
Streamline testing on Travis CI

pfrenssen:
This is a followup and cleanup of my experiment with using Scrutinizer CI for coding standards checks in https://github.com/amitaibu/og/pull/301

I was unhappy about a number of things with our current implementation of testing on Travis:
- Test dependencies were being installed in hard coded `before_script` instructions, instead of declaring our dev dependencies in `composer.json` and leveraging that.
- Some magic was done by downloading Coder and symlinking its coding standards inside PHP_CodeSniffer before executing the coding standards checks. The standard way of doing this is to declare our coding standards in a ruleset, and configurating PHP_CodeSniffer with a `phpcs.xml` file. This can then also be used by developers working on the module by simply running `./vendor/bin/phpcs` from the module folder.
- The coding standards checks were running before the "real" tests, delaying the important results.  Also having a coding standards violation would fail every test. These should run in a separate instance.
- The coding standards checks were running for all different PHP versions and all different versions of core, this is overkill, we only need to check it once.

I have experimented by splitting the work between Travis CI and Scrutinizer CI. Scrutinizer CI is dedicated to static analysis, and it returns some useful suggestions, but mainly a whole bunch of false positives. The things it finds are quite minor, and this does not weigh up against the difficulty of setting it up, and the burden of having to maintain a long list of false positives. I have abandoned this idea, but I have made a commit with the improvements that were suggested by the service, so this work is not lost.

I also tried out the ContinuousPHP service. This was very easy to set up, I had the coding standards checks set up and running in 5 minutes, but this service unfortunately requires to set up pipelines manually through a web UI. It does not yet support a `.continuousphp.yml` configuration file (it is promised for the future though).

So I finally reworked the Travis CI configuration to run the coding standards checks in a separate process. I have also done some major cleanup to the config file. I have removed the deprecated PHP 5.5 version, and have added Drupal 8.3.x which is now open for development.

I have added a small script to run the Travis CI tests. I am personally not a big fan of using shell scripts for CI, but this is recommended by Travis: the exit error codes of the test scripts are piped through and fail or pass the build. If the bash script is unwanted I can work this functionality in the main Travis configuration file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,5 +9,7 @@
         "irc": "irc://irc.freenode.org/drupal-og",
         "source": "https://cgit.drupalcode.org/og"
     },
-    "minimum-stability": "dev"
+    "require-dev": {
+        "drupal/coder": "^8.2"
+    }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,20 @@ language: php
 sudo: false
 
 php:
-  - 5.5
   - 5.6
   - 7.0
 
 env:
-  - DRUPAL_CORE=8.1.x
-  - DRUPAL_CORE=8.2.x
+  - TEST_SUITE=8.1.x
+  - TEST_SUITE=8.2.x
+  - TEST_SUITE=8.3.x
+  - TEST_SUITE=PHP_CodeSniffer
+
+# Only run the coding standards check once.
+matrix:
+  exclude:
+    - php: 5.6
+      env: TEST_SUITE=PHP_CodeSniffer
 
 mysql:
   database: og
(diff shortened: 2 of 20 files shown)


composer.json
```diff
@@ -9,5 +9,7 @@
         "irc": "irc://irc.freenode.org/drupal-og",
         "source": "https://cgit.drupalcode.org/og"
     },
-    "minimum-stability": "dev"
+    "require-dev": {
+        "drupal/coder": "^8.2"
```
amitaibu, line 13: maybe we should have a fixed version?

amitaibu, line 13: anyway, it can be in a follow up.

amitaibu:
Merged, thanks!

composer.json
pfrenssen: This should follow stable releases, meaning it will be "fixed" for 3-4 months in between Coder releases. When a new release is out we will be instantly alerted ;)

I prefer this over having to manually check if there is a new Coder release out every day, but it's up to you if you want to fix it to a specific version.

amitaibu: np, we can keep as-is

PR #165, opened 2016-09-07
Prevent OgMemberships to be created with the 'non-member' and 'member' role 

dakala:
Continuation of #293 from https://github.com/Gizra/og/pull/147

PR #174, opened 2016-09-10
Prevent OgMemberships to be created with the 'non-member' and 'member' role

dakala:
Re-rolled https://github.com/Gizra/og/pull/165

### DIFF
--- a/tests/src/Kernel/Access/AccessByOgMembershipTest.php
+++ b/tests/src/Kernel/Access/AccessByOgMembershipTest.php
@@ -16,11 +16,11 @@ use Drupal\user\Entity\Role;
 use Drupal\user\Entity\User;
 
 /**
- * Tests Wiki-style access: non-members have the right to edit any content.
+ * Tests access to content by OgMembership.
  *
  * @group og
  */
-class WikiTest extends KernelTestBase {
+class AccessByOgMembershipTest extends KernelTestBase {
 
   use ContentTypeCreationTrait;
 
--- a/tests/src/Kernel/Access/AccessByOgMembershipTest.php
+++ b/tests/src/Kernel/Access/AccessByOgMembershipTest.php
@@ -8,9 +8,11 @@ use Drupal\block_content\Entity\BlockContentType;
 use Drupal\node\Entity\Node;
 use Drupal\node\Entity\NodeType;
 use Drupal\og\Entity\OgMembership;
+use Drupal\og\Entity\OgRole;
 use Drupal\og\Og;
 use Drupal\og\OgGroupAudienceHelper;
 use Drupal\og\OgMembershipInterface;
+use Drupal\og\OgRoleInterface;
 use Drupal\simpletest\ContentTypeCreationTrait;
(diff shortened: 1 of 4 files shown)


tests/src/Kernel/Access/WikiTest.php
```diff
@@ -204,15 +225,6 @@ public function entityOperationAccessProvider() {
         ],
       ],
       [
-        // Non-members should have the right to edit any group content.
```
amitaibu, line 207: @dakala you have removed this part for now, just because the tests failed, right?

dakala: @amitaibu Not just now. I removed it because the old test assigned a non-member role to a membership. The member role is assigned to a membership automatically so that's not a problem here.

I mentioned some time ago I didn't know how to make the Wiki functionality work for a core authenticated user. Once I understand how the wiki works (i.e. what enables wiki on a piece of content) I can write a test for the "non-member" role - a core authenticated use r who isn't a member of the group.

amitaibu: There's isn't really `wiki` feature - the name of the file can change. It's just about checking different scenarios of permissions. So `wiki` refers here to the fact that a non-member can edit existing group content.

dakala:
Ok, let's rename it. That name and functionality made sense when an OgMembership could have a "non-member" role. Shall we say `OgMembershipRoleTest`? Or `OgMembershipDefaultRoles`? Anything more fitting?

amitaibu:
`AccessByOgMembershipTest`?

tests/src/Kernel/Access/AccessByOgMembershipTest.php
```diff
@@ -204,15 +225,6 @@ public function entityOperationAccessProvider() {
         ],
       ],
       [
-        // Non-members should have the right to edit any group content.
```
amitaibu, line 207: Ok, I'll give this a look now

tests/src/Kernel/Access/AccessByOgMembershipTest.php
```diff
@@ -129,15 +130,16 @@ public function setUp() {
...
     foreach ([OgRoleInterface::AUTHENTICATED, OgRoleInterface::ANONYMOUS] as $role_name) {
-      $role_id = "block_content-group-$role_name";
-      /** @var \Drupal\og\Entity\OgRole $role */
-      $role = $this->container->get('entity_type.manager')
-        ->getStorage('og_role')
-        ->load($role_id);
-      $role->grantPermission('edit any group_content content');
-      $role->save();
+      /** @var OgRole $role */
+      $role = OgRole::getRole('block_content', 'group', $role_name);
```
amitaibu, line 134: @dakala I've re-added saving the roles for non-members as-well (and even used the new `OgRole::getRole`). The error was that a few lines below we didn't re-load the $role for the auth user, thus saved the membership with the wrong role.

amitaibu:
Merged, thanks.

PR #176, opened 2016-09-13
What is a user based entity?

dakala:
https://github.com/amitaibu/og/issues/220

Fixed description of field type


amitaibu:
Thanks

PR #178, opened 2016-09-14
Set accessCheck() to FALSE on \Drupal\og\MembershipManager::getGroupIds

chx:


src/MembershipManager.php
```diff
@@ -189,6 +189,9 @@ public function getGroupIds(EntityInterface $entity, $group_type_id = NULL, $gro
       $query = $this->entityTypeManager
         ->getStorage($target_type)
         ->getQuery()
+        // When rebuilding node_access grants then checking node_access is
```
amitaibu: Maybe: `// Disable entity access so fetching the groups related to group content are not affected by the current user.`

chx: The current user matters less -- when rebuilding node access and your groups are nodes then you are trying to retrieve node access records which do not exist because node_access_rebuild have erased the table in the first place...

amitaibu: I think it actually solves both cases. So maybe combine them, something like:

`// Disable entity access so fetching the groups related to group content are not affected by the current user. Furthermore, when rebuilding node access and the groups are nodes, we should not try to retrieve node access records which do not exist because the rebuild process has already erased the grants table.`

amitaibu: one more thought - I think we can change the function signature to `public function getGroupIds(EntityInterface $entity, $group_type_id = NULL, $group_bundle = NULL, $check_access = FALSE) {`

amitaibu:
> because this parameter changes the return value and usage quite a bit

Fair enough, I'll revert the change (but keep the extended comment).

chx:
Can we chat on the comment somewhere first? skype? irc (pm me)?

amitaibu:
Sure, I'll log to drupal-contribute

amitaibu:
@chx ^^ (I'll be available for the next ~10 min - got a movie after that 😄 )

amitaibu:
@chx can you share why you think the current comment is wrong?

PR #180, opened 2016-09-15
Port OG-Context module as OgContext provider

pfrenssen:
This is a continuation of https://github.com/amitaibu/og/pull/252. It is a simplified version of the initial implementation of OgContext that was created in that PR.


src/Plugin/OgGroupResolver/Entity.php
```diff
@@ -0,0 +1,91 @@
...
+   * {@inheritdoc}
+   */
+  public function getGroup() {
+    foreach ($this->routeMatch->getParameters() as $parameter) {
+      if (!($parameter instanceof ContentEntityBase)) {
+        continue;
+      }
+
+      if ($this->groupTypeManager->isGroup($parameter->getEntityTypeId(), $parameter->bundle())) {
+        return $parameter;
```
claudiu-cristea: The interface states that ::getGroups() should return EntityInterface[] but here returns a single item. Shouldn't be `return [$parameter];`?

src/OgGroupResolverInterface.php
```diff
@@ -0,0 +1,67 @@
...
+   */
+  public function getBestCandidate();
+
+  /**
+   * Declares that no further group resolving is necessary.
+   *
+   * Use this if the plugin has determined the relevant group in the current
+   * context with 100% certainty.
+   */
+  public function stopPropagation();
```
amitaibu, line 55: nice solution - 👍

src/Plugin/OgGroupResolver/Entity.php
```diff
@@ -0,0 +1,91 @@
...
+      }
+
+      if (!Og::isGroupContent($parameter->getEntityTypeId(), $parameter->bundle())) {
+        continue;
+      }
+
+      return Og::getGroups($parameter);
```
amitaibu: I think it's better to inject `MembershipManager` directly (would make unit tests a bit easier)

tests/modules/og_test/src/Plugin/GroupResolver/UserGroupReference.php
```diff
@@ -0,0 +1,29 @@
...
+ * @file
+ * Contains \Drupal\og_test\Plugin\OgResolver\UserGroupReference.
+ */
+
+use Drupal\og\OgContextBase;
+
+/**
+ * Get the group from a field reference attached to the current user.
+ *
+ * @GroupResolver(
```
amitaibu: => `@OgGroupResolver`? (Just because there could be other `group` of different things - happened already in D7 even with the vertical tabs module or something similar)

amitaibu (commented):
Haven't reviewed all, but just a few small comments

src/Plugin/OgGroupResolver/Entity.php
pfrenssen: This is still left from the original approach, this is still adhering to the previous implementation of the interface.

I think this plugin should probably be entirely removed. This is not a reliable way of discovering entities on routes. I am planning on using the entity paths instead.

This code as well as the test code is still to be refactored, it's better to wait a bit with reviewing. I can't manage the labels on this issue, but I'll remove the "WIP" from the title and post a comment  when it is ready to be reviewed.

src/OgGroupResolverInterface.php
pfrenssen: This was an idea from @sandervd! There are some cases where you really don't want other plugins influencing the outcome, for example if you are on the canonical path for a group,  or if you have a dedicated plugin written for a project that handles the resolving in a custom way.

src/Plugin/OgGroupResolver/Entity.php
pfrenssen: I am thinking of replacing this plugin with a more reliable solution. We can get a list of entity paths from the entity definition and check if the current route matches any of these entity paths. This would avoid false positives on routes that have multiple arguments and happen to have a group in one of them.

Imagine you have a search page that allows to filter on related content:

> /search/related/{entity}/{entity}/{entity}

Imagine the 3rd entity is an article that happens to be group content. If we blindly loop over the route parameters then we will suddenly decide that this search results page belongs to the group of the 3rd entity.

This could cause all kinds of unexpected behaviour.

tests/modules/og_test/src/Plugin/GroupResolver/UserGroupReference.php
pfrenssen: :+1:  I had already fixed this but I hadn't yet pushed my changes.

pfrenssen:
@amitaibu in https://github.com/amitaibu/og/pull/252#issuecomment-247079122 you outline the algorithm to resolve the groups:

```
OgContextProvider::getRuntimeContexts() will call the OgGroupResolver plugins, and let them return all their values. That is, it will not break on the first group that is found.

Then, it will apply some simple (hardcoded) rule to get the "Best matching" group, by order:

1. User has that group in the session, because in the previous page requests they go it as context + user has access to it. This is for cases when yuu previously visited a group content that belonged to group A, and now you visit a group content that belongs to groups A and B -- so we should be nice and keep you in the A context.
2. User has view access to the group
3. The first rule
```

The "view access to the group" is a bit problematic in combination with plugins returning their own best candidate. It is possible that a plugin has two potential candidates, but the "best" one is not accessible by the user, but the second best candidate is. I think it would be better that the plugins always return the full set of groups, and simply order them so that the best candidate is the first result. This would mean to get rid of `OgGroupResolverInterface::getBestCandidate()`.

amitaibu:
> The "view access to the group" is a bit problematic in combination with plugins returning their own best candidate

In OG7, and I think we can keep in OG8 for simplicity the `getBestCandidate` should be done once on the top level by the ContextProvider . That is, the plugins will return an _un-ordered_ list of groups. Then it would be the job of `getBestCandidate` to iterate over them, filter out the non-accessible ones, and choose one single group.

pfrenssen:
OK, that sounds like a great approach, I'll rework it to work like this.

pfrenssen:
@amitaibu I realized something else: the session approach won't work out of the box for anonymous users on Drupal 8. For performance reasons anonymous users no longer have a cookie, so they don't have a session.

It is still possible to initiate a session for anonymous users, but it would be highly inappropriate for OG to do this out of the box, since it would make any website that has OG enabled to be uncacheable by reverse proxies such as Varnish.

I'm guessing that for this particular PR it is OK not to worry about this. We can come up with a good solution in the future, e.g. rewriting outgoing links to add query arguments, or allow to opt-in to sessions. I just wanted to mention this. I'll also add a mention of it in the documentation of the user session plugin.

pfrenssen:
This was discussed IRL between @amitaibu and me at Drupalcon. We are going to get rid of the distinction between two "types" of plugins ("providers" vs "selectors"). Instead we will have a single plugin type that can either provide groups, or remove groups that were discovered by previous plugins. This will work by passing an object to the plugins that contains a collection of the previously discovered groups.

This will simplify the approach.

tests/src/Unit/Plugin/OgGroupResolver/OgRouteGroupResolverTestBase.php
```diff
@@ -0,0 +1,317 @@
...
+
+      // It is not being tight lipped about whether it is a group or group
+      // content.
+      $this->groupTypeManager->isGroup($entity_type_id, $bundle_id)
+        ->willReturn($is_group);
+
+      // @todo Provide a mock for a call to ??::isGroupContent() as soon as this
+      //   is available in a service.
+      // $this->groupContentTypeManager??->isGroupContent($entity_type_id, $bundle_id)
+      //  ->willReturn($is_group_content);
```
pfrenssen: I'm using unit tests to test the plugins, but I hit a blocker: I need to be able to check if an entity is group content, but this is currently not possible in OO code. It is currently handled by `Og::isGroupContent()` which delegates to a static method `OgGroupAudienceHelper::getAllGroupAudienceFields()`. I cannot mock this :-1:

It seems like `OgGroupAudienceHelper` is one of the final bastions of procedural code left in OG. It only contains 2 methods so it will probably be a very quick conversion.

I will make a separate ticket for this.

pfrenssen:
I have converted `OgGroupAudienceHelper` into a service in #199. I will merge that branch in here so this is unblocked.

pfrenssen:
The approach may be simplified but these tests are far from simple :sweat_smile:

pfrenssen:
I must say I find it much more rewarding to write unit tests as opposed to kernel tests, just the fact that they run so fast makes it much more interesting to work on. I always lose my focus if my test takes longer than a few seconds to run.

amitaibu:
I also like unit tests, as they allow you to perform own code review really well.

src/ContextProvider/OgContext.php
```diff
@@ -0,0 +1,209 @@
...
+    });
+
+    // @todo This should be handled in the collection.
+    // Since we have filtered the results by the current user, we need to add
+    // the user as a cache context.
+    $this->addCacheContextIds(['user']);
+
+    // @todo Move this in the collection? Seems to make sense, since the
+    //   collection has knowledge about the groups and votes.
+    // Find the best matching group by iterating over the candidates and return
```
amitaibu: so, is `votes` like +1 that each plugin may give?

pfrenssen: Yes, with a small difference, their vote will be depending on their position in the list, so the first plugin's vote will count for more than the second plugin's vote. This will allow us to easily choose the correct "winner" if we have two groups with the same number of votes.

pfrenssen: Note that the OgContext code still needs to be cleaned up and tested, it still contains parts from the original implementation. But the basic idea will remain the same.

pfrenssen:
This still needs a test for `OgResolvedGroupCollection` and for `OgContext` itself. `OgResolvedGroupCollectionTest` will be a very straightforward unit test, but I'm not yet sure what would be the best approach to test `OgContext`.

I feel hesitant to write a functional test for this, since we would have to provide a test module that contains "something" (e.g. a block or a page controller) that consumes the available context and outputs data about it that we can check. For me providing a test module specifically for this seems overkill. Also I have more fun things to do in my spare time than waiting for functional tests to run :grin:

A unit test will be quite complicated since much of the business logic in in protected methods, so this would mean that we need to do a ton of mocking.

Maybe a kernel test would be a solution, but this functionality relies on requests being made, which is normally in the domain of functional tests. I don't know how to 'fake' a request in a kernel test.

I will investigate this further in the weekend. For the moment I'm pretty happy with how this is turning out. It also seems to work quite nice in practice. I just replaced our project's custom context provider with this and ran our entire Behat test suite against it and it came back green.

amitaibu:
This will require a re-roll now.

pfrenssen:
Thanks for letting me know. Going to pick this back up!

pfrenssen:
Haha thanks for the compliment! :) I'm having great fun writing these, it's like a bunch of little puzzles to solve, and I'm learning so much about the internal workings of things like typed data.

This test is not very clean, but this is due to `ContextProviders` not using dependency injection. They use typed data which they get from `\Drupal::typedDataManager()` so I have to put a mocked `TypedDataManager` on the container.

I'm currently sprinting at a local open source conference so there is a good chance I will be able to finally finish this today. Fingers crossed :)

pfrenssen:
Alright that went pretty smooth! I think this is ready!

@amitaibu care for a look?

amitaibu:
Yay! (Hectic days so might take a couple of days for me to have time to do _proper_ review.)

amitaibu:
I saw two `todo`s there, but there's really no reason to not get this in. Thank you!!

PR #189, opened 2016-09-23
Drupal console admin tasks integration

RoySegall:
This is a first try for integration with Drupal Console which we can improve most of the development issues(setting up a node type for group and content type with dummy nodes and users).

Things I need to figure:
1. How to translate or we just don't translate
2. How to unit test it

og 1


amitaibu:
Interesting approach. Seems it could be generalized to any field, not just OG ones, no?

RoySegall:
Amm... yes. Any field. But we can't attach any field with using OG field plugin manager. Other actions we can have - setting an entity as group type, assign permission to user, create roles, create permissions, generate OG plugins etc. etc.

amitaibu:
Ok, I see your point. Sounds cool 😄

RoySegall:
Hmmm.. I think I figured out how to test it. I copied code from DrupalConsole unit testing and hope travis will like all the stuff I put there.

RoySegall:
@amitaibu I moved the test to kernel testing.  And it's working. What I need to handle is Travis CI not recognising Drupal console name spacing.

.travis.yml
```diff
@@ -50,6 +50,14 @@ before_script:
   # Remember the Drupal installation path.
   - DRUPAL_DIR=$(pwd)/drupal
 
+  # Reference OG in the Drupal site.
+  - cp -R $MODULE_DIR $DRUPAL_DIR/modules/og
+
+  # Geting composer depdandecies.
+  - cd $DRUPAL_DIR
+  - composer require drush/drush:8.0.5
+  - vendor/drush/drush/drush dl composer_manager
```
pfrenssen: Why not add drush and composer_manager as dev dependencies in `composer.json`?

RoySegall: This could be an option but I wanted to this from a paths I already know. I'm still working on Improving the process.

.travis.yml
```diff
@@ -50,6 +50,14 @@ before_script:
   # Remember the Drupal installation path.
   - DRUPAL_DIR=$(pwd)/drupal
 
+  # Reference OG in the Drupal site.
+  - cp -R $MODULE_DIR $DRUPAL_DIR/modules/og
```
pfrenssen: This is already taken care of, a symlink is created that puts the module in the drupal installation.

scripts/travis-ci/run-test.sh
```diff
@@ -12,7 +12,10 @@ case "$1" in
         ;;
     *)
         ln -s $MODULE_DIR $DRUPAL_DIR/modules/og
+        composer install
         cd $DRUPAL_DIR
+        php modules/composer_manager/scripts/init.php
```
pfrenssen: Why do you need composer manager?

.travis.yml
RoySegall: Yeah, I saw it. Moving the work to there.

scripts/travis-ci/run-test.sh
RoySegall: I need it to update drupal autoload so when I test the integration I won't failed.

.travis.yml
RoySegall: On the other hand, since the module is deprecated and I only need the script I get clone it.

scripts/travis-ci/run-test.sh
pfrenssen: You don't need composer manager for that, you can do this:

```
$ composer dump-autoload
```

But that should also not be necessary, the autoload files that are initially created should normally work out of the box. If it needs to be rebuilt this indicates that something is wrong with the order in which you perform your build steps.

.travis.yml
pfrenssen: ok go ahead, you can still move it in composer.json when this is closer to completion :)

scripts/travis-ci/run-test.sh
RoySegall: Since we are not thousands KM away as we are usually - lets talk about it :)

RoySegall: No - https://travis-ci.org/Gizra/og/jobs/163120777#L645 - Not working.

scripts/travis-ci/run-test.sh
```diff
@@ -13,6 +13,9 @@ case "$1" in
     *)
         ln -s $MODULE_DIR $DRUPAL_DIR/modules/og
         cd $DRUPAL_DIR
+        git clone --branch 8.x-1.x https://git.drupal.org/project/composer_manager.git modules/composer_manager
```
amitaibu: Why do we need this? I this @pfrenssen said it was not needed?

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure() {
+    $this
+      ->setName('og:add_field')
+      ->setDescription('Attach OG field to entities')
+      ->addOption('field_id', '', InputArgument::OPTIONAL, 'Field ID')
+      ->addOption('field_name', '', InputArgument::OPTIONAL, 'Field name')
+      ->addOption('entity_type', '', InputArgument::OPTIONAL, 'The entity type. i.e node, user, taxonomy_term')
```
amitaibu: `The entity type ID (for example node, user, taxonomy_term)`

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,146 @@
...
+      ->addOption('bundle', '', InputArgument::OPTIONAL, 'The bundle name')
+      ->addOption('target_entity', '', InputArgument::OPTIONAL, 'The referenced entity type. i.e node, user, taxonomy_term');
```
amitaibu: (same as above)

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Check if the field is an audience field or not.
+   *
+   * @param string $field_id
+   *   The field ID.
+   *
+   * @return bool
+   *   True or false if the field is an audience or not.
+   */
+  protected function fieldIsAudienceField($field_id) {
```
amitaibu, line 174: I don't think you need specific check for audience field. you just need a check if it's an ER field

src/Command/OgAddFieldCommand.php
amitaibu: although, it's worth adding a comment it only applies for ER related fields.

tests/src/Kernel/Console/DrupalConsoleAddFieldTest.php
```diff
@@ -0,0 +1,83 @@
...
+    $this->installSchema('system', 'sequences');
+
+    NodeType::create([
+      'name' => $this->randomString(),
+      'type' => 'article',
+    ])->save();
+  }
+
+  /**
+   * Tests creating membership for an un-saved group.
```
amitaibu: comment and method name seem to be wrong.

scripts/travis-ci/run-test.sh
RoySegall: I asked @webflo and he says that for now this legit. We can also set up a composer.json file for travis but this will take more time.

src/Command/OgAddFieldCommand.php
RoySegall: > applies for ER related fields 

You want this as a comment on the code or part of the help text?

amitaibu:
Part of the help text

scripts/travis-ci/run-test.sh
pfrenssen: Composer manager is not needed in D8. Setting up the dependencies in composer.json takes less than 10 seconds.

pfrenssen:
Why did you commit the `composer.lock` file? Do we need to lock the dependencies? Why?

Leaving this uncommitted will alert us immediately when we have an incompatibility with a new release in one of our dependencies, and we will not need to check manually for updates for the entire lifetime of the project. Maintaining a `composer.lock` file is very expensive.

src/Command/OgAddFieldCommand.php
amitaibu: `The referenced entity type, when an entity-reference type field is added.`

pfrenssen:
I better put my money where my mouth is, I have made a commit that removes the dependency of Composer Manager, and simply downloads Drupal Console by adding it to `composer.json`. I don't have commit access to the gizra repo, but I have made it on amitaibu/og: https://github.com/amitaibu/og/commit/a8921861ea96a381b67852b95df48bf738fe623a

It also removes the `composer.lock` file so that @amitaibu does not have to check every day if there is an update to Coder :)

scripts/travis-ci/run-test.sh
RoySegall: I'll try construct something with composer manager but I have a feeling this will make things much more complex.

pfrenssen:
I made a small update so that the installation of Drupal Console is not done when doing a coding standards check: https://github.com/amitaibu/og/commit/38d641a5218604c182933319ba8b566c3cb70f7c

You can get the two commits by merging in this branch: https://github.com/amitaibu/og/tree/drupal-console

amitaibu:
@RoySegall please assign to me when ready for review. A blog post about it would be also great :)

RoySegall:
Sure. I'll merge @pfrenssen changes and see if this helped. I also talked with the maintainer and he said there is an update to drupal console that would probably will break the code.

Do you think we need to handle it for now or because we in a close version we can tackle it later?

amitaibu:
Better to tackle now

RoySegall:
Iv'e merged @pfrenssen work. Let see if travis accept it or not.

RoySegall:
@pfrenssen - thanks for your work, the composer manager is no longer needed. I'll keep working on the test on Tuesday.

RoySegall:
@amitaibu @pfrenssen When downloading the drupal console with `composer require drupal/console --working-dir=...` we get version 0.11.3 or something but the new version with the dependency injection is in `1.0.0-rc5`.

I tried couple of ways on my local machine and I managed to download it using the command `composer require drupal/console-core:1.0.0-rc5 drupal/console:1.0.0-rc5` but it did not end well in [travis](https://travis-ci.org/Gizra/og/jobs/166826247#L202)

I'll try to update Drupal Core on my local machine and update.

RoySegall:
OK. I managed to fix the download of the composer  by using `composer require symfony/dom-crawler:2.8.11 drupal/console-core:1.0.0-rc5 drupal/console:1.0.0-rc5 --working-dir=$DRUPAL_DIR`

So now the tests are failing because of something else. But I have a feeling we over exaggerating. People download drupal console using compoer will get `0.11.3` when we targetting the tests and the workflow for `1.0.0-rc5`

I'll fix up the tests and we can decide what we want to do next.

RoySegall:
Tests are green again. Moving on to fix the comments.

src/Command/OgAddFieldCommand.php
RoySegall: This one is a bit complex.

.travis.yml
```diff
@@ -50,6 +50,10 @@ before_script:
   # Remember the Drupal installation path.
   - DRUPAL_DIR=$(pwd)/drupal
 
+  # Download Drupal console so we can run the test for it. Skip this for the
+  # coding standards test.
+  - test ${TEST_SUITE} == "PHP_CodeSniffer" || composer require symfony/dom-crawler:2.8.11 drupal/console-core:1.0.0-rc5 drupal/console:1.0.0-rc5 --working-dir=$DRUPAL_DIR
```
pfrenssen: It's better if you don't hardcode the versions, otherwise we will have to maintain this manually and check every day if there is a new version available of Drupal Console.

If you need 1.0.0-rc5 as a minimum version, can you check if you can get it to work with this command?

```
$ composer require drupal/console:~1.0.0-rc5
```

Also we don't depend on `symfony/dom-crawler` and `drupal/console-core`, only on `drupal/console`. If you need to add these to work around a dependency problem inside Drupal Console then you need to document this really clearly. The documentation says:

```
# Download Drupal console so we can run the test for it.
```

It doesn't say anything about why we need `dom-crawler`. I think if we maybe update the documentation like this it would be clearer to future maintainers why we are doing weird stuff:

```
# Download Drupal console so we can run the test for it. Skip this for the
# coding standards test.
# Drupal console requires a newer version of symfony/dom-crawler than
# Drupal core. Work around this by updating core to this version.
```

Also don't hardcode `dom-crawler` to 2.8.11, that will just make it necessary to look at this sooner, what if in the future Drupal 8.2.x is released which requires 2.8.12?

I think we should probably report this to Drupal Console and make sure they are compatible with the same version of `symfony/dom-crawler` that Drupal core uses.

RoySegall: > I think we should probably report this to Drupal Console

Checked it up and this what he suggest. Drupal need a version of doc crawler and this cause for dependency issues.

> It's better if you don't hard code the versions, 

If we won't specify the newer version then we get 0.11.3 which is a version with, if I use nice words, bad API.

I'll open up an issue in Drupal Console with the versioning issues.

pfrenssen: Who is talking about 0.11.3? I suggested to not hardcode the version but use the standard way of doing it, specifying a minimum version, like this:

```
$ composer require drupal/console:~1.0.0-rc5
```

.travis.yml
```diff
@@ -52,7 +52,7 @@ before_script:
 
   # Download Drupal console so we can run the test for it. Skip this for the
   # coding standards test.
-  - test ${TEST_SUITE} == "PHP_CodeSniffer" || composer require symfony/dom-crawler:2.8.11 drupal/console-core:1.0.0-rc5 drupal/console:1.0.0-rc5 --working-dir=$DRUPAL_DIR
+  - test ${TEST_SUITE} == "PHP_CodeSniffer" || composer require symfony/dom-crawler:~2.8.11 drupal/console-core:~1.0.0-rc5 drupal/console:~1.0.0-rc5 --working-dir=$DRUPAL_DIR
```
pfrenssen: Great, it works!

Can you also add documentation to explain that we do not directly depend on `dom-crawler` and `console-core`, but that this is a workaround that should be removed in the future?

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @package Drupal\og
+ */
+class OgAddFieldCommand extends Command {
+
+  /**
+   * OG plugin manager.
+   *
+   * @var OgFieldsPluginManager
+   */
```
pfrenssen, line 27: Use fully qualified namespaces in the documentation.

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,178 @@
...
+  protected $ogFieldsPluginManager;
+
+  /**
+   * Bundle info service.
+   *
+   * @var EntityTypeBundleInfo
+   */
```
pfrenssen, line 34: Use fully qualified namespaces, and also the interface if it is available, so:

```
/**
 * @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface
 */
```

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,178 @@
...
+  protected $bundleInfo;
+
+  /**
+   * Entity type repository service.
+   *
+   * @var EntityTypeRepository
+   */
```
pfrenssen, line 41: `@var \Drupal\Core\Entity\EntityTypeRepositoryInterface`

src/Command/OgAddFieldCommand.php
```diff
@@ -0,0 +1,178 @@
...
+   * Constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeBundleInfo $bundle_info
+   *   The bundle info service.
+   * @param \Drupal\Core\Entity\EntityTypeRepository $entity_type_repository
+   *   The type repository service.
+   * @param \Drupal\og\OgFieldsPluginManager $og_field_plugin_manager
+   *   The OG plugin manager.
+   */
+  public function __construct(EntityTypeBundleInfo $bundle_info, EntityTypeRepository $entity_type_repository, OgFieldsPluginManager $og_field_plugin_manager) {
```
pfrenssen: Here in the type hints it is really important to use the interfaces and not the services.

pfrenssen: By the way, the reason why it is important to specify the interface and not the service directly is because the services are swappable. As long as they adhere to the same interface, the code should not care whether the default implementation from Drupal core is used, or whether they are using a compatible service that is customized to fulfill some business need.

If you do not specify the interface you are literally preventing developers from overriding services with their own versions, they will get a fatal error if they do. This kind of flexibility is very important for Drupal developers.

RoySegall:
@pfrenssen  10x for the comments.

I also opened an issue in Drupal Console - https://github.com/hechoendrupal/DrupalConsole/issues/2803

tests/src/Kernel/Console/DrupalConsoleAddFieldTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Drupal\Tests\og\Kernel\Console;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\node\Entity\NodeType;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Console\Helper\HelperSet;
+
+/**
+ * Tests create membership helper function.
```
pfrenssen: This does more than just testing a function, what do you think about:

> Tests creating group audience fields through the command line interface.

RoySegall: I'll take it.

RoySegall:
@pfrenssen Care to have a second look?

pfrenssen:
Looks great now, thanks! There is still @amitaibu 's comment about `fieldIsAudienceField()`, and I personally wouldn't have used an URL shortener to point to the followup issue, but that too minor to block this issue on.

Great work, this will be really handy!

RoySegall:
> Looks great now, thanks! There is still @amitaibu 's comment about fieldIsAudienceField()

@amitaibu suggested a follow up issue for this one.

src/Command/OgAddFieldCommand.php
amitaibu: Please open a follow up issue (and explain the complexity there)

RoySegall:
@amitaibu Don't merge yet, there some changes in the Drupal console composer.json file. I need to see if that fixed something.

RoySegall:
Still no new from the Drupal console part - https://github.com/hechoendrupal/DrupalConsole/issues/2803#issuecomment-254413059

RoySegall:
@amitaibu I opened a follow-up issue and removed the Drupal console from the JSON file and now It's read for review.

.travis.yml
```diff
@@ -50,6 +50,15 @@ before_script:
...
   - DRUPAL_DIR=$(pwd)/drupal
 
+  # Download Drupal console so we can run the test for it. Skip this for the
+  # coding standards test.
+
+  # When downloading DrupalConsole using Composer an old version is supplied.
+  # Sepficing the correct version and the versions of DrupalComposer dependancy
+  # will fix the problem for now. Need to be removed when https://goo.gl/i6rVhT
+  # will be merged.
+  - test ${TEST_SUITE} == "PHP_CodeSniffer" || composer require drupal/console-core:~1.0.0-rc5 drupal/console:~1.0.0-rc5 --working-dir=$DRUPAL_DIR
```
pfrenssen: Great news, the dependencies have been fixed in the 1.0.0-rc6 release of Drupal Console, this means that we normally don't need this workaround any more. Could you try replacing this with the following?

```
  - test ${TEST_SUITE} == "PHP_CodeSniffer" || composer require --dev drupal/console:~1.0@rc --working-dir=$DRUPAL_DIR
```

If this works then we can also remove the notice about the workaround.

RoySegall: Push the changes. Let's see what will be the outcomes.

RoySegall:
@amitaibu @pfrenssen I applied the small fix and the tests still green which mean this works OK. But I still think we need to keep the workaround so people would know they need the RC version until Drupal console finish that stage.

amitaibu:
Great stuff. Next is the Og Context and removal of the Og complex widget -- and we can cut a first alpha!

PR #191, opened 2016-09-27
Add Og::getParentGroups() and initial tests

pwolanin:


src/Og.php
```diff
@@ -128,6 +128,34 @@ public static function createField($plugin_id, $entity_type, $bundle, array $set
...
   /**
+   * Get the parent groups of an entity.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity objects.
+   *
+   * @return array
+   *   Array of group entities.
+   */
+  public static function getParentGroups(EntityInterface $entity) {
```
amitaibu, line 139: Maybe we could move this to an `OgSubgroups.php`, as it will not be as used as `Og.php`

tests/src/Kernel/Entity/GetSubGroupContentTest.php
```diff
@@ -0,0 +1,227 @@
...
+    for ($i = 0; $i < 5; $i++) {
+      $values = [
+        'title' => $this->randomString(),
+        'type' => $bundle,
+      ];
+      if ($i > 0) {
+        // Make this group a sub group of the one we created before.
+        $values[$field_name] = [['target_id' => $groups[$i]->id()]];
+      }
+      $entity = $this->entityTypeManager->getStorage('node')->create($values);
```
amitaibu, line 206: I think we can use `Node::create()` here

amitaibu: also `$entity` => `$node`

tests/src/Kernel/Entity/GetSubGroupContentTest.php
```diff
@@ -0,0 +1,227 @@
...
+    $values = [
+      'title' => $this->randomString(),
+      'type' => $bundle,
+      $field_name => [['target_id' => $groups[4]->id()]],
+    ];
+    $entity = $this->entityTypeManager->getStorage('node')->create($values);
+    $entity->save();
+    $groups[] = $entity;
+    $expected_parent_count[$entity->id()] = 5;
+    foreach ($groups as $idx => $node) {
```
amitaibu: `foreach ($groups as $group) {`

amitaibu (commented):
Looks good. Maybe we could also add an access method that will return an `AccessResult` based in a given entity.

amitaibu:
@pwolanin are you still around? :)

pwolanin:
@amitaibu yes - but this was on the back burner while Justin revamped various custom code to use og instead of group module.  That's just done, so hope to give it more time soon.

amitaibu:
Sounds good :)

tests/src/Kernel/Entity/GetSubGroupContentTest.php
MPParsley:
```suggestion
    foreach ($groups as $node) {
```

Issue #192, opened 2016-09-29
\Drupal\og\Routing\RouteSubscriber should use interfaces as type hints in the constructor


brummbar:
While enabling the Webprofiler module, I got errors due to the type hints used in the RouteSubscriber constructor.
Interfaces should be used instead of classes as injected services:
- `EntityTypeManagerInterface` should be used instead of `EntityTypeManager`;
- `EventDispatcherInterface` should be used instead of `ContainerAwareEventDispatcher`.

zerolab:
This can be closed since #193 was merged in.

PR #193, opened 2016-09-29
Fix type hints in RouteSubscriber

brummbar:
Fix for https://github.com/Gizra/og/issues/192 .


brummbar:
This failure should not be related to the changes. Can the tests be re-run?

amitaibu:
We seem to have a false positive in one of the tests. Anyway, it passes now. Thanks.

PR #197, opened 2016-10-03
Group roles load

idimopoulos:
Given the work in progress of the UI and the use cases where we need to check roles per group, I was thinking that maybe we can have a method as part of the Role manager class that can load the roles of a group.

Currently, there is only a method to load a role given a group and a role name (in OgRole) but we do not have the sense of having all the roles of a group retrieved.

This was based on https://github.com/amitaibu/og/pull/196/files#diff-f5fb6aad0c225e8559cacebfcbf5d915R142 

Can this be valid? Is this functionality useful to be separate (since it is just a few lines of code)?

### DIFF
--- a/src/OgRoleManagerInterface.php
+++ b/src/OgRoleManagerInterface.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace Drupal\og;
+use Drupal\Core\Entity\EntityInterface;
 
 /**
  * Defines an interface for OG role manager.
@@ -44,6 +45,17 @@ interface OgRoleManagerInterface {
    *   role_type.
    */
   public function getRequiredDefaultRoles();
+
+  /**
+   * Returns all the roles of a provided group.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $group
+   *   The group for which to return the roles.
+   *
+   * @return \Drupal\og\OgRoleInterface[]
+   *   An array of roles indexed by their ids.
+   */
+  public function loadRolesByGroup(EntityInterface $group);
 
   /**
    * Deletes the roles associated with a group type.
--- a/src/OgRoleManagerInterface.php
+++ b/src/OgRoleManagerInterface.php
@@ -56,7 +56,7 @@ interface OgRoleManagerInterface {
    * @return \Drupal\og\OgRoleInterface[]
    *   An array of roles indexed by their ids.
    */
-  public function getRolesByGroupTypeAndBundle($entity_type_id, $bundle);
+  public function loadRolesByBundle($entity_type_id, $bundle);
 
   /**
    * Deletes the roles associated with a group type.
--- a/src/OgRoleManagerInterface.php
+++ b/src/OgRoleManagerInterface.php
@@ -49,13 +49,15 @@ interface OgRoleManagerInterface {
   /**
    * Returns all the roles of a provided group.
    *
-   * @param \Drupal\Core\Entity\EntityInterface $group
-   *   The group for which to return the roles.
+   * @param string $entity_type_id
+   *    The entity type id of the group.
+   * @param string $bundle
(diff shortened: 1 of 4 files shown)


src/OgRoleManagerInterface.php
```diff
@@ -46,6 +46,19 @@ public function getDefaultRoles();
...
+   *
+   * @param string $entity_type_id
+   *    The entity type id of the group.
+   * @param string $bundle
+   *    The bundle of the group.
+   *
+   * @return \Drupal\og\OgRoleInterface[]
+   *   An array of roles indexed by their ids.
+   */
+  public function loadRolesByGroup($entity_type_id, $bundle);
```
pfrenssen: Wouldn't it be better to pass in a group object? This makes more sense according to the method name, and the current documentation "of a provided group". It would also be consistent with `OgRole::loadByGroupAndName()`.

On the other hand it might be really useful sometimes to load the `OgRole` entities without having a group entity. For example when we are building the UI for the permission matrix - then we do not have a group entity handy, but we do know the entity type and bundle.

So another idea would be to rename this method to `loadRolesByGroupTypeAndBundle()` or  even better `getRolesByGroupTypeAndBundle()` since we usually use the term `get()` instead of `load()` when we request data from a service in OG.

`OgRole::loadByGroupAndName()` is an exception, here we use `load...()` so it matches `Entity::load()`.

pfrenssen:
I like this proposal, being able to request a full list of `OgRole` entities from the `OgRoleManager` seems very useful to me. It is not hard to call `EntityStorageInterface::loadByProperties()` to get the list directly from the storage, but this requires intimate knowledge about how the `OgRole` enties work internally.

Calling a method on the `OgRoleManager` on the other hand is very easy to discover, and doesn't require any special knowledge. Great PR! :+1:

src/OgRoleManagerInterface.php
idimopoulos: Well, I can rename if it's better and more consistent. As for the group, if you check the commits, I implemented it passing the group but reaching the tests I had to create a dummy entity only to get this functionality which makes it unusable so I changed it to retrieve the machine names of the entity keys needed.

pfrenssen: Yeah that makes sense, let's not force our users to have to call `OgRole::create()` if they have no group entity nearby :D

amitaibu: Maybe we can call it `loadRolesByBundle($entity_type_id, $bundle)`. Other than that - nice addition 👍

pfrenssen:
Ok now it is changed to `loadRolesByBundle()`, I'm not sure how I feel about this name :) We're not only passing in a bundle, and as I said before we always use 'get' instead of 'load' in OG. I just grepped the code and we have 186 instances of `function getX()` and only one single instance of `function loadX()`. Using load will make this harder to discover I think.

@amitaibu what do you think?

pfrenssen:
Oh amitai already commented and suggested the `loadRolesByBundle()`, never mind :)

amitaibu:
Yeah, I think I got it wrong :/ getRolesByGroupTypeAndBundle() is probably the right one. Do you agree?

pfrenssen:
I found some precedence where we are using `get...ByBundle()`, implying that this is actually by entity type ID and bundle. There are a couple of these in `GroupTypeManager`, for example `GroupTypeManager::getGroupContentBundleIdsByGroupBundle($group_entity_type_id, $group_bundle_id)`.

So following this pattern `getRolesByBundle()` would be good too :)

amitaibu:
Cool. Let's go indeed with getRolesByBundle -- it sounds most right (and not too verbose) to me

idimopoulos:
Renaming done :)

amitaibu:
Thanks!

PR #198, opened 2016-10-04
Remove unused OgPermissionHandler service

pfrenssen:
An earlier implementation a service that handled OG permissions using YAML files is still present in the repository. This is no longer used, it is superseded by `OgPermissionManager`. Let's get rid of it.


amitaibu:
Needs a re-roll

amitaibu:
(I'll re-roll)

amitaibu:
btw, this made me think. Nothing that I would change now, but as an idea - The current implementation we have - couldn't we keep declaring the permissions via yaml, and have our `GroupPermission` and `GroupContentPermission`?

So we'd still have the convenience of config yaml files, and the safety of the wrapping objects?

amitaibu:
@pfrenssen ^^

amitaibu:
- Thanks :)

pfrenssen:
I think it could be possible to do this, have a YAML file that declares permissions, and load these in addition to the ones that are provided by the event. We'd need to think a bit for the structure of the YAML file but it should be no problem for declaring simple static permissions.

amitaibu:
Opened #200

PR #199, opened 2016-10-04
Convert OgGroupAudienceHelper to an injectable service

pfrenssen:
`OgGroupAudienceHelper` is one of the last bastions of procedural code in OG. Let's convert it into a service so we can inject it into whatever OO code that needs it. This will make it unit testable and will make it possible to swap it out for a customized implementation if this is needed for some reason.


pfrenssen:
This is blocking #180. The unit tests in that PR need to be able to mock a method that checks if an entity is group content. The conversion to a service will allow to do that.

amitaibu:
👍

pfrenssen:
Awesome, you're on a roll! Thanks!

PR #201, opened 2016-10-10
Add the ability to remove and adjust existing routes in OgAdminRoutesEventInterface

webflo:
I think its currently not possible to add or alter existing routes to the route list. Expanded the interface to match the documentation.

PR #202, opened 2016-10-11
Manage members route should use og-based access checks

webflo:


webflo:
Follow-up for https://github.com/Gizra/og/pull/201

amitaibu:
Merged, Thanks.

PR #181, opened 2016-09-15
Remove OG complex widget

amitaibu:
#160

PR also makes sure a group member can create content if they have `create` access.

PR #203, opened 2016-10-18
Show only groups in Og audience field settings

amitaibu:
This is cherry picked out of #181

PR also removed some obsolete schema settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ language: php
 sudo: false
 
 php:
-  - 5.6
+#  - 5.6
   - 7.0
 
 env:
-  - TEST_SUITE=8.1.x
+#  - TEST_SUITE=8.1.x
   - TEST_SUITE=8.2.x
-  - TEST_SUITE=8.3.x
-  - TEST_SUITE=PHP_CodeSniffer
+#  - TEST_SUITE=8.3.x
+#  - TEST_SUITE=PHP_CodeSniffer
 
 # Only run the coding standards check once.
 matrix:
--- a/config/schema/og.schema.yml
+++ b/config/schema/og.schema.yml
@@ -1,25 +1,3 @@
-field.storage_settings.og_membership_reference:
-  type: mapping
-  label: 'Organic Groups reference field storage settings'
-  mapping:
-    target_type:
-      type: string
-      label: 'Type of entity to reference'
-
-field.field_settings.og_membership_reference:
-  type: mapping
-  label: 'Organic Groups reference field settings'
-  mapping:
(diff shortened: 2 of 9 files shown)


amitaibu:
For example, the `Article` bundle was filtered out

groups_audience_settings_for_article___site-install

amitaibu:
Needs a Kernel test of course :)

amitaibu:
hmm, I don't see Kernel tests for forms in core - ones that tests forms that are created outside of the test itself. So I might opt for a Functional test instead.

amitaibu:
@liatsade fyi, this PR has a [Functional](https://github.com/Gizra/og/pull/203/files#diff-488dae2ab638f67a4970ccc28f809895R1) automatic test.

amitaibu:
Update: Test is passing locally, but fails on Travis :/

amitaibu:
Update: Trying to debug the tests on Travis, it [shows](https://travis-ci.org/Gizra/og/jobs/168748844#L574) that it gets a 404 on that page. 
It works locally, so  I wonder what it can be.

amitaibu:
@pfrenssen have you ever stumbled upon such a case?

(btw, I will of course rebase all those debug commits)

amitaibu:
First hint, [seems](https://travis-ci.org/Gizra/og/builds/168762632#L562) the user doesn't have access to that route (again, even-though it works on my local).

amitaibu:
@pfrenssen Ignore for now, I'm checking the access stuff based on the above hint.

amitaibu:
@pfrenssen nope, not access :/. Anyway, if you happen to have seen something like this..

amitaibu:
@nirgn975 if I had Travis [running a docker](http://www.gizra.com/content/docker-travis-ci/) it would be really helpful, as this debug turns out to be very hard. Can you look at this please in contribution hour?

amitaibu:
(as a different PR)

amitaibu:
@klausi do you happen to know [this](https://github.com/Gizra/og/pull/203#issuecomment-254640091) problem? Seems core does it with SimpleTest, but I prefer to use PHPunit.

klausi:
Looks like the response comes from the PHP webserver. Probably it cannot deal with the "." in the URL? Time to use a real web server like apache or nginx?

amitaibu:
@klausi sounds like a good guess! I wish I would have thought about it as-well too many hours ago 😉

amitaibu:
I've converted test to Kernel, and merged it back to #181 - as otherwise it would have required fixes to code that is soon going to be deleted anyway.

PR #208, opened 2016-10-25
WIP: Drupal docker.

nirgn975:
#203

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,7 @@
-FROM drupal:8.1-apache
+FROM nirgn975/docker-drupal:test
 
-# Install Composer.
-RUN curl -sS https://getcomposer.org/installer | php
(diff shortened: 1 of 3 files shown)


nirgn975:
@amitaibu Now I pull `drupal:8.1-apache` and `mysql` and `--link` them. I think about move it to a separate repo.

The things I'll get from move this to a separate repo:
- I'll have a build of this docker image (because on top of `drupal:8.1-apache` I also install `composer` and `drush`) - and in this repo I'll just pull the image, which will be faster then build it here every time `travis` run.
- After I'll do it to `og`, I can do PRs to `message`, etc.. 

Your thoughts?

amitaibu:
This needs a re-roll, I think you are not using the latest code.

 Also, lets start with 7.x instead of 8.x

PR #212, opened 2016-11-01
Filter OgRoleInterface::AUTHENTICATED on setRoles

webflo:
I use hook_og_membership_presave to add an additional role to the membership entity. hook_og_membership_presave is fired after OgMembership::preSave which revokes OgRoleInterface::AUTHENTICATED. But OgMembership::addRole calls ::getRole and this adds  the authenticated role again. ::setRoles should filter authenticated role automatically.


amitaibu:
Thanks.

PR #213, opened 2016-11-03
[BUGFIX] og membership access should use og_user_access.

ordavidil:


amitaibu:
Can we have a simpleTest to assert this please

Xilis:
Why should it use og_user_access? There is a comment right above that line, stating
 `// No-end user needs access to this entity, so restrict it to admins.`

Xilis:
Found some more related issues, maybe something like https://www.drupal.org/project/og/issues/2583303 goes in a better direction?

claudiu-cristea:

PR #215, opened 2016-11-15
Leverage OgContext for determining the membership state cache context

pfrenssen:
Now that `OgContext` is in, we can use it for determining the membership state cache context. This means we can remove a whole bunch of code from `OgMembershipStateCacheContext` and its accompanying test.

### DIFF
--- a/src/Cache/Context/OgMembershipStateCacheContext.php
+++ b/src/Cache/Context/OgMembershipStateCacheContext.php
@@ -4,9 +4,8 @@ namespace Drupal\og\Cache\Context;
 
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Cache\Context\CacheContextInterface;
-use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
 use Drupal\Core\Session\AccountInterface;
-use Drupal\og\GroupTypeManager;
 use Drupal\og\MembershipManagerInterface;
 use Drupal\og\OgMembershipInterface;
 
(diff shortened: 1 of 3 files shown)


pfrenssen:
This is ready for review (at least if the tests come back green :)

src/Cache/Context/OgMembershipStateCacheContext.php
```diff
@@ -82,25 +69,15 @@ public static function getLabel() {
...
-    if (!$entity_type_ids = array_keys($this->groupTypeManager->getAllGroupBundles())) {
-      // No group entities.
-      return self::NO_CONTEXT;
-    }
-
-    if (!$entity_type_ids = array_intersect(array_keys($route_contexts), $entity_type_ids)) {
-      // No parameters that match the group entities.
+    // Do not provide a cache context if there are no groups in the current
+    // context.
+    $contexts = $this->ogContext->getRuntimeContexts(['og']);
```
amitaibu, line 74: > `getRuntimeContexts(['og']);`

I realize I'm not clear on this `'og'` in the array. What other values can we pass here?

amitaibu (commented):
Looks good, just added a question.

src/Cache/Context/OgMembershipStateCacheContext.php
pfrenssen: A context provider can potentially return multiple contexts of different types, so you should pass in an array of context types that you are interested in. This allows the context provider to work more efficiently and not look up any unnecessary contexts.

Here's the relevant documentation from `ContextProviderInterface`:

```
  /**
   * Gets runtime context values for the given context IDs.
   *
   * @param string[] $unqualified_context_ids
   *   The requested context IDs. The context provider must only return contexts
   *   for those IDs.
   */
  public function getRuntimeContexts(array $unqualified_context_ids);
```

amitaibu: yeah, I was just not clear on what _other_ value you might have there apart of `['og']` (so maybe we can have a helper method to call `getRuntimeContexts(['og']);` since that's what we'll normally use, no?). 

Anyway, looks good

amitaibu:
👍

Issue #216, opened 2016-11-22
Clear an 'og-group-content' cache tag whenever group content changes


pfrenssen:
In my project I am displaying a listing of group content belonging to a specific group. I would like to invalidate the cache of this listing whenever new content is added to the group, or the existing content changes.

For the moment I have implemented this with two hook implementations in my project. I'm wondering if this would perhaps be useful to add to OG itself? I'm not sure how we can make this easily discoverable though, we currently don't offer any views or so that would benefit from having this cache tag.

```
/**
 * Implements hook_entity_update().
 */
function collection_entity_update(Drupal\Core\Entity\EntityInterface $entity) {
  // If group content is created or updated, invalidate the group content cache
  // tags for each of the groups this group content belongs to. This allows
  // group listings to be cached effectively. The cache tag format is
  // 'og-group-content:{group entity type}:{group entity id}'.
  if (Og::isGroupContent($entity->getEntityTypeId(), $entity->bundle())) {
    $tags = [];
    foreach (\Drupal::service('og.membership_manager')->getGroupIds($entity) as $entity_type => $ids) {
      foreach ($ids as $id) {
        $tags[] = "$entity_type:$id";
      }
    }
    Cache::invalidateTags(Cache::buildTags('og-group-content', $tags));
  }
}

/**
 * Implements hook_entity_delete().
 */
function collection_entity_delete(Drupal\Core\Entity\EntityInterface $entity) {
  collection_entity_update($entity);
}
```

amitaibu:
This is pretty sweet. We can add a `Recent group content` view to demo it (+ it's going to be a handy view)

pfrenssen:
OK thanks! Will make a PR for it.

pfrenssen:
> We can add a Recent group content view to demo it (+ it's going to be a handy view)

I just started on this but this is actually impossible, a View is bound to a single entity type, and group content can be of multiple entity types!

I can make a simple block though that presents the most recent group content in an unordered list.

amitaibu:
We can do recent group content _nodes_ - this is going to be the majority of cases.

pfrenssen:
OK!

pfrenssen:
[...]

PR #220, opened 2016-11-24
Provide a Views default argument plugin that gets the group from OgContext

pfrenssen:
I was trying to create a "Recent group content" view as part of my work on #216 but I had to abandon this idea because of my inadequate level of Views-fu. I still would like to contribute the default argument plugin I wrote as part of this work. It will get the active group from OgContext and supply it as a default argument.

How to test:

1. Set up a group that has group content of type `node`.
1. Create a new view named "Recent group content" that shows "Content" of type "All" sorted by "Newest first". Enable the "Create a block" checkbox and set it to display an "Unformatted list" of "titles (linked)".
1. Add a contextual filter to the view for the group audience field of your group content type.
1. Configure the contextual filter. Set the option "Provide default value" for "When the filter value is not available". From the dropdown select "Group ID from OG context".
1. Save the view.
1. In the block configuration, enable the "Recent group content" block.

You will now see a list of recent group content on your group pages.

### DIFF
--- a/src/Plugin/views/argument_default/Group.php
+++ b/src/Plugin/views/argument_default/Group.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Drupal\og\Plugin\views\argument_default;
+
+use Drupal\Core\Cache\Cache;
+use Drupal\Core\Cache\CacheableDependencyInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Default argument plugin to provide the group from the current context.
+ *
+ * @ViewsArgumentDefault(
+ *   id = "og_group_context",
+ *   title = @Translation("Group ID from OG context")
+ * )
+ */
+class Group extends ArgumentDefaultPluginBase implements CacheableDependencyInterface {
+
+  /**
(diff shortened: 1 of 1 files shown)


pfrenssen:
Made sure this is already providing the group content cache tags. These will be invalidated as soon as #216 is ready.

src/Plugin/views/argument_default/Group.php
```diff
@@ -0,0 +1,130 @@
...
+  public function getCacheMaxAge() {
+    return Cache::PERMANENT;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    // @todo This should return the active group cache context as soon as we
+    //   have it.
```
RoySegall, line 99: The indentation of the `have it` hurts my eyes :)

RoySegall:
Is there a way that you could a view with this PR? In OG7 we have couple of default views and I think a view with that contextual would be a good idea.

src/Plugin/views/argument_default/Group.php
pfrenssen: This is how it should be according to coding standards, two spaces inwards from the `@todo`.

pfrenssen:
I tried to create a view but failed. Views expects you to configure the field names, but this is user configurable, so we don't know in advance how the group audience field will be called. There might also be multiple group audience fields on a group content type.

I don't know enough about views to know if it is possible to have dynamically adapting field names.

I also don't care much about this PR, so if you think it's not much use without an accompanying view then feel free to close it. I made this as part of a proof of concept for #216, the thing I care about right now is clearing cache tags :)

I just thought it would be helpful for someone who wants to make views with OG.

amitaibu:
Looks good.

PR #221, opened 2016-11-24
Provide a block that lists recent group content

pfrenssen:
This PR offers a block that can be placed on the page and will list recent group content if there is a group in the current context (as discovered by `OgContext`).

I tried before to create this with Views but I failed miserably, so I provided a simple block instead. The site builder can configure the entity type and bundles of the group content to show.

The block is actually a proof of concept that shows that it is possible to now label render arrays with the `og-group-content` cache tag, and that it will correctly be invalidated if any content is added or remove from a group, or if existing group content changes. So this is actually a fix for #216 ;)


amitaibu:
Ready for review?

pfrenssen:
It was and then I fixed a little bug and now it isn't any more :)

Need to fix that test.

pfrenssen:
Fixed the test, nice to see that my little helper method makes the unit test a lot simpler and more understandable :) That really validates the usefulness of the method :)

Tagging for review, but the tests are still running. If they fail then this still needs work :)

pfrenssen:
Retracting the 'Ready for review' so I can get rid of one of the TODOs. This block needs a cache context for the active group (the one returned by `OgContext`). Currently it varies by URLs but this is not correct.

amitaibu:
np, I haven't reviewed properly yet -  got DrupalCamp TLV tomorrow :)

pfrenssen:
:+1: enjoy!!

pfrenssen:
This is ready for review now. I have not written a test for the block itself, but I have for the cache context which is the primary objective for me. We need the caching improvements for a project at work.

I can write a test for the block too but I'll have to do this in my personal time. That might take some time :)

src/Cache/Context/OgGroupContextCacheContext.php
```diff
@@ -0,0 +1,72 @@
...
+    return t('OG active group');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContext() {
+    // Do not provide a cache context if there is no group in the current
+    // context.
+    $group = $this->ogContext->getGroup();
```
amitaibu, line 55: I really like this new `getGroup()`

src/GroupTypeManager.php
```diff
@@ -233,6 +233,35 @@ public function getAllGroupBundles($entity_type = NULL) {
...
 
   /**
+   * Returns a list of all group content bundles keyed by entity type.
+   *
+   * This will return a simple list of group content bundles. If you need
+   * information about the relations between groups and group content bundles
+   * then use getGroupRelationMap() instead.
+   *
+   * @param string $entity_type_id
+   *   Optional entity type ID to filter the bundles by. If omitted bundles of
```
amitaibu: maybe split to two different methods - `getAllGroupContentBundles` and `getGroupContentBundlesBy EntityType`?

My reason is, that if you think you have an entity_type_id, but in fact you have a NULL, nothing warns you, and you get a "weird " answer.

src/Plugin/Block/RecentGroupContentBlock.php
```diff
@@ -0,0 +1,265 @@
...
+
+  /**
+   * The bundle information service.
+   *
+   * @var EntityTypeBundleInfoInterface
+   */
+  protected $entityTypeBundleInfo;
+
+  /**
+   * Constructs a CollectionContentBlock object.
```
amitaibu: `CollectionContentBlock` => `RecentGroupContentBlock`

src/Plugin/Block/RecentGroupContentBlock.php
```diff
@@ -0,0 +1,265 @@
...
+   *   The plugin ID for the plugin instance.
+   * @param string $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\og\OgContextInterface $og_context
+   *   The OG context provider.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\og\GroupTypeManager $group_type_manager
+   *   The OG group type manager.
+   * @param EntityTypeBundleInfoInterface $entity_type_bundle_info
```
amitaibu: `EntityTypeBundleInfoInterface` should be the full name, like the other params

src/Plugin/Block/RecentGroupContentBlock.php
```diff
@@ -0,0 +1,265 @@
...
+        //   for the created date.
+        ->sort('created', 'DESC')
+        // @todo Make the number of results configurable.
+        ->range(0, 5)
+        ->execute();
+
+      $ids = array_merge($ids, $results);
+    }
+
+    return $this->entityTypeManager->getStorage($entity_type)->loadMultiple($ids);
```
amitaibu, line 271: we should `$entity->access()` here as-well, to prevent showing content user doesn't have access to.

src/Plugin/Block/RecentGroupContentBlock.php
```diff
@@ -0,0 +1,265 @@
...
+    $ids = [];
+    foreach ($fields as $field) {
+      // Query all group content that references the group through this field.
+      $results = $this->entityTypeManager
+        ->getStorage($entity_type)
+        ->getQuery()
+        ->condition($field->getName() . '.target_id', $group->id())
+        ->condition($definition->getKey('bundle'), $bundles, 'IN')
+        ->accessCheck()
+        // @todo Add support for entity types that use a different column name
```
amitaibu, line 262: is there an easy way to do it? In fact, you can't even be sure you have a `created`, no?

Anyway, it's probably an edge case.

src/Plugin/Block/RecentGroupContentBlock.php
```diff
@@ -0,0 +1,265 @@
...
+        //   for the created date.
+        ->sort('created', 'DESC')
+        // @todo Make the number of results configurable.
+        ->range(0, 5)
```
amitaibu: maybe worth tackling this as part of this PR?

tests/src/Unit/Cache/Context/OgCacheContextTestBase.php
```diff
@@ -0,0 +1,121 @@
...
+    parent::setUp();
+
+    $this->ogContext = $this->prophesize(OgContextInterface::class);
+    $this->group = $this->prophesize(EntityInterface::class);
+  }
+
+  /**
+   * Tests the result of the cache context service with active context objects.
+   *
+   * This is the 'normal' test case. The service will be able to retrieve data
```
amitaibu: what do you mean by `'normal'` - You mean the fact it's `abstract`?

amitaibu (commented):
Thanks, I've added a few comments.

src/Cache/Context/OgGroupContextCacheContext.php
pfrenssen: Yeah I was tired of doing the little dance every time to open the context object and see if there is something inside it :)

src/GroupTypeManager.php
pfrenssen: Good idea! This also solves the problem that the method could return data in two different formats, depending on whether or not `$entity_type_id` is passed.

src/Plugin/Block/RecentGroupContentBlock.php
pfrenssen: No I think this is actually really hard to do. In core we have only a few entity keys that we can depend on, and 'created' is not one of them. We don't even have an interface for the created date that we can use to detect if an entity type even supports a creation timestamp. The closest we have is `EntityChangedInterface`.

To fix this we probably need to introduce an 'official' entity key for the creation date in core. This `@todo` is mainly here to inform people that we know this problem exists :)

I think the most practical way to circumvent this for a custom entity type is to simply subclass the block and override the method.

src/Plugin/Block/RecentGroupContentBlock.php
pfrenssen: Sure, let me add this, that's a quick fix that would make it more user friendly.

src/Plugin/Block/RecentGroupContentBlock.php
pfrenssen: Doesn't the `->accessCheck()` in the query account for this? To be honest I haven't tried it with inaccessible content.

tests/src/Unit/Cache/Context/OgCacheContextTestBase.php
pfrenssen: I mean that it is testing the regular process, how it is intended to be used. I'll update the wording.

pfrenssen:
Addressed the remarks, this is ready for a fresh inspection!

amitaibu:
Thanks!

PR #222, opened 2016-12-06
Add field_ui integration to OG membership types

jludwig:
#223 
I have been experimenting with membership types and added some code to make it easier to add/edit/remove membership types and membership type fields by adding a listbuilder and field_ui integration. I think it will be useful to others without adding too much bloat.

### Todo

- Tests
- Locking the default membership type from being deleted
- Possible cleanup, this is a pretty rough draft

Note that before doing any of this, my priority would be getting a UI going for dealing with the memberships themselves.

### DIFF
--- a/src/Form/OgMembershipTypeForm.php
+++ b/src/Form/OgMembershipTypeForm.php
@@ -6,7 +6,6 @@ use Drupal\Core\Entity\BundleEntityFormBase;
 use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\language\Entity\ContentLanguageSettings;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
--- a/src/Form/OgMembershipTypeForm.php
+++ b/src/Form/OgMembershipTypeForm.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace Drupal\og\Form;
+
+use Drupal\Core\Entity\BundleEntityFormBase;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\language\Entity\ContentLanguageSettings;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Form handler for og membership type forms.
+ */
+class OgMembershipTypeForm extends BundleEntityFormBase {
+
+  /**
+   * The entity manager.
+   *
+   * @var \Drupal\Core\Entity\EntityManagerInterface
+   */
+  protected $entityManager;
+
+  /**
+   * Constructs the OgMembershipTypeForm object.
+   *
+   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
+   *   The entity manager.
+   */
+  public function __construct(EntityManagerInterface $entity_manager) {
+    $this->entityManager = $entity_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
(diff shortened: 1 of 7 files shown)


amitaibu:
Oh nice. Could you also post screenshot(s), so it's easier to see how it looks.

jludwig:
2016-12-06-071215_1920x1200_scrot
2016-12-06-071234_1920x1200_scrot
2016-12-06-071241_1920x1200_scrot
2016-12-06-092307_1920x1200_scrot
2016-12-06-092313_1920x1200_scrot

amitaibu:
Thanks. Is this ready for review?

jludwig:
@amitaibu 

I think it's ready for review.

A few quick questions:

1. The default membership type is deletable as is the field that comes with it (Request Membership). I imagine these be "locked" (as in not deletable), right? Is this a blocker for getting the pull request accespted?
2. Should other membership types come with the "request membership" field, or should that remain as is? Currently a site builder can add that field using the standard add existing field way.
2. What do you think of the URL and name? The page is called "Membership types" and the base path is /admin/structures/membership-types

src/Form/OgMembershipTypeForm.php
```diff
@@ -0,0 +1,128 @@
...
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    $form = parent::form($form, $form_state);
+
+    $type = $this->entity;
+    if ($this->operation == 'add') {
+      $form['#title'] = $this->t('Add membership type');
+    }
+    else {
+      $form['#title'] = $this->t('Edit %label membership type', array('%label' => $type->label()));
```
amitaibu, line 53: `array()` => `[]` (here and below)

src/Form/OgMembershipTypeForm.php
```diff
@@ -0,0 +1,128 @@
...
+namespace Drupal\og\Form;
+
+use Drupal\Core\Entity\BundleEntityFormBase;
+use Drupal\Core\Entity\EntityManagerInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Form handler for og membership type forms.
```
amitaibu, line 12: og => OG

src/Form/OgMembershipTypeForm.php
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    parent::validateForm($form, $form_state);
+
+    $id = trim($form_state->getValue('type'));
+    // '0' is invalid, since elsewhere we check it using empty().
+    if ($id == '0') {
```
amitaibu, line 96: I wonder why it's not in a core class, since it's used by node as-well.

src/OgMembershipTypeListBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+    if (isset($operations['edit'])) {
+      $operations['edit']['weight'] = 30;
+    }
+    return $operations;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
```
amitaibu, line 50: can probably be removed

amitaibu (commented):
Haven't tested yet, but just a quick code review

amitaibu:
> imagine these be "locked" (as in not deletable),

Yeah, I think we should throw an exception if the bundle/ field is tried to be deleted. And probably also remove the delete link from the default bundle.

> Should other membership types come with the "request membership" field

Yeah, we should probably auto-create this field when a new bundle is added

> What do you think of the URL and name?...

👍

amitaibu:
Is this being replaced by #231 ?

jludwig:
No, they would work together. I'm going to try to get these ready this week.

jludwig:
This code was merged into #231 because they play so well together.

PR #226, opened 2016-12-14
OgMembership static cache are not cleared on membership save

idimopoulos:
I ran across a bug when I was trying to create a test using multiple users and groups which ends up being an issue in OgMembership static cache.

I am going to provide a failing test first and then a small fix but the main idea is (Steps to reproduce):

- An OgMembership is being created and saved.
- The OgMembership above is loaded. The static cache is being populated by the membership and user tag.
- A new OgMembership is being created for the same user in another group. The static cache is not being cleared.
- The new OgMembership is attempted to be loaded. No membership is found because the static cache only has the first value.

This happens because the static cache identifier tag is in the form of <code>\<OgMembership class>:\<user id>:\<state type></code>. When the second one is being loaded, the user does not change, so the static cache returns the first result only which was already cached.

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -369,6 +369,9 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
     Og::reset();
     \Drupal::service('og.access')->reset();
 
+    // Invalidate the group membership manager.
+    \Drupal::service('og.membership_manager')->reset();
+
     return $result;
   }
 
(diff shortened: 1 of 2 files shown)


idimopoulos:
By the way, MembershipManager class uses the same way to build the cache for the groupIds.
I haven't looked at it but there is a possibility that the issue can be the same there as well.
@see \Drupal\og\MembershipManager::getGroupIds

pfrenssen:
:+1:

Looks good to me! Failing test also proves that it works. Thanks!

amitaibu:
Thanks!

PR #230, opened 2017-01-05
OgDeleteOrphans Simple Plugin does not delete content

webflo:
The content deletion should be deferred because the OgDeleteOrphans plugins run in ``og_entity_predelete``. The group entity still exists at this point. ``OgDeleteOrphansBase::deleteOrphan`` does delete the group content because the ``$group_count`` is not zero (contains the original group).

I can think of two solutions:

1. defer the simple plugin until shutdown
2. adjust OgDeleteOrphansBase::deleteOrphan and exclude the the current group from the count

### DIFF
--- a/src/OgDeleteOrphansBase.php
+++ b/src/OgDeleteOrphansBase.php
@@ -128,6 +128,9 @@ abstract class OgDeleteOrphansBase extends PluginBase implements OgDeleteOrphans
    */
   protected function deleteOrphan($entity_type, $entity_id) {
     $entity = $this->entityTypeManager->getStorage($entity_type)->load($entity_id);
+    if (!$entity) {
+      return;
+    }
     // Only delete content that is fully orphaned, i.e. it is no longer
     // associated with any groups.
     $group_count = $this->membershipManager->getGroupCount($entity);
--- a/src/OgDeleteOrphansBase.php
+++ b/src/OgDeleteOrphansBase.php
@@ -128,9 +128,13 @@ abstract class OgDeleteOrphansBase extends PluginBase implements OgDeleteOrphans
    */
   protected function deleteOrphan($entity_type, $entity_id) {
     $entity = $this->entityTypeManager->getStorage($entity_type)->load($entity_id);
+
+    // The entity might already be removed by other modules that implement
+    // hook_entity_delete().
     if (!$entity) {
       return;
     }
+
     // Only delete content that is fully orphaned, i.e. it is no longer
     // associated with any groups.
     $group_count = $this->membershipManager->getGroupCount($entity);
--- a/tests/src/Kernel/OgDeleteOrphansTest.php
+++ b/tests/src/Kernel/OgDeleteOrphansTest.php
@@ -44,6 +44,13 @@ class OgDeleteOrphansTest extends KernelTestBase {
   protected $group;
 
   /**
+   * A test group content.
+   *
+   * @var \Drupal\Core\Entity\EntityInterface
+   */
+  protected $group_content;
+
+  /**
    * {@inheritdoc}
    */
   public function setUp() {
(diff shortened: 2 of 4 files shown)


webflo:
ping @pfrenssen and @sandervd for reviews. thanks!

tests/src/Kernel/OgDeleteOrphansTest.php
```diff
@@ -142,14 +151,19 @@ public function testDeleteOrphans($plugin_id, $run_cron, $asynchronous, $queue_i
...
+
     // Simulate the initiation of the queue process by an asynchronous operation
     // (such as pressing the button that starts a batch operation).
     if ($asynchronous) {
       $this->process($queue_id, $plugin_id);
     }
 
     // Verify the group content is deleted.
-    $this->assertFalse($this->group_content, 'The orphaned node is deleted.');
+    $this->assertNull(Node::load($this->group_content->id()), 'The orphaned node is deleted.');
```
webflo: $this->group_content was not defined, therefor always FALSE. Node::load return NULL for non-existing entities

src/OgDeleteOrphansBase.php
```diff
@@ -128,6 +128,9 @@ protected function query(EntityInterface $entity) {
    */
   protected function deleteOrphan($entity_type, $entity_id) {
     $entity = $this->entityTypeManager->getStorage($entity_type)->load($entity_id);
+    if (!$entity) {
```
webflo, line 134: Some implementations delete og related in their own delete functions (e.g. ogmenu). The queue implementation should not fail for these entities.

webflo (commented):
Self review with some further information.

src/OgDeleteOrphansBase.php
pfrenssen: OK makes sense. Let's add a comment to explain this switch.

tests/src/Kernel/OgDeleteOrphansTest.php
```diff
@@ -142,14 +151,19 @@ public function testDeleteOrphans($plugin_id, $run_cron, $asynchronous, $queue_i
...
+
     // Simulate the initiation of the queue process by an asynchronous operation
     // (such as pressing the button that starts a batch operation).
     if ($asynchronous) {
       $this->process($queue_id, $plugin_id);
     }
 
     // Verify the group content is deleted.
-    $this->assertFalse($this->group_content, 'The orphaned node is deleted.');
+    $this->assertNull(Node::load($this->groupContent->id()), 'The orphaned node is deleted.');
```
pfrenssen, line 166: OK this makes a lot of sense, to actually check that the entity is no longer available in the entity storage.

pfrenssen:
Looks good to me, just needs a line to explain why an entity could already have gone missing before it is being processed.

amitaibu:
@pfrenssen you now have write access to both GitHub and d.o. (sorry 😝 ) and congrats! 🎉

pfrenssen:
Hmm this was green here but causes a failure on the main branch after merging, only on Drupal 8.2.x: https://travis-ci.org/Gizra/og/builds/191619858

Out of scope for this issue though, it's the DrupalConsole integration that causes the failure.

webflo:
Thank you!

pfrenssen:
I have an interesting side effect for this issue: when running one of our Behat test suites I get over 100 duplicate callbacks in `drupal_register_shutdown_function()`. This happens because during this long running test a large number of groups are created and deleted, and each time a group with orphans is deleted a new duplicate callback is added.

This is not causing any problems so far, since the duplicate callbacks will not any processing, but we can maybe check if we the callback already exists before adding a new one.

webflo:
I think each instance of the ``Simple`` contains a different entity object. I would have to refactor it before we can register a single callback.

pfrenssen:
Good point, that seems indeed to be the case.

I just stepped through it to check what happens. It seems that the queue only needs to be processed once. Regardless of which `Simple` instance registers the entity to be deleted, they all get added to the same queue with ID `og_orphaned_group_content`.

Then the first time `Simple::process()` gets called, it will iterate over the entire queue and clean up all of it. All subsequent calls just iterate over an empty queue.

This is very minor though, just something that caught my attention.

PR #231, opened 2017-01-11
Adds basic UI for adding, editing and removing group memberships

jludwig:
#209 #223 

This adds Field UI integration w/Membership Types and Memberships, allowing admins to add/edit/remove both membership types and memberships in groups.

Note: This work was sponsored by BioRAFT. https://www.drupal.org/bioraft

## Todo

- Tests... Will get at at least a test base setup before getting this merged in. I know, I should have worked on the tests during development but that's not a habbit for me yet.
- Form validation for the membership UID currently only works with the OG group context autocomplete, not the standard Drupal autocomplete. This can be fixed by applying the constraint (already create) on the Entity level rather than just the widget level
- No way to tell someone's membership right now. Adding the bundle column to the view would be easy, but it would just clutter a membership view on a site with only one type of membership
- The last member can be removed from a group. Not sure how to approach this. Should it enforce that the group owner is always a member?
- A membership can be a group/group content (lol)
- The add member link from the membership page only works on nodes, because it uses og.links.action.yml rather than a dynamic way to achieve that
- Some performance tweaks, especially in the OG autocomplete where it loads the memberships associated with the current group context

## What it looks like in action

[membership_ui.pdf](https://github.com/Gizra/og/files/772904/membership_ui.pdf)

PR #244, opened 2017-03-10
Bulk operation for adding/removing roles from members

sandervd:
This PR adds a bulk operation to the members overview page that allows adding/removing roles from the members.

members-overview


zerolab:
Would be great to provide a summary of the changes and some screenshots. Currently "Isaicp 2709" gives little context before one dives into code.

Having said that, this looks like a nice addition!

src/OgMembershipViewsData.php
```diff
@@ -14,6 +14,13 @@ class OgMembershipViewsData extends EntityViewsData {
    */
   public function getViewsData() {
     $data = parent::getViewsData();
+    $data['og_membership']['og_membership_bulk_form'] = array(
```
zerolab: use array shorthand - `[]`

src/Plugin/Action/ChangeOgMembershipRoleBase.php
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $roles = user_role_names(TRUE);
+    unset($roles[RoleInterface::AUTHENTICATED_ID]);
+    $form['rid'] = array(
```
zerolab: ditto, use array shorthand

src/Plugin/Action/ChangeOgMembershipRoleBase.php
```diff
@@ -0,0 +1,103 @@
...
+      $this->addDependency('config', $prefix . $this->configuration['rid']);
+    }
+    return $this->dependencies;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function access($object, AccountInterface $account = NULL, $return_as_object = FALSE) {
+    return TRUE;
```
zerolab: Looks like this is not finished.

src/Plugin/views/field/OgMembershipBulkForm.php
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * {@inheritdoc}
+   *
+   * Provide a more useful title to improve the accessibility.
+   */
+  public function viewsForm(&$form, FormStateInterface $form_state) {
+    // Remove all actions related to (non-)member role.
+    foreach ($this->actions as $key => $action) {
+      if (strstr($key, '-member') !== FALSE) {
```
zerolab: This is prone to fail if one create a "Special member" role, for example. I think it is better to do `list( , , $rid) = explode('-', $key, 3)` and then check that `$rid` is not `OgRoleInterface::ANONYMOUS` or `OgRoleInterface::AUTHENTICATED`

zerolab (commented):
This is looking great. I started a few bulk actions for approving/blocking/removing memberships and happy to see that `OgMembershipBulkForm` is taking roughly the same shape.

A few minor comments from me

amitaibu:
PR & reviews from the community. It's a module's maintainer dream come true! 😄

src/Plugin/views/field/OgMembershipBulkForm.php
zerolab: Thinking more about this, the check for the default member role should happen in the actions themselves. Non-member is needed.

I am working on actions for the actual membership and they apply to all members (e.g. approve membership, block or even remove membership).

sandervd: I was already working on it 👍

sandervd: Non-member and member are more 'implied' roles, right?
(\Drupal\og\Entity\OgMembership::getRoles)
Removing the member should remove the actual membership entity, removing the role won't do the trick I guess...
Having those actions (remove membership) is valuable, but I should probably get rid of the  two actions related to the member and non member entirely, and replace it with an action that removes the OgMembership entity itself. But as you're working on those :)
What do you think about removing the role related actions here (not creating them in the insert hook)?

zerolab: non-member, member and administrator are default roles. The first one is implied as it applies any user that is not a member of the group. Group members can be group administrators. And a group administrator is always a group member.

The role related actions are useful. You'd want to want to make someone and administrator or give them custom group roles. They should not apply to member/non-member and that is.

Will provide a diff of the membership actions shortly.

zerolab: https://gist.github.com/zerolab/32a445c1a929084a613a5c545680ad6c

zerolab:
@sandervd I think you need an update hook to provide the administrator role action for existing builds.

og.module
```diff
@@ -290,3 +296,50 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
...
 }
+
+/**
+ * Implements hook_ENTITY_TYPE_insert() for user_role entities.
+ */
+function og_og_role_insert(OgRoleInterface $role) {
+
+  $add_id = 'og_membership_add_role_action.' . $role->id();
+  if (!Action::load($add_id)) {
+    $action = Action::create(array(
```
zerolab: I am sorry to nitpick, but there still are 8 occurrences of `array()` in this PR

src/Plugin/Action/RemoveOgMembershipRole.php
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute(OgMembership $member = NULL) {
+    if (!$member) {
+      return;
+    }
+    $rid = $this->configuration['rid'];
+    $roles = $member->getRoles();
```
idimopoulos: Why not do something like:
`

    if (in_array($rid, $member->getRolesIds())) {
      $role = OgRole::load($rid);
      $member->revokeRole($role);
      $member->save();
    }
`

src/Plugin/Action/AddOgMembershipRole.php
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute(OgMembership $member = NULL) {
+    if (!$member) {
+      return;
+    }
+    $rid = $this->configuration['rid'];
+    $roles = $member->getRoles();
```
idimopoulos: I think this can be simplified by changing from this line on to
`

    if (!in_array($rid, $member->getRolesIds())) {
      $member->addRole(OgRole::load($rid));
      $member->save();
    }
`
The og is already taking care of filtering out duplicated roles when adding one.

og.module
```diff
@@ -290,3 +296,50 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
...
+function og_og_role_delete(OgRoleInterface $role) {
+
+  $actions = Action::loadMultiple([
+    'og_membership_add_role_action.' . $role->id(),
+    'og_membership_remove_role_action.' . $role->id(),
+  ]);
+  foreach ($actions as $action) {
+    $action->delete();
+  }
+}
```
idimopoulos: New line?

config/schema/og.schema.yml
```diff
@@ -145,3 +145,15 @@ field.widget.settings.og_complex:
...
+  type: views_field_bulk_form
+  label: 'OG Membership bulk form'
+
+action.configuration.og_membership_add_role_action:
+  type: action_configuration_default
+  label: 'Add a role to a group member'
+
+action.configuration.og_membership_remove_role_action:
+  type: action_configuration_default
+  label: 'Remove a role to a group member'
```
idimopoulos: New line here please.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
idimopoulos: I can imagine that this is meant to be something like "If the user '$account' has access to perform the 'add|remove' 'specific-role' to a group member'".
I do not know if there are plans to add per role and add|remove action permission (at least for now) but until then, we can check the 'administer group' permission on the account since that is needed to access the admin page currently.

Though, in the future, we should add the per role action-role permission to allow these actions to be attached to custom views outside the admin area.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
```diff
@@ -0,0 +1,103 @@
...
+    return [
+      'rid' => '',
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $roles = user_role_names(TRUE);
```
idimopoulos: Where is this configuration form accessed from? this is a list of non og roles and is not called in the form build of the members management page.
And also, what does it do? The class has the user_role definition injected while I can imagine that site wide roles do not affect access to these actions. 
Can you add some commenting on this class to explain why and how you handle values from normal roles?

src/Plugin/Action/ChangeOgMembershipRoleBase.php
```diff
@@ -0,0 +1,103 @@
...
+   */
+  public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $this->configuration['rid'] = $form_state->getValue('rid');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function calculateDependencies() {
+    if (!empty($this->configuration['rid'])) {
```
idimopoulos: Again, why are you using these dependencies? Can you add some commenting to explain it?
Sorry if it is a simple thing and I did not get it.

idimopoulos (changes requested):
Good work overall.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
zerolab: This form show when one the newly added role action is triggered from the members overview page. 

I missed it in my initial review, but it does look like it is copied from `ChangeUserRoleBase`. and these roles definitely must be those of the group, not the global roles

src/Plugin/Action/ChangeOgMembershipRoleBase.php
zerolab: From `DependentPluginInterface`:

> Dependencies are saved in the plugin's configuration entity and are used to
> determine configuration synchronization order. For example, if the plugin
> integrates with specific user roles, this method should return an array of
> dependencies listing the specified roles.

The role actions are created per role, so it makes sense to add that as a dependency

src/Plugin/Action/ChangeOgMembershipRoleBase.php
zerolab: I don't think it makes sense to have per-role permissions. This should depend on "administer group" / "manage members" (from #231)

pfrenssen:
On a newly created group I see that the checkbox for the bulk operations is showing up in the last column. Traditionally this is in the first column.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
pfrenssen: I addressed this so access is only granted if the user has either the "Administer group" permission or "Manage members" inside the group that is being changed.

pfrenssen:
> @sandervd I think you need an update hook to provide the administrator role action for existing builds.

We don't need to do this, the module doesn't have a stable release yet.

src/Plugin/views/field/OgMembershipBulkForm.php
pfrenssen: So the actions to change the membership role to `member` or `non-member` are created first, and then filtered in the bulk form. It makes more sense to not make these actions in the first place. They are not useful and then we don't need to filter them either.

Instead of having an action to change the membership of a user to `non-member` it would be better to have a dedicated action `og_membership_remove`.

og.module
pfrenssen: Fixed them.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
pfrenssen: I have never seen this form show up, and I have tested it extensively. Where can we see this?

zerolab: hmm.. I could swear I saw it while doing operations on users (mind you, I am talking about `ChangeUserRoleBase`). Yet I can't find it.

config/optional/views.view.og_members_overview.yml
```diff
@@ -297,6 +297,124 @@ display:
...
+          delta_offset: 0
+          delta_reversed: false
+          delta_first_last: false
+          multi_type: separator
+          separator: ', '
+          field_api_classes: false
+          entity_type: og_membership
+          entity_field: roles
+          plugin_id: field
+        og_membership_bulk_form:
```
zerolab: Re-arrange the fields through UI and export? the bulk operation checkboxes should be in the first column. Core does it like that everywhere

zerolab:
Nice improvements on the membership actions @pfrenssen !

config/optional/views.view.og_members_overview.yml
pfrenssen: I also noticed this, I will update it today.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
pfrenssen: I found the form! It's part of the Action module and shows up when configuring actions at `admin/config/system/actions/`.

src/Plugin/views/field/OgMembershipBulkForm.php
pfrenssen: I have fixed this in `og_og_role_insert()` by checking if the newly created role is required:

```
  // Create actions to add or remove the role, except for the required default
  // roles 'member' and 'non-member'. These cannot be added or removed.
  if ($role->getRoleType() === OgRoleInterface::ROLE_TYPE_REQUIRED) {
    return;
  }
```

src/Plugin/Action/RemoveOgMembershipRole.php
pfrenssen: Good suggestion but this highlights a DX issue, why do we need to load the role in order to remove it? I solved it by introducing a new `revokeRoleById()` method, so this can be reduced to a single line of code:

```
$membership->revokeRoleById($rid)->save()
```

pfrenssen: Fixed.

og.module
pfrenssen: Fixed.

src/Plugin/Action/AddOgMembershipRole.php
pfrenssen: Good suggestion, fixed!

pfrenssen:
In OG roles are created per bundle, so it is possible for one group type to have an entirely different set of roles than another group type. This also needs to be accounted for.

pfrenssen:
While looking into the configuration form for the actions (currently in `ChangeOgMembershipRoleBase::buildConfigurationForm()`) it makes me question if the basic approach for this PR is the right one.

In OG roles are defined per bundle, so if you have for example two group types "City" and "District" then there will be two "Administrator" roles: `node-city-administrator` and `node-district-administrator`.

Regarding the actions there are two possible approaches:

1. The approach currently taken in this PR: create a separate action for each role, so there will be two actions: one for `node-city-administrator`, and one for `node-district-administrator`.

    Advantages:
    * Simpler code path.
    * Minimal configuration needed: there is only one option.

    Disadvantages:
    * In sites with lots of groups and roles the number of actions can run in the hundreds or even thousands. This is not something the Action API is designed for, I can imagine this causing performance problems for sites that use modules that depend on actions (e.g. Rules). Also any UI that displays actions will suffer from this.

2. Create a single action `administrator` and use this for both the "City" and "District".

    Advantages:
    - Reduces the number of actions to a manageable minimum.
    - Maps well to how a human administrator manages their groups: "I want to remove the 'administrator' role from this member."
    
    Disadvantages:
    - Bit more difficult to code, especially when a role is being removed we will need to find out if there are other bundles using this role before we can remove the action.
    - Configuration can become ambiguous if there are multiple bundles using the same machine name for a role but different human readable labels. What if the city administrator is labeled "The Mayor" and the district administrator "Head of district"?

Both solutions will work fine on small sites, but if there is 1 thing I learned from all my years in Drupal is that having a potentially very large number of instances of a specific type is Bad News™. There will always be some site out there that does an `EntityStorageInterface::loadMultiple()` to retrieve ALL entities, or that will do a database join against the table, and things will start to slow down and won't be trivial to fix.

@sandervd @zerolab @idimopoulos @amitaibu thoughts?

src/Plugin/Action/ChangeOgMembershipRoleBase.php
pfrenssen: Fixed it to use the OgRole instances rather than the regular user roles.

src/Plugin/Action/ChangeOgMembershipRoleBase.php
pfrenssen: Yes this makes sense, we should keep it.

config/optional/views.view.og_members_overview.yml
pfrenssen: Fixed.

zerolab:
@pfrenssen I had similar thoughts when I started unpicking OG 8.x

My gut reaction is that if you have a lot of group bundles with a lot of different roles, you have a different kind of problem.

Option 1 feels better. I am not sure the Actions API is designed to cope with the issues you describe for option 2. Also, less code is better code, no ? :)

From a UX perspective, as a city administrator (or member manager) I really care that I give a "Mayor" role, not a "Head of district" one, even if under the hood they are one and the same thing.

D

pfrenssen:
So this is how it currently looks when I have 3 group types and 1 role, the duplicated actions make the interface really unclear.

The UI from the Actions module:
actions-ui

The View Bulk Operations dropdown in the Members overview:
bulk-select-action

Now imagine how this would look if you have 10 group types with 5 roles each. Or 100 group types.

Moving to unified actions would solve this too. I think we can work around the Mayor / Head of District issue too if we are clever about it.

zerolab:
Ugh, you are right. It looks rather bad.

pfrenssen:
Last Friday I started on the tests for this PR, but in doing so I hit two bugs that popped up when deleting group content types:

* The `OgRole` entities are depending on the parent entity type and bundle, but this dependency is not declared in the config entity. I fixed this in https://github.com/Gizra/og/pull/244/commits/47be43ca36bcc5c843a894699f455c7bb5ae73a0.
* When a group content type is deleted the `og.settings` config file is not updated, and the deleted content type is still declared as a group type. This was causing errors when a new bundle is later on created with the same name as the previously deleted one. Fixed in https://github.com/Gizra/og/pull/244/commits/cec61a3ea8592ca6313287047b56b803e0d93bd8.

pfrenssen:
I've encountered another problem while working on additional test coverage in `OgRoleTest`, I have split this off to a separate PR: #259.

pfrenssen:
The current test failures are due to the fact that the `OgRole` config entities now correctly declare that they depend on an entity bundle. Apparently we have a bunch of tests that are creating roles without first creating the group bundles.

In order to not confuse this PR too much I will split this off to a separate ticket.

pfrenssen:
OK this is green and ready for a look!

amitaibu:
Awesome! I'll review this properly in a couple of days

amitaibu:
(it's a big one, so need to find enough time for it 😉 )

amitaibu:
Before diving in and just by looking at the screenshots, here's how we've done it in OG7.2:

Setting the roles is done under the OG context (`/group/node/[id]/admin/people`):

people_in_group_board1-live_08-05-2017_11_46___unio_by_harness_and_client_ _amitai_itamars-imac_ _-zsh_ _160x46

And then in one go, we can add/ remove multiple roles.

set_parameters_for_modify_og_user_roles___unio_by_harness

The advantage of doing this under the group context is that we are guaranteed to use only the relevant roles.

pfrenssen:
@amitaibu this is a really good suggestion. Doing it like this solves the final "Mayor / Head of District" problem we were discussing above.

Thanks for the feedback. I will pick this back up at the end of this week or at the beginning of next week.

zerolab:
Nice idea! Would be great if the new form is just on multi-select field. From a UX point of view the two selects in D7 are rather confusing for regular users.

amitaibu:
Yeah, in general I'd say having a peek at OG7.2  is always a good idea 😜

pfrenssen:
I can understand the reasoning for having separate fields for adding and removing roles. You might want to add a role to a bunch of users that might have various different roles already assigned to them. In that case having separate add / remove actions makes sense. Otherwise all selected users would always end up with an identical set of roles after the action is applied.

@amitaibu what is your preference?

marcorcau:
I hope you don't mind I step into the discussion.

For simple sites with only some roles shared by all groups, the current UI implemented by @pfrenssen is great, specially with the "Mayor / Head of District" problem mostly solved:

- Add administrator role
- Add first-class citizen role
- Remove administrator role
- Remove first-class citizen role

Select action -> Confirm -> Done

Maybe these actions could be kept, and also add one (Modify OG user roles as in D7) or two (Add OG user roles, Remove OG user roles) more general actions for complex sites, where context is taken into account.

amitaibu:
@budalokko I suspect your use case is that you want to have an admin be part of all the groups? If that is the case, I think there are better solutions then assigning them a role one by one. That is, you could have some custom code that will grant you access to the groups as usually you want those admin to _not_ be part of the groups.

Is this indeed your use case?

amitaibu:
@pfrenssen the separate fields are to append/ substract those roles. But I guess It could be split into two different actions, as it will be more inline with core.

pfrenssen:
The use case mentioned by @budalokko also came up during a discussion I had with @sandervd this morning.

In our project we have 1 additional group role: "moderator". We only want to have the possibility for this role to be granted / removed. The "administrator" role can not be granted through bulk operations, we have a separate UI for this to transfer ownership of a group.

So for us it would be ideal if we had a bulk operation specifically for adding / removing a single role, like how it was originally implemented by @sandervd. This saves our users from needlessly going through a second form.

But the 2-step form definitely has benefits in other use cases too. It allows to assign multiple roles, it solves the "Mayor / Head of District" problem, it avoids potentially showing roles that are not available in particular groups, and since it worked like this in D7 we should aim to support it in D8 too.

I think a good way to solve it is by providing both types of actions, but only enable the D7-style version by default in the bulk operations configuration so that the user is not overwhelmed by similar options in the dropdown. Alternatively we can split off the "simple" actions to a small contrib module like "OG single role actions".

pfrenssen:
I have some time to work on this now. I will implement the possible solution as I suggested in my previous comment. If it is not desirable to have both the simple and 2-step versions of the role actions in OG then we can still move them to a contrib module.

pfrenssen:
Hmm the bot is not triggered?

pfrenssen:
Still to do:

* Hide the 'simple' actions to add/remove roles in the member overview.
* Add test coverage for the new actions + the new methods that were added in `OgRole` and `OgMembership`.

amitaibu:
> So for us it would be ideal if we had a bulk operation specifically for adding / removing a single role

I think my [comment](https://github.com/Gizra/og/issues/264#issuecomment-301392705) about how roles with the same name, but different bundles are actually not the same. My concern about the "out of group context" is that I want to prevent granting invalid roles. That is, roles that don't apply to the user selected, as `OgRoles` are referenced from `OgMemebership` not from the `User`.

pfrenssen:
We discussed this here before, it is what we call the "Mayor / Head of District" problem, as in we can have a role that is called 'administrator' but for one group type "City" the administrator is the Mayor, and for another group type "District" the administrator is the Head of District.

It is more a problem of being able to distinguish which role is meant since we are showing the human readable name. This is solvable though, we can check if there are multiple different human readable names in use for a single role name, and in that case show the machine name.

Now this is all very theoretical, and assumes that we are managing members of multiple group types at the same time in the same view. That is not something OG provides out of the box, we have a separate "Members" tab for each group type.

I think if sites need to add a UI for managing multiple group types in a single overview, then they probably have custom requirements specific to their use case and it's probably best to extend the existing plugins and customize them according to their own needs.

On a side note, I just had a look at the code that adds roles to memberships and it actually allows to add invalid roles at the moment. `Membership::addRole()` doesn't check if the role that is passed in is valid and matches the group. It is also possible to add roles and then change the group afterwards. That will also make the roles invalid. We should add a check on presave to ensure all roles are valid. I will make a separate ticket for this.

pfrenssen:
I have split off the checking of the roles on memberships to #266 - this will enforce it on API level by throwing an exception if a membership with invalid roles is saved.

amitaibu:
> It is more a problem of being able to distinguish which role is meant since we are showing the human readable name. This is solvable though, we can check if there are multiple different human readable names in use for a single role name, and in that case show the machine name.

Indeed, It's the kind of complexity I'd prefer to avoid from having in OG core, which seems to be dealing with a very specific case. The "in group context" bulk operation seems to be one that solves the problem, without needing for extra complexity.

@pfrenssen sorry, to be sure I understand you -- are you proposing to keep "non-group context" bulk operation or not?

pfrenssen:
They're still in right now, but I'm fine to move them to a separate small contrib module. I was not planning on adding the complexity needed to make the distinction between different roles, because I don't think managing members of multiple group types at once is something OG should offer out of the box.

pfrenssen:
Travis doesn't start the tests again :-1:

pfrenssen:
@amitaibu just so you are on the same line, the problem of showing the human readable name of the single role actions _does not occur in the current implementation of this PR_. It only occurs in extremely rare cases, all three of these conditions need to be met for it to occur:

1. The site needs to share roles with the same machine name but different human readable names across different group types.
2. The site should create a custom membership overview that manages multiple group types at once in the same view.
3. The site maintainer _has to willfully choose the wrong action plugin_ instead of using the default which works fine.

In 99% of the use cases the single role plugin works just fine, and in the 1% where it doesn't work, the site admin just needs to switch to the default plugin and the problem is solved.

And it would also help us, as well as @budalokko (and I think @zerolab too since he gave a thumbs up at @budalokko's comment).

I suggest to try it out :) it really works well for group type that have a simple set of 1 or 2 roles.

pfrenssen:
OK so this has grown dramatically, it's reaching 3000 lines now. I'm going to split off some changes that are not fully in scope to separate PRs so this becomes a bit easier to review.

amitaibu:
@pfrenssen thanks for the clarification in https://github.com/Gizra/og/pull/244#issuecomment-301518814 . Sorry, I had the impression that duplication was the common case. This makes me relaxed again! :)

pfrenssen:
I am currently in the process of splitting off all off-topic work into separate PRs so this becomes easier to review. This will take some time since I have been assigned to a different ticket at work which was high priority.

I am dedicated to see this through, so I will try to revisit this daily and make some progress. Once this is fully on-topic I will add back the "Needs review" label.

amitaibu:
Much appreciated!

pfrenssen:
I went throught the entire patch to check what can still be split off in separate PRs before submitting this for review:

* New method `OgMembership::revokeRoleById()`
* New method `OgMembership::hasRole()`
* Remove duplicated documentation on `OgRole::getRoleType()` and `OgRole::setRoleType()`
* It should be possible to remove a group that has default roles assigned to it (ref. the code in `OgRole::onDependencyRemoval()`).
* New method `OgRole::loadRolesByGroupType()`
* New method `OgRole::isDefaultRole()`

pfrenssen:
Alright! This has taken some time but I'm through all the tasks. When the split off PRs are all green and merged then this is finally fully on topic and ready for review.

pfrenssen:
All split offs are merged, so this is finally ready for review!

amitaibu:
Oh my, 40 files! Lovely, I'll review -- but it might take me a few days ;)

og.module
```diff
@@ -306,3 +313,64 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
...
+  if ($role->getRoleType() === OgRoleInterface::ROLE_TYPE_REQUIRED) {
+    return;
+  }
+
+  $add_id = 'og_membership_add_single_role_action.' . $role->getName();
+  if (!Action::load($add_id)) {
+    $action = Action::create([
+      'id' => $add_id,
+      'type' => 'og_membership',
+      'label' => t('Add the @label role to the selected members', ['@label' => $role->label()]),
```
claudiu-cristea: Unfortunately, we cannot use `$role->label()` here because we might face this case: 2 roles have the same name:

- `node-software-tester`, with label "Software Tester"
- `node-hardware-tester`, with label "Hardware Tester"

The role that comes first will steal the label of the action. If  `node-software-tester` is created first, the action label will be "Add the Software Tester role to the selected members". But this action is used also for `node-hardware-tester`. In the context of the 2nd role, we'll still show the label from the 1st role. The action will run correctly, only the label of the action will be wrong.

For this reason we should use `$role->getName()` instead. I know, it's a machine-name but, at least, is not misleading.

I've created a PR for this change https://github.com/ec-europa/og/pull/2.

og.module
```diff
@@ -306,3 +313,64 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
...
+      'plugin' => 'og_membership_add_single_role_action',
+    ]);
+    $action->trustData()->save();
+  }
+  $remove_id = 'og_membership_remove_single_role_action.' . $role->getName();
+  if (!Action::load($remove_id)) {
+    $action = Action::create([
+      'id' => $remove_id,
+      'type' => 'og_membership',
+      'label' => t('Remove the @label role from the selected members', ['@label' => $role->label()]),
```
claudiu-cristea: Same applies here as for `og_membership_add_single_role_action`.

claudiu-cristea (changes requested):

og.module
pfrenssen: Very good find, since roles for different groups can have different labels this makes sense. We cannot use the labels unless we know on which group(s) they apply.

Unfortunately we do not have the right context available to know on which group(s) the views bulk operations take effect. In the default view that we supply with OG the group type and group ID are part of the contextual arguments, but we cannot rely on these being always available. Views are very complex beasts, it is possible to limit the group using regular filters, or even `hook_views_query_alter()`.

I agree that using the role name (machine name) is not the prettiest solution but it will at least be 100% clear and correct.

amitaibu:
Sorry it took me some time -- I'm on vacation 😸  Things looks great, but found a few stuff on manual tests that should probably be addressed:

1) As UID 1, after I block myself, I cannot unblock (it's a silly scenario, but UID 1 should always have access):

members___drush_site-install

2)   `add roles to the selected members`

members___drush_site-install

results with `Fatal error: Call to undefined method Drupal\og\Entity\OgRole::loadRolesByGroupType() in /.../modules/og/src/Form/OgAddMultipleRolesForm.php on line 27`

3) When an operation isn't valid I think that "no access" is a misleading message (.e.g `No access to execute Block the selected membership(s) on the OG membership). Maybe, `Can't execute ... as there selected OG membership do not permit it`?

4) I'd expect to have also a "Set pending" action, like we have for the `active` and `blocked`

5) We shouldn't allow removing or changing the status of the group owner

src/Plugin/Action/ApprovePendingOgMembership.php
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function execute(OgMembership $membership = NULL) {
+    if (!$membership) {
+      return;
+    }
+    $membership->setState(OgMembershipInterface::STATE_ACTIVE)->save();
```
amitaibu: maybe if the membership is already `active`, we could skip the save, and save a little time.

pfrenssen:
Thanks for the review! Those all look like really good points. I'll address them when I have some free time!

pfrenssen:
> 1) As UID 1, after I block myself, I cannot unblock (it's a silly scenario, but UID 1 should always have access):
>
> members___drush_site-install

I looked into this and was surprised that there is actually test coverage guaranteeing that UID1 can always unblock users. Then I saw that you accidentally mixed up two states, the correct action to unblock users is "Unblock the selected membership(s)" but you chose "Approve the pending membership(s)" - the membership is not pending so it cannot be approved, so this action cannot be executed. The error message is not helpful indeed, but we cannot easily change that unfortunately (see point 3).

> 2)   `add roles to the selected members` results with `Fatal error: Call to undefined method Drupal\og\Entity\OgRole::loadRolesByGroupType() in /.../modules/og/src/Form/OgAddMultipleRolesForm.php on line 27`

Fixed! This method had been renamed recently. This wasn't caught because these forms are untested, I thought it wasn't "important enough" since this is quite trivial code :-/ I still think that adding tests for these forms is not worth the time needed to write them, but then we will not catch regressions :(

> 3) When an operation isn't valid I think that "no access" is a misleading message (.e.g `No access to execute Block the selected membership(s) on the OG membership). Maybe, `Can't execute ... as there selected OG membership do not permit it`?

This is a very good point, these messages are VERY unhelpful. From an end user's point of view it has very little to do with "access" in most cases. The problem is that we have no control over it. The message is hardcoded in Views in `BulkForm::viewsFormSubmit()`. I had a look at how this could possibly be solved and came up with two possible approaches, but both require changes in Views as well as another core component:

1. Views gets the access result from the `Action` plugins so we could add another entry to the `Action` annotation that contains an optional customized error message. This is not ideal though since it would not be very flexible. There might be different reasons for denying access, but the message would be static. We also cannot easily pass dynamic arguments in the message like `This action cannot be executed because the %role role is not granted.`.
2. A possible solution could be to enrich the actual `AccessResult` object with a reason why access is denied. This can then be passed through the `Action` plugin to `Views` who can then use it. I think this would be a very nice thing to have, it is useful for other places where `AccessResult` objects are used, and would even allow to pass on _multiple_ reasons why access would be denied or granted. But it feels that this would require a lot of work in core to get this in.

So for the moment there's not much we can do unfortunately to improve these messages.

src/Plugin/Action/ApprovePendingOgMembership.php
pfrenssen: This is already accounted for, the access check will not allow the action to be executed if it is already active.

pfrenssen:
> 4. I'd expect to have also a "Set pending" action, like we have for the active and blocked.

Implemented this. I also refactored the other 2 actions that change membership state so they share the same base class which contains the logic.

I still need to do point 5: preventing to change the membership state of the group owner. I'll get to this later today or tomorrow.

pfrenssen:
Implemented point 5 and added full test coverage for the group owner, both when the `group_manager_full_access` configuration option is enabled or not.

I think that "group manager" is a misleading term in this context, it is common to have a "manager" role and then there can be multiple group managers. It would be better to rename this to `group_owner_full_access`. I will make a separate issue for this.

For some reason the test results are not being linked back here, but they are running on Travis. Here is the build for the latest commit: https://travis-ci.org/amitaibu/og/builds/264358618

amitaibu:
Didn't forget about this -- I'll review this next week, when I'm back :)

amitaibu:
Wonderful work!! Thanks you so much 😀

PR #246, opened 2017-03-21
GitHub issue #245 by fdeimos: og_get_groups_by_user(, ...) returns th…

### LINKED ISSUE #245: og_get_groups_by_user($anonymous_user, ...) returns the current user's groups instead of anonymous's
https://github.com/Gizra/og/blob/7.x-2.x/og.module#L3562
Second param for entity_metadata_wrapper is UID, not account entity. 
For anonymous user UID is 0. As result, 'data' property for $wrapper is FALSE, but there should be anonymous user object.
When `$og_memberships = $wrapper->{'og_membership__' . OG_STATE_ACTIVE}->value();` is called (next line in code) then after a few steps og_get_entity_groups() is called and $entity param is FALSE. So, it takes current user.
https://github.com/Gizra/og/blob/7.x-2.x/og.module#L2305-L2307

fdeimos:
…e current user's groups instead of anonymous's

### DIFF
--- a/og.module
+++ b/og.module
@@ -3559,7 +3559,7 @@ function og_get_groups_by_user($account = NULL, $group_type = NULL) {
   $gids = array();
 
   // Get all active OG membership that belong to the user.
-  $wrapper = entity_metadata_wrapper('user', $account->uid);
+  $wrapper = entity_metadata_wrapper('user', $account);
   $og_memberships = $wrapper->{'og_membership__' . OG_STATE_ACTIVE}->value();
   if (!$og_memberships) {
     return;


amitaibu:
@alvar0hurtad0 please review.

alvar0hurtad0:
@fdeimos entity_metadata_wrapper accept an entity_id as the second parameter.

Maybe a better workaround for this could be check if $account->uid is 0 once we get the account and them just return because anonymous should not have any groups.

Something like: 

```
function og_get_groups_by_user($account = NULL, $group_type = NULL) {
  if (empty($account)) {
    global $user;
    $account = $user;
  }

  if ($account->uid == 0) {
    // Anonymous user should not have groups.
    return;
  }
```

jmorahan:
Too late at that point, if $account was 0 it's now already $user. You would have to check immediately if $account is 0 or '0', no?

alvar0hurtad0:
@jmorahan Indeed you're right.

so we can check at the beginning if current user is anonymous and account is empty to just return.

jmorahan:
No, that case is already handled correctly. The problem occurs when the current user is *not* anonymous. The code checks if $account is empty, and if so, overwrites it with $user, which is *not* the anonymous user and therefore returns that user's groups instead of the expected none.

alvar0hurtad0:
In this case I didn't understand the issue.

Can you provide the steps to reproduce the issue, please?

jmorahan:
Log in as a user who's in one or more groups, and call `og_get_groups_by_user(0);` - you expect no groups, because user 0 is in no groups. Instead you get the groups of the user you're logged in as.

jmorahan:
Sorry, I read that completely wrong! That's what I get for trying to review patches on my phone. I was looking at the wrong function - you're right, checking for $account->uid in og_get_groups_by_user() should be fine, because *that* function only accepts an account, not a uid.

Sorry for the confusion.

alvar0hurtad0:
@jmorahan thanks a lot for all, :D

No problem, let me check in deep.

Thanks again for getting involved.

alvar0hurtad0:
I've been testing and passing an account object on the second param looks like the behavior is the expected so I think there's no issue.

I called the function with this code:

```
$user = user_load(0);
og_get_groups_by_user($user);
```

And worked as expected.

@fdeimos @jmorahan can you check it too?

jmorahan:
How did you run that code though @alvar0hurtad0 ? What user was logged in, and were they in any groups?

Try this: create a group node as user 1 (so that user 1 is a member of that group), then run:

    $GLOBALS['user'] = user_load(1);
    $account = user_load(0);
    $groups = og_get_groups_by_user($account);
    print_r($groups);

I get the group that user 1 created. Don't you?

alvar0hurtad0:
Indeed you're right

I think the best option to solve the problem is just check if the account is passed and the id is 0

```
function og_get_groups_by_user($account = NULL, $group_type = NULL) {
  if (isset($account->uid) && $account->uid == 0) {
    // In general there can't be groups for anon users.
    return;
  }
....
```

claudiu-cristea:

PR #250, opened 2017-04-05
Allow global administrators to reference any group

zerolab:
If a user has the global 'administer group' permission, they should be able to reference any group.

Currently, this is only possible when they do not belong to any group. The moment they are a member of a group, referencing one they don't belong to fails with "The entity (entity_type:id) cannot be referenced)

### DIFF
--- a/src/Plugin/EntityReferenceSelection/OgSelection.php
+++ b/src/Plugin/EntityReferenceSelection/OgSelection.php
@@ -77,6 +77,11 @@ class OgSelection extends DefaultSelection {
       $query->condition($bundle_key, $bundles, 'IN');
     }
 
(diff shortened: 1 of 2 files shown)


amitaibu:
Thanks @zerolab . Seems a test need to be updated

zerolab:
Found the issue and updated the tests.

The Travis build failed because of unrelated code sniffer violations, but I think that should be tackled in a follow-up PR

zerolab:
With #181 almost here, I think this PR is obsolete.

https://github.com/Gizra/og/pull/181/files#diff-669ced81d1fae6b8b9a940fcf8dc8d35R87 does the same thing.

PR #253, opened 2017-04-13
Default argument for current user memberships.

tombola:
Issue #254:

Default argument to populate the contextual filter (eg groups_audience) used by the view.

### DIFF
--- a/src/Plugin/views/argument_default/Membership.php
+++ b/src/Plugin/views/argument_default/Membership.php
@@ -146,4 +146,19 @@ class Membership extends ArgumentDefaultPluginBase implements CacheableDependenc
     return [];
   }
 
+  /**
+   * Returns the group from the runtime context.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface|null
+   *   The group from context if found.
+   */
+  protected function getGroup() {
+    $contexts = $this->ogContext->getRuntimeContexts(['og']);
+    if (!empty($contexts['og']) && $group = $contexts['og']->getContextValue()) {
+      if ($group instanceof ContentEntityInterface) {
+        return $group;
+      }
+    }
+  }
+
 }
--- a/src/Plugin/views/argument_default/Membership.php
+++ b/src/Plugin/views/argument_default/Membership.php
@@ -41,7 +41,7 @@ class Membership extends ArgumentDefaultPluginBase implements CacheableDependenc
    *
    * @var \Drupal\Core\Session\AccountInterface
    */
-  protected $user;
+  protected $ogUser;
 
   /**
    * Constructs a new User instance.
@@ -64,7 +64,7 @@ class Membership extends ArgumentDefaultPluginBase implements CacheableDependenc
 
     $this->ogContext = $og_context;
     $this->ogMembership = $og_membership;
-    $this->user = $og_user;
+    $this->ogUser = $og_user;
   }
 
   /**
@@ -139,7 +139,7 @@ class Membership extends ArgumentDefaultPluginBase implements CacheableDependenc
    *   An array of groups, or an empty array if no group is found.
    */
   protected function getCurrentUserGroupIds($entity_type = 'node') {
-    $groups = $this->ogMembership->getUserGroupIds($this->user);
+    $groups = $this->ogMembership->getUserGroupIds($this->ogUser);
     if (!empty($groups) && isset($groups[$entity_type])) {
       return $groups[$entity_type];
     }
--- a/src/Plugin/views/argument_default/Membership.php
+++ b/src/Plugin/views/argument_default/Membership.php
@@ -1,0 +1,149 @@
+<?php
+
+namespace Drupal\og\Plugin\views\argument_default;
+
+use Drupal\Core\Cache\Cache;
+use Drupal\Core\Cache\CacheableDependencyInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Drupal\og\MembershipManagerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Default argument plugin to provide the group from the current context.
+ *
+ * @ViewsArgumentDefault(
+ *   id = "og_group_membership",
+ *   title = @Translation("Group memberships from current user")
+ * )
+ */
+class Membership extends ArgumentDefaultPluginBase implements CacheableDependencyInterface {
+
+  /**
+   * The OG context provider.
+   *
+   * @var \Drupal\Core\Plugin\Context\ContextProviderInterface
+   */
+  protected $ogContext;
+
+  /**
+   * The OG membership manager.
+   *
+   * @var \Drupal\og\MembershipManagerInterface
+   */
+  protected $ogMembership;
+
+  /**
+   * The user to be evaluated.
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected $user;
+
+  /**
+   * Constructs a new User instance.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $og_context
+   *   The OG context provider.
+   * @param \Drupal\og\MembershipManagerInterface $og_membership
+   *   The OG membership manager.
+   * @param \Drupal\Core\Session\AccountInterface $og_user
+   *   The user to be evaluated.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ContextProviderInterface $og_context, MembershipManagerInterface $og_membership, AccountInterface $og_user) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->ogContext = $og_context;
+    $this->ogMembership = $og_membership;
+    $this->user = $og_user;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('og.context'),
+      $container->get('og.membership_manager'),
+      $container->get('current_user')->getAccount()
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function defineOptions() {
+    $options = parent::defineOptions();
+    $options['og_group_membership'] = ['default' => ''];
+
+    return $options;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
+    $form['og_group_membership'] = [];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getArgument() {
+    // Currently restricted to node entities.
+    return implode(',', $this->getCurrentUserGroupIds('node'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge() {
+    return Cache::PERMANENT;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return ['og_role'];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    $group = $this->getGroup();
+    if ($group instanceof ContentEntityInterface) {
+      $tag = $group->getEntityTypeId() . ':' . $group->id();
+      return Cache::buildTags('og-group-content', [$tag]);
+    }
+    return [];
+  }
+
+  /**
+   * Returns groups that current user is a member of.
+   *
+   * @return array
+   *   An array of groups, or an empty array if no group is found.
+   */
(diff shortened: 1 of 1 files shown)


src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Returns groups that current user is a member of.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface|null
+   *   The group, or NULL if no group is found.
+   */
+  protected function getCurrentUserGroups($entity_type = 'node') {
+    $user = \Drupal::currentUser();
+    $membership_manager = \Drupal::service('og.membership_manager');
```
zerolab: The currentUser and OG membership manager services need to be injected

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,127 @@
...
+    $groups = $membership_manager->getUserGroupIds($user);
```
zerolab: `\Drupal::currentUser()` returns an `AccountProxyInterface`, the method needs `AccountInterface`, so.. `\Drupal::currentUser()->getAccount()`

amitaibu:
Thanks! Test seems to still be red, can you please check it.

tombola:
Added the final parameter comments that were being flagged by test, and rebased to clean up history and add drupal style commit author.

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,149 @@
...
+  public function buildOptionsForm(&$form, FormStateInterface $form_state) {
+    $form['og_group_membership'] = [];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getArgument() {
+    // Currently restricted to node entities.
+    return implode(',', $this->getCurrentUserGroupIds('node'));
```
amitaibu, line 106: I think we could make it configurable - and expose the group related entities

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,149 @@
...
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $og_context
+   *   The OG context provider.
+   * @param \Drupal\og\MembershipManagerInterface $og_membership
+   *   The OG membership manager.
+   * @param \Drupal\Core\Session\AccountInterface $og_user
+   *   The user to be evaluated.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ContextProviderInterface $og_context, MembershipManagerInterface $og_membership, AccountInterface $og_user) {
```
amitaibu, line 62: `$og_user` => `$user`

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,149 @@
...
+   */
+  public function getCacheMaxAge() {
+    return Cache::PERMANENT;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return ['og_role'];
```
amitaibu, line 125: is this correct? Shouldn't it be per user?

src/Plugin/views/argument_default/Membership.php
tombola: good point. amended

src/Plugin/views/argument_default/Membership.php
tombola: My understanding was that though memberships vary by user, the user's role within each of those groups serves as a proxy for that that membership, whilst possibly avoiding clearing caches based on every user's individual memberships.

@zerolab may have comment.

tombola:
Could use $user instead if preferred. I selected $ogUser to emphasise intended usage within an organic groups (membership) context and to maintain consistency/predictability with the other properties.

amitaibu:
> I selected $ogUser to emphasise intended usage within an organic groups (membership) context and to maintain consistency/predictability

Ok, got your point. Lets keep as is.

src/Plugin/views/argument_default/Membership.php
tombola: That does sound a worthwhile aim.

I would likely need to return to this at a later date if I were to help with that though.

amitaibu: 👍

src/Plugin/views/argument_default/Membership.php
zerolab: I believe `og_role` was the best next thing. User is good, however this is a tricky one as what changes is the membership, not the user.
The more I think about it, the more it feels we need an og membership cache context.

zerolab: Having dug more into this, I think this is a good context. `og_role` takes in consideration the user memberships and the roles held in the corresponding groups.

If you think this is overkill, then we do need an `OgMembershipCacheContext` that does is similar to `OgRoleCacheContext`.

Opinions?

amitaibu: I guess in that case the name `og_role` isn't great. probably `og_membership_and_roles` would be more descriptive.

@gh640 care to create a PR rename [that](https://github.com/Gizra/og/blob/9d81808c1db299267f67d5143dd32c81adc6c1c3/src/Cache/Context/OgRoleCacheContext.php#L30) context?

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    $group = $this->getGroup();
```
zerolab, line 132: The `getGroup()` method was removed, so this will fail.

```
-  protected function getGroup() {
-    $contexts = $this->ogContext->getRuntimeContexts(['og']);
-    if (!empty($contexts['og']) && $group = $contexts['og']->getContextValue()) {
-      if ($group instanceof ContentEntityInterface) {
-        return $group;
-      }
-    }
-  }
```

src/Plugin/views/argument_default/Membership.php
pfrenssen: I agree that this cache context is the best thing we have right now for this purpose, it is one level too granular, so the context will be more fragmented than strictly needed, but it will work.

src/Plugin/views/argument_default/Membership.php
zerolab: @tombola this now affects our project. Please update?

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * The user to be evaluated.
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected $ogUser;
+
+  /**
+   * Constructs a new User instance.
```
MPParsley: User => Membership

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,164 @@
...
+      $tag = $group->getEntityTypeId() . ':' . $group->id();
+      return Cache::buildTags('og-group-content', [$tag]);
+    }
+    return [];
+  }
+
+  /**
+   * Returns groups that current user is a member of.
+   *
+   * @return array
```
MPParsley: Missing parameter in this dochead:

   * @param string $entity_type
   *   The entity type, defaults to 'node'.

MPParsley (commented):
Added some small comments wrt the phpdocs.

Since a user with the permission "administer group" has access to all of the groups, wouldn't it be useful to add an option to list all groups for users with this permission?

With exception for @amitaibu's remark on making the entity type configurable this seems to be ready.

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,164 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Drupal\og\MembershipManagerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Default argument plugin to provide the group from the current context.
```
MPParsley: Default argument plugin to provide the **group memberships** from the current context.

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Drupal\og\MembershipManagerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Default argument plugin to provide the group memberships from the current context.
```
MPParsley, line 16:
```suggestion
 * Default argument plugin to provide the group memberships from the current
 * context.
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+  public function getCacheMaxAge() {
+    return Cache::PERMANENT;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    // This cache context is the best thing we have right now.
+    // og_role takes in consideration the user memberships and
```
MPParsley, line 121:
```suggestion
    // og_role takes in consideration the user memberships and
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+    return Cache::PERMANENT;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    // This cache context is the best thing we have right now.
+    // og_role takes in consideration the user memberships and 
+    // the roles held in the corresponding groups, and while it
```
MPParsley, line 122:
```suggestion
    // the roles held in the corresponding groups, and while it
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    // This cache context is the best thing we have right now.
+    // og_role takes in consideration the user memberships and 
+    // the roles held in the corresponding groups, and while it 
+    // is one level too granular, i.e. the context will be more
```
MPParsley, line 123:
```suggestion
    // is one level too granular, i.e. the context will be more
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Plugin\Context\ContextProviderInterface $og_context
+   *   The OG context provider.
+   * @param \Drupal\og\MembershipManagerInterface $og_membership
+   *   The OG membership manager.
+   * @param \Drupal\Core\Session\AccountInterface $og_user
```
MPParsley, line 59:
```suggestion
   * @param \Drupal\Core\Session\AccountInterface $user
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+   *   The user to be evaluated.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ContextProviderInterface $og_context, MembershipManagerInterface $og_membership, AccountInterface $og_user) {
```
MPParsley, line 62:
```suggestion
  public function __construct(array $configuration, $plugin_id, $plugin_definition, ContextProviderInterface $og_context, MembershipManagerInterface $og_membership, AccountInterface $user) {
```

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+
+    $this->ogContext = $og_context;
+    $this->ogMembership = $og_membership;
+    $this->ogUser = $og_user;
```
MPParsley, line 67:
```suggestion
    $this->ogUser = $user;
```

MPParsley:
@tombola, could you add me as an author so I can fix the remarks/tests?

src/Plugin/views/argument_default/Membership.php
```diff
@@ -0,0 +1,172 @@
...
+   * Returns groups that current user is a member of.
+   *
+   * @param string $entity_type
+   *   The entity type, defaults to 'node'.
+   *
+   * @return array
+   *   An array of groups, or an empty array if no group is found.
+   */
+  protected function getCurrentUserGroupIds($entity_type = 'node') {
+    $groups = $this->ogMembership->getUserGroupIds($this->ogUser);
```
MPParsley, line 150: getUserGroupIds has been refactored to use the user_id instead of the user object:

```suggestion
    $groups = $this->ogMembership->getUserGroupIds($this->ogUser->id());
```

Deprecated: Passing an account object is deprecated in og:8.1.0-alpha4 and is removed from og:8.1.0-beta1. Instead pass the user ID as an integer value. See https://github.com/Gizra/og/issues/542

MPParsley:
Closing in favor of #565.

src/Plugin/views/argument_default/Membership.php
MPParsley: @zerolab, getGroup() is there. Is this remark still valid?

zerolab: I added it in f6e0e52, so sorted

PR #258, opened 2017-04-28
Remove obsolete schema hook

pfrenssen:
We still have the old schema hook from the D7 version in the install file. We don't need it any more.


zerolab:
👍 I was wondering why we have the unused schema, especially that roles are now entities

amitaibu:
Thanks :)

Issue #259, opened 2017-05-02
Possibility to "override" group roles and permissions in the D8 version


pfrenssen:
In Drupal 7 we had the possibility to "override" roles and permissions on a per-group basis. Here is a screencast that introduces the concept: [D7 Organic Groups pt 11 - Override Roles & Permissions](https://www.youtube.com/watch?v=otSD-uDsu9o).

To get access this functionality the 'OG access control' submodule needed to be enabled and this unlocked configuration forms to manage roles and permissions for individual groups.

This resulted in a lot of added complexity in the D7 code base, see for example `og_roles()`. Some of it has crept into D8 already (`OgRole::setGroupId()` and `OgRole::getGroupId()`) but this is not used anywhere at the moment.

We need to decide first if we want to keep this functionality in D8. To me it sounds like an edge case, and one that can also be trivially solved by creating a separate group type for the group(s) that require custom roles and permissions. A well-architected website should not have a lot of divergence in roles and permissions across different group types.

If we do decide to keep this functionality (e.g. because a large number of the D7 user base are making use of this feature) then we need to figure out how to implement this. I would like to keep the added complexity to a minimum, so maybe a good approach would be to create an additional config entity type `OverriddenOgRole` which extends `OgRole`. It can have its own handling of the role ID etc, but for the rest it can inherits almost everything from `OgRole`.

amitaibu:
The feature should stay (in DrupalCon conversation several people confirmed they were using it to give different permissions to different roles).

I have an idea on how to allow a scalable solution via having `OgRole` ( as `ConfigEntity`) and `OgRoleContent` (as `ContentEntity`), and have `OgMembership` two ER properties  -- one to reference the config and another to the overridden content.

Lets discuss this before we start coding ;)

bradbulger:
Is there any practical way to get movement on this if we are unable to do the work ourselves? (Not yet, anyway.) I work with several sites that rely on this feature, and which will not be able to move to Drupal 8 without it.

PR #260, opened 2017-05-02
Don't allow to set a group ID on a role, it breaks everything

pfrenssen:
In `OgRole` we have two methods that allow to get and set the group ID. Calling this will inject the group ID in the `OgRole` entity ID. This actually causes the ID to be mangled and will be invalid for the remainder of the code base which expects the ID to consist of `{$entity_type_id}-{$bundle_id}-{$role_name}`. Once this group ID is set on a role and the role is saved practically everything related to roles breaks, resulting in errors like:

```
Call to a member function getPermissions() on null in OgAccess.php:203
```

The only way to return to a normal working state is to delete the role from the database and recreate it without the group ID.

These methods are there because in D7 we had the possibility to override roles on a per-group basis (ref #259). However, if we want to keep this functionality in D8 we will probably not implement by mangling the ID but by providing a separate config entity for overridden roles.

These broken methods are currently not used anywhere so we can safely remove them.

Once we have made a decision in #259 we can make something that actually works :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,13 +52,7 @@ before_script:
 
   # Download Drupal console so we can run the test for it. Skip this for the
   # coding standards test.
-
-  # When downloading DrupalConsole using Composer an old version is supplied.
-  # Sepficing the correct version and the versions of DrupalComposer dependancy
-  # will fix the problem for now. Need to be removed when https://goo.gl/i6rVhT
-  # will be merged.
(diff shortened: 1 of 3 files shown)


pfrenssen:
There were some errors during the installation of Drupal Console. These were caused by https://github.com/hechoendrupal/drupal-console-extend-plugin/pull/8, a fix has been released in the meantime. I have updated the Travis config.

amitaibu:
Yeah, this one isn't baked yet, so we can remove it for now.

PR #261, opened 2017-05-03
Declare the dependency between `OgRole` entities and their parent group

pfrenssen:
This has been split off from #244. It was originally included in the scope of that ticket but this has uncovered quite a few small bugs in our tests, so it has been split off to not confuse the parent issue.

`OgRole` entities are depending on the parent entity type and bundle, but this dependency is not declared in the config entity. We were instead cleaning up the roles manually whenever a parent bundle is deleted. With the dependency in place the cleaning up of orphaned roles is taken care of automatically by the configuration management system.

Due to the dependency not being declared it was possible to accidentally create orphaned roles, by e.g. making a typo in the group type. This is something that can happen easily, as evidenced by 8 of our tests being affected by this!

It might also cause problems when importing / syncing configuration since the exact order in which the config entities are updated depends on a correct dependency chain.

### DIFF
--- a/src/Entity/OgRole.php
+++ b/src/Entity/OgRole.php
@@ -337,4 +337,15 @@ class OgRole extends Role implements OgRoleInterface {
     return \Drupal::service('og.access');
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function calculateDependencies() {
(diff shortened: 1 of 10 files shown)


amitaibu:
Thanks!

PR #266, opened 2017-05-15
Validate that roles are valid for a membership before saving it

pfrenssen:
This has been split off from PR #244 - @amitaibu remarked:

> My concern about the "out of group context" is that I want to prevent granting invalid roles. That is, roles that don't apply to the user selected, as `OgRoles` are referenced from `OgMembership` not from the `User`.

It is currently possible to save memberships that have roles from a different group type, either by calling `OgMembership::addRole()` with the wrong `OgRole` entity, or by changing the group type of an existing membership by calling `OgMembership::setGroup()` with a different group type.

This PR enforces valid roles on API level by validating the roles in `OgRole::preSave()`.

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -134,6 +134,11 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   public function getGroup() {
     $entity_type = $this->get('entity_type')->value;
     $entity_id = $this->get('entity_id')->value;
+
+    if (empty($entity_type) || empty($entity_id)) {
+      return NULL;
+    }
+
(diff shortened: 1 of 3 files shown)


pfrenssen:
Going to extend this a bit, going to add a method `isRoleValid()` so that this code can be reused.

pfrenssen:
Added the `isRoleValid()` method, and also fixed a bug uncovered during testing: if `OgMembership::getGroup()` was called when no group was set yet on the membership, it was throwing a `PluginNotFoundException`. The expected behaviour is that it returns `NULL` in that case.

amitaibu:
Lovely, thanks!

PR #268, opened 2017-05-17
A bundle should no longer be marked as a group type after it is deleted

pfrenssen:

Currently when an entity type is declared as a group type and then is subsequently deleted, the configuration is not updated, and the deleted group type is still marked as a group.

We should make sure the list of group types is kept up to date and no orphaned entries stay behind when the parent entity type is deleted.


pfrenssen:
I made two commits: the first is the failing test that proves the bug is real, and the second is the fix.

amitaibu:
Merged, thanks!

PR #269, opened 2017-05-18
Type hint on `OgRoleInterface` instead of on `OgRole`

pfrenssen:
This is split off from PR #244.

We should always type hint on the interface instead of the implementation if possible. This will allow developers to customize and extend our entities and services.


amitaibu:
Thanks.

PR #270, opened 2017-05-18
Fix random failure in OgRoleManagerTest

pfrenssen:
During the work on #244 a random failure in `OgRoleManagerTest` was discovered. Entities are not guaranteed to be returned in the same order every time, so we should sort the results before comparing them with the expected values.

For an example of this random failure occurring, check build https://travis-ci.org/Gizra/og/builds/226714806 - in jobs 804.1, 804.2 and 804.6 this test failed, in the others it was fine.


amitaibu:
Thanks 👍

PR #271, opened 2017-05-28
Introduce OgMembershipInterface::revokeRoleById()

pfrenssen:
Currently if we want to revoke a role from a membership we need to provide the full `RoleInterface` object to `OgMembership::revokeRole()`. Let's introduce a new method to be able to delete a role when we only have the ID available. This will make sure we don't unnecessarily need to load the role from the database in order to delete it from a membership.


pfrenssen:

amitaibu:
Thanks :)

PR #272, opened 2017-05-28
Introduce OgMembershipInterface::hasRole()

pfrenssen:
It would be great to have a handy little method to check if a certain role exists on a membership.


pfrenssen:
This was split off from #244.

amitaibu:
Thanks!

PR #267, opened 2017-05-16
Add missing methods to `OgRoleInterface`

pfrenssen:
This has been split off from #244.

There are a couple of methods in `OgRole` that are not declared in `OgRoleInterface`. This PR adds the missing methods.

PR #273, opened 2017-05-28
Remove duplicate documentation

pfrenssen:
In #267 we added some missing methods to `OgRoleInterface`. Now we can replace the documentation on the class itself with `{@inheritdoc}`.


pfrenssen:
This was split off from #244.

amitaibu:
👍

PR #279, opened 2017-06-13
Membership manager static cache misbehaves

idimopoulos:
So I noticed the following wrong behaviour in the Membership manager related to the static cache.
Let's say we have 3 entity types, A, B, C.
A and B are group content both pointing to group C.
Now, since A and B are different entity types, they have different base tables so they have their own unique Ids.

So let's create A1 with id 1 and B1 with id 1. A1 points to C1 with Id 1 and B1 points to C2 with id 2.
If that creation happens at the same request and the groups are requested, the membership is using the identifier `<method name>:$entity->id():<group_type(optional)>:<group_bundle(optional)>`.
In the above case it will be something like `OgMemberhsip::getGroupIds:1::`.

The problem is that when the B1 group will be requested, the C1 will be returned instead of C2.
This is because the content's entity type Id is not taken into account.

I am providing a test to prove the issue. I had to extend the group membership test by including two more entity types as the case is quite difficult to reproduce. It takes that two entities from different types share the same Id and point to different groups.

### DIFF
--- a/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
+++ b/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
@@ -175,6 +175,9 @@ class GroupMembershipManagerTest extends KernelTestBase {
   /**
    * Tests that the static cache loads the appropriate group.
    *
+   * Verify that entities from different entity types with colliding Ids that
+   * point to different groups do not confuse the membership manager.
+   *
    * @covers ::getGroupIds
    */
   public function testStaticCache() {
@@ -182,7 +185,7 @@ class GroupMembershipManagerTest extends KernelTestBase {
     $membership_manager = \Drupal::service('og.membership_manager');
     $bundle_rev = Unicode::strtolower($this->randomMachineName());
     $bundle_update = Unicode::strtolower($this->randomMachineName());
-    $field_settigns = [
+    $field_settings = [
       'field_name' => 'group_audience_node',
       'field_storage_config' => [
         'settings' => [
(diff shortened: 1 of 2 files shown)


idimopoulos:
I will provide a patch as soon as the tests complete.

tests/src/Kernel/Entity/GroupMembershipManagerTest.php
```diff
@@ -169,6 +173,57 @@ public function testGetGroupIds($group_type_id, $group_bundle, array $expected)
...
+   * Tests that the static cache loads the appropriate group.
+   *
+   * @covers ::getGroupIds
+   */
+  public function testStaticCache() {
+    /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+    $membership_manager = \Drupal::service('og.membership_manager');
+    $bundle_rev = Unicode::strtolower($this->randomMachineName());
+    $bundle_update = Unicode::strtolower($this->randomMachineName());
+    $field_settigns = [
```
pfrenssen: Small spellign error here.

pfrenssen:
This is an interesting bug which can lead to mysterious behaviour when a group has group content of multiple types with colliding IDs, or has group content which happens to have the same ID as a member of the group. I discussed this with @idimopoulos IRL and he told me this was discovered when 2 colliding IDs were written in the database after 4000 lines of Behat test, resulting in an inexplicable failure. I cannot imagine the amount of effort this took to debug.

Thanks a lot for the contribution!

pfrenssen:
This looks good, just waiting for the test results to come back before merging.

PR #281, opened 2017-06-23
Optimize some calls to og_is_group()

jorrit:
`og_is_group()` is faster when an entire entity is provided. I found two places where an entity is already available.

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -640,7 +640,7 @@ function _group_context_handler_entity($entity_type = 'node', $entity = NULL, $p
   }
 
   // Check if the entity is itself a group.
-  if ($group = og_is_group($entity_type, $id)) {
+  if (og_is_group($entity_type, $entity)) {
     $contexts[$entity_type][] = $id;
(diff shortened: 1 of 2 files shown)


amitaibu:
Needs a re-roll (tests on `7.x-1.x` are green again)

jorrit:
The patch seems to be valid for the current branch. Is there anything I can do so it can be merged?

og_context/og_context.module
```diff
@@ -634,13 +634,9 @@ function _group_context_handler_entity($entity_type = 'node', $entity = NULL, $p
     return;
   }
 
-  list($id) = entity_extract_ids($entity_type, $entity);
```
amitaibu: I don't think we can remove this. If we pass an unsaved entity, we need this validation.

jorrit:
Thanks again for your comment. I have updated the PR according to your review.

amitaibu:
Thanks.

jorrit:
Thank you!

PR #282, opened 2017-06-23
Check if  is set in og_is_group()

jorrit:
`entity_extract_ids()` throws an exception if it isn't.

### DIFF
--- a/og.module
+++ b/og.module
@@ -1816,12 +1816,16 @@ function og_membership_access($op, $entity, $account = NULL, $entity_type = 'og_
  *   The entity object, or the entity ID.
  *
  * @return
- *   TRUE or FALSE if the entity is a group.
+ *   TRUE if the entity exists and is a group, FALSE otherwise.
  */
 function og_is_group($entity_type, $entity) {
   if (is_numeric($entity)) {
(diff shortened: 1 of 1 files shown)


og.module
```diff
@@ -1816,12 +1816,16 @@ function og_membership_access($op, $entity, $account = NULL, $entity_type = 'og_
...
  *
  * @return
- *   TRUE or FALSE if the entity is a group.
+ *   TRUE if the entity exists and is a group, FALSE otherwise.
  */
 function og_is_group($entity_type, $entity) {
   if (is_numeric($entity)) {
     $entity = entity_load_single($entity_type, $entity);
   }
+  if (!$entity) {
```
amitaibu, line 1825: that is "baby sitting" I prefer not to add. The API function is expecting to get an object or an entity ID. If you didn't pass it, I prefer it to throw you an exception.

amitaibu:
Sorry - but closing, due to above comment. Thanks for the PR, though :)

jorrit:
I was hoping there could be some discussion before you closed it. There is a specific use case for this in combination with the Panelizer module. I will try to get it fixed through them then.

PR #283, opened 2017-06-27
Fixed assigning default overridden role to user

jorrit:
When roles are overridden for a group while the overridden roles are not yet created, `og_role_grant()` refuses to assign the default role.

By adding the `$skipcheck` parameter the role is inserted in the database and changed to the overridden role later by `og_roles_override()`. Adding this parameter was the easiest way of solving the chicken-and-egg problem because `og_roles_override()` is already capable of mapping the default role to the overridden one.

I added a test case to `OgGroupAndUngroup` even though `OgDefaultAccessFieldTestCase` also seemed a good candidate because `OgGroupAndUngroup` already contains a test method specifically for assigning default roles to the group creator.

### DIFF
--- a/og.module
+++ b/og.module
@@ -855,7 +855,7 @@ function og_entity_insert($entity, $entity_type) {
     $name = 'og_group_manager_default_rids_' . $entity_type . '_' . $bundle;
     if ($rids = variable_get($name)) {
       foreach ($rids as $rid) {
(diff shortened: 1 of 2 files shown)


jorrit:
Is there something I can improve to this pull request?

claudiu-cristea:

PR #285, opened 2017-07-06
OgRoleException thrown when trying to remove a group type

pfrenssen:
Currently when a group type is being removed an `OgRoleException` occurs:

> Drupal\og\Exception\OgRoleException: The default roles "non-member" and "member" cannot be deleted.

This happens because the default roles are also being deleted as part of the cleanup of the dependency chain. Before throwing this exception we should check that the deletion is happening as part of a dependency cleanup, i.e. when `::onDependencyRemoval()` has been called.

### DIFF
--- a/src/Entity/OgRole.php
+++ b/src/Entity/OgRole.php
@@ -45,6 +45,14 @@ class OgRole extends Role implements OgRoleInterface {
   protected $name;
 
   /**
+   * Whether or not the parent entity we depend on is being removed.
+   *
+   * @var bool
+   *   TRUE if the entity is being removed.
+   */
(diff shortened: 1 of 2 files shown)


pfrenssen:
The test fails because of the exception. Now pushing the fix.

pfrenssen:
Tests failed on 8.3.x because of #286.

pfrenssen:
I merged in #286 so this can become green. This is now blocked on #286.

pfrenssen:
#286 has been merged into 8.x-1.x, so this is now unblocked.

amitaibu:
Lovely, thanks.

PR #286, opened 2017-07-06
Test failures in 8.3.x caused by moving  entity_test_update to a separate module

pfrenssen:
In [Issue #2856808: Break out the 'entity_test_update' entity type into its own module and add additional test db dumps](https://www.drupal.org/node/2856808) the `entity_test_update` entity type was moved into a separate module, causing test failures in our 8.3.x tests.

Let's use a different test entity type in our test suite, one that is still compatible with all tested versions.

This makes me think that we shouldn't even be testing on 8.1.x and 8.2.x any more since they are deprecated. Not even core is testing those versions anymore. Instead we should be testing the current and the next version (now, 8.3.x and 8.4.x). But that is out of scope for this issue.

### DIFF
--- a/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
+++ b/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
@@ -3,8 +3,9 @@
 namespace Drupal\Tests\og\Kernel\Entity;
 
 use Drupal\Component\Utility\Unicode;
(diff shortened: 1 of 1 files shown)


pfrenssen:
I created a followup issue to test only on the current and next version of Drupal core: #287.

amitaibu:
Thanks!

PR #287, opened 2017-07-06
Test current versions of Drupal and PHP

pfrenssen:
There is little point in testing obsolete versions of Drupal 8, since not even core is doing this. Once OG is released and our users decide to stick with an older version of D8 for some business reason, they should pin their contrib modules in `composer.json` at the same time as they pin the core version to maintain compatibility.

I think it would be a good idea to always test the current version as well as the upcoming version (currently 8.3.x and 8.4.x).

There is also a new stable PHP release, let's test this too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ matrix:
   exclude:
     - php: 5.6
       env: TEST_SUITE=PHP_CodeSniffer
+    - php: 7.0
+      env: TEST_SUITE=PHP_CodeSniffer
 
 mysql:
   database: og
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,9 @@ before_script:
 
   # Download Drupal console so we can run the test for it. Skip this for the
   # coding standards test.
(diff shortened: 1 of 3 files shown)


pfrenssen:
There appear to be some problems :)

pfrenssen:
The failures occur because Drupal Console doesn't work with Drupal 8.4.x.

See https://github.com/hechoendrupal/drupal-console-core/issues/151 and https://github.com/hechoendrupal/drupal-console/issues/2987

pfrenssen:
Hmm I cannot work around this easily. I'd actually much rather remove the Drupal console command, it's not terribly useful; it is intended to attach the group audience field to a group content type which is equally easy to do in the Drupal UI. We have already wasted more time on maintaining the tests with Drupal console than with anything else.

We can maybe move the `OgAddField` console command to a separate project?

pfrenssen:
I was able to work around it after all, by simply creating a separate test group for the console command, and only testing this on 8.3.x.

amitaibu:
👍

PR #288, opened 2017-07-06
Provide a method to check if an OgRole is a default role

pfrenssen:
Default OgRole entities (`anonymous` and `member`) are different from the standard roles. For example it is not possible to delete them.

Because of these differences we should provide a method so that it can be checked whether a role is a default role.

### DIFF
--- a/src/Entity/OgRole.php
+++ b/src/Entity/OgRole.php
@@ -242,13 +242,20 @@ class OgRole extends Role implements OgRoleInterface {
     // The default roles are required. Prevent them from being deleted for as
     // long as the group still exists, unless the group itself is in the process
     // of being removed.
-    if (!$this->parentEntityIsBeingRemoved && in_array($this->getName(), [self::ANONYMOUS, self::AUTHENTICATED]) && $this->groupTypeManager()->isGroup($this->getGroupType(), $this->getGroupBundle())) {
+    if (!$this->parentEntityIsBeingRemoved && $this->isDefaultRole() && $this->groupTypeManager()->isGroup($this->getGroupType(), $this->getGroupBundle())) {
       throw new OgRoleException('The default roles "non-member" and "member" cannot be deleted.');
     }
 
     // Reset access cache, as the role is no longer present.
     $this->ogAccess()->reset();
     parent::delete();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isDefaultRole() {
+    return in_array($this->getName(), [self::ANONYMOUS, self::AUTHENTICATED]);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


pfrenssen:
This is blocked by #286 (fixes a test failure in the current HEAD) and #285 (to avoid a merge conflict).

pfrenssen:
Note: this has been split off from #244.

src/Entity/OgRole.php
```diff
@@ -243,6 +252,13 @@ public function delete() {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function isDefaultRole() {
+    return in_array($this->getName(), [self::ANONYMOUS, self::AUTHENTICATED]);
```
zerolab: `self::ADMINISTRATOR` is also a default role

zerolab: ah, but then this does not throw an exception.
Took some more reading the code to clear that up. How do we distinguish the default roles provided by `OgRoleInterface` from "default roles that will throw an exception when deleted"?

pfrenssen: Very good point. `isDefaultRole()` is not the correct name for this method.

Maybe a better name would be `isRequiredRole()`, then it explains more clearly what this returns, and why you would need to call this method.

pfrenssen: I changed it to `isRequired()` so it matches the other available methods such as `isNew()` and `isAdmin()`. I also updated the test coverage which was talking about 'default roles' where it should be 'required roles'.

pfrenssen:
#286 and #285 are merged so this is unblocked and ready for review!

amitaibu:
I think `isRequired` is still not the best choice, maybe `isNonDeletable`?

amitaibu:
After this change ,or if you think we should stick to `isRequired` -- MWG (Merge When Green 😉 )

pfrenssen:
It called it like this because the constant is called `OgRoleInterface::ROLE_TYPE_REQUIRED`. I do agree that it is not 100% clear because in most of the documentation it is additionally clarified that "the role cannot be deleted". On the other hand "required" implies also that it cannot be deleted I think.

I did a grep for the usage of this terminology. Apart from the constant there is only one other method `OgRoleManagerInterface::getRequiredDefaultRoles()`, so it would be not too complicated to change it.

Apart from this there are many mentions in the comments, mainly in `OgRoleTest` and some other tests.

I also found one instance where this new method should be used: `ChangeSingleOgMembershipRoleBase` can use this new method on line 128:

```
    // Do not return the default roles 'member' and 'non-member'. These are
    // required and cannot be added to or removed from a membership.
    $role_names = [];
    foreach ($roles as $role) {
-     if ($role->getRoleType() !== OgRoleInterface::ROLE_TYPE_REQUIRED) {
+     if (!$role->isRequired()) {
        $role_names[$role->getName()] = $role->label();
      }
    }
```

Actually in this code the role type is checked not only to see if the role can be deleted, but also if it can be added. In this context `$role->isRequired()` makes more sense than `$role->isNonDeletable()`.

amitaibu:
Thanks.

PR #289, opened 2017-07-06
Provide a helper method to load roles by group type

pfrenssen:

Let's provide a simple helper method that allows to load all OgRole entities that belong to a given group type.

### DIFF
--- a/src/Entity/OgRole.php
+++ b/src/Entity/OgRole.php
@@ -162,6 +162,17 @@ class OgRole extends Role implements OgRoleInterface {
   /**
    * {@inheritdoc}
    */
+  public static function loadRolesByGroupType($group_entity_type_id, $group_bundle_id) {
+    $properties = [
+      'group_type' => $group_entity_type_id,
+      'group_bundle' => $group_bundle_id,
+    ];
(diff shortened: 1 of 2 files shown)


pfrenssen:
This is blocked on #286 which fixes a failure that affects the current HEAD.

amitaibu:
What is the use case for this?

pfrenssen:
The use case for this in #244 is that it is providing a form to add new roles to a group type. In order to generate the form it needs to retrieve the roles that are associated with the group type:

https://github.com/ec-europa/og/blob/ISAICP-2709/src/Form/OgAddMultipleRolesForm.php#L27

I figured this could be useful for all cases where somebody needs to get the list of roles that exist for a certain group type, but this is just a minor DX addition so I can live without it.

I also see that the method name `OgRole::loadRolesByGroupType()` is not consistent with the other helper method (`OgRole::loadByGroupAndName()`). If we do decide to keep this then we can maybe change the name to `loadByGroupType()` or `loadByGroupEntityTypeAndBundle()`.

amitaibu:
Yeah, let's go ahead with the proposed renaming.

pfrenssen:
Renamed the method.

amitaibu:
Thanks.

PR #290, opened 2017-07-14
Wrong references for og_ui_admin_roles validate and submit functions.

Xilis:
Was looking for validate and submit handlers for og_ui_admin_roles, spotted the invalid function names.


amitaibu:
Thanks.

PR #291, opened 2017-07-20
Clean up orphaned memberships when deleting users

pfrenssen:
This is fixing a long standing @todo we had left: cleaning up orphaned memberships when a user is deleted.


amitaibu:
👍

Issue #295, opened 2017-08-03
og_entity_insert does not care if there is a memberhsip


idimopoulos:
I have the following issue:
Og, since it wants to behave agnostically, uses hook_entity_insert in order to create a membership for the user that creates the memberhsip (if the entity inherits the publisher interface).

The problem comes by the fact that this blocks users from tampering with the membership on creation. The only way that a user can interfere is that his module is named something that starts with later than 'o' so that it's listed after og, and use the hook_entity_insert in order to catch and alter the membership.

In my case, I want to use the hook_ENTITY_TYPE_insert, but since og does not check for a membership before creating one, an exception is thrown as there is an active membership.

Proposed solution:
Since og_entity_insert simply creates and saves a membership, escape early if there is already a membership found.

pfrenssen:
Fixed in #296.

Issue #292, opened 2017-07-25
Default to not filtering on active memberships in MembershipManagerInterface::getMembership()?


pfrenssen:
I had a few cases this week where I needed to load a membership but forgot that `MembershipManagerInterface::getMembership()` only returns active memberships by default. If you actually want to get a membership you need to do:

```
    $membership_manager = \Drupal::service('og.membership_manager');
    $states = [
      OgMembershipInterface::STATE_ACTIVE,
      OgMembershipInterface::STATE_PENDING,
      OgMembershipInterface::STATE_BLOCKED,
    ];
    $membership = $membership_manager->getMembership($group, $user, $states);
```

I feel this makes it less DX friendly. Can we maybe default to all states, or removing the `$states` parameter altogether? After all it's very easy to check if a membership is active:

```
$membership = $membership_manager->getMembership($group, $user);
if ($membership->isActive()) {
  // ...
}
```

amitaibu:
My concern here is that people would assume getMembership will return a "valid" membership. One could argue (as you do 😉 ) that `node_load()` also loads unpublished nodes.

So I think we can go with a mix of both:

`getMembership` -- returns all states
`getActiveMembership` -- to explicetly get an active one.

pfrenssen:
Now that is a great idea!

PR #296, opened 2017-08-03
Avoid exception in case a membership is already found.

idimopoulos:
From issue https://github.com/Gizra/og/issues/295

### DIFF
--- a/og.module
+++ b/og.module
@@ -47,8 +47,11 @@ function og_entity_insert(EntityInterface $entity) {
     return;
   }
 
-  $membership = Og::createMembership($entity, $entity->getOwner());
-  $membership->save();
+  // Other modules might interfere and create a membership.
+  if (!Og::getMembership($entity, $entity->getOwner())) {
+    $membership = Og::createMembership($entity, $entity->getOwner());
+    $membership->save();
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


og.module
```diff
@@ -47,8 +47,11 @@ function og_entity_insert(EntityInterface $entity) {
     return;
   }
 
-  $membership = Og::createMembership($entity, $entity->getOwner());
-  $membership->save();
+  // Other modules might interfere and create a membership.
+  if (!Og::getMembership($entity, $entity->getOwner())) {
```
amitaibu: I think here you will get only the `active` memberships.

og.module
```diff
@@ -47,8 +47,11 @@ function og_entity_insert(EntityInterface $entity) {
     return;
   }
 
-  $membership = Og::createMembership($entity, $entity->getOwner());
-  $membership->save();
+  // Other modules might interfere and create a membership.
```
amitaibu: maybe `// Implementing modules might have created a membership ahead of us.`

amitaibu (commented):
Thanks. I'd love to get a test as-well to confirm the functionality.

og.module
pfrenssen: Indeed!! I have also been trapped by this in the past, which is why I created #292 to lobby for changing this to return all memberships by default. I think it is not intuitive that this method returns filtered results by default, and an additional parameter should be passed to get the actual membership.

pfrenssen:
I think it's OK for me to merge this since I did not write the original patch but just the test coverage that proves the bug is real.

PR #300, opened 2017-08-24
Create og_perm views access plugin

geertvd:
Fixes issue #299

PR #301, opened 2017-08-24
Fix entity_reference field schema issues

geertvd:
Fixes failing tests in #300


geertvd:
This was caused by https://www.drupal.org/node/2787873

amitaibu:
Thanks!

PR #311, opened 2017-09-18
Throw an exception when calling `MembershipManager::getGroupIds()` with an entity which is not group content

pfrenssen:
While working on a test for #296 I noticed that `og_invalidate_group_content_cache_tags()` calls `MembershipManager::getGroupIds()` even when the entity is a group and not group content. This is useless, `::getGroupIds()` only works on group content and is clearly documented as such.

This results in a useless method being called and its useless results cached every time one of the entity hooks is firing on a group entity. Let's fix the bug and also throw an exception when the wrong entity is being passed, so that developers are alerted if they are using the method incorrectly.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -154,6 +154,11 @@ class MembershipManager implements MembershipManagerInterface {
       throw new \InvalidArgumentException('\Drupal\og\MembershipManager::getGroupIds() cannot be used for user entities. Use \Drupal\og\MembershipManager::getUserGroups() instead.');
     }
 
+    // This should only be called on group content types.
+    if (!$this->groupAudienceHelper->hasGroupAudienceField($entity->getEntityTypeId(), $entity->bundle())) {
+      throw new \InvalidArgumentException('Can only retrieve group IDs for group content entities.');
+    }
+
     $identifier = [
       __METHOD__,
       $entity->getEntityTypeId(),
--- a/src/OgAccess.php
+++ b/src/OgAccess.php
@@ -262,36 +262,37 @@ class OgAccess implements OgAccessInterface {
     }
 
     $is_group_content = $this->groupAudienceHelper->hasGroupAudienceField($entity_type_id, $bundle);
-    $cache_tags = $entity_type->getListCacheTags();
-
-    // The entity might be a user or a non-user entity.
-    $groups = $entity->getEntityTypeId() == 'user' ? $this->membershipManager->getUserGroups($entity) : $this->membershipManager->getGroups($entity);
-
-    if ($is_group_content && $groups) {
-      $forbidden = AccessResult::forbidden()->addCacheTags($cache_tags);
-      foreach ($groups as $entity_groups) {
-        foreach ($entity_groups as $group) {
-          // Check if the operation matches a group content entity operation
-          // such as 'create article content'.
-          $operation_access = $this->userAccessGroupContentEntityOperation($operation, $group, $entity, $user);
-
-          if ($operation_access->isAllowed()) {
-            return $operation_access->addCacheTags($cache_tags);
-          }
-
-          // Check if the operation matches a group level operation such as
-          // 'subscribe without approval'.
-          $user_access = $this->userAccess($group, $operation, $user);
-          if ($user_access->isAllowed()) {
-            return $user_access->addCacheTags($cache_tags);
-          }
-
-          $forbidden->inheritCacheability($user_access);
-        }
-      }
-      return $forbidden;
-    }
     if ($is_group_content) {
+      $cache_tags = $entity_type->getListCacheTags();
+
+      // The entity might be a user or a non-user entity.
+      $groups = $entity->getEntityTypeId() == 'user' ? $this->membershipManager->getUserGroups($entity) : $this->membershipManager->getGroups($entity);
+
+      if ($groups) {
+        $forbidden = AccessResult::forbidden()->addCacheTags($cache_tags);
+        foreach ($groups as $entity_groups) {
+          foreach ($entity_groups as $group) {
+            // Check if the operation matches a group content entity operation
+            // such as 'create article content'.
+            $operation_access = $this->userAccessGroupContentEntityOperation($operation, $group, $entity, $user);
+
+            if ($operation_access->isAllowed()) {
+              return $operation_access->addCacheTags($cache_tags);
+            }
+
+            // Check if the operation matches a group level operation such as
+            // 'subscribe without approval'.
+            $user_access = $this->userAccess($group, $operation, $user);
+            if ($user_access->isAllowed()) {
+              return $user_access->addCacheTags($cache_tags);
+            }
+
+            $forbidden->inheritCacheability($user_access);
+          }
+        }
+        return $forbidden;
+      }
+
       $result->addCacheTags($cache_tags);
     }
 
--- a/src/OgDeleteOrphansBase.php
+++ b/src/OgDeleteOrphansBase.php
@@ -40,6 +40,13 @@ abstract class OgDeleteOrphansBase extends PluginBase implements OgDeleteOrphans
   protected $membershipManager;
 
   /**
+   * The OG group audience helper.
+   *
+   * @var \Drupal\og\OgGroupAudienceHelperInterface
+   */
+  protected $groupAudienceHelper;
+
+  /**
    * Constructs an OgDeleteOrphansBase object.
    *
    * @param array $configuration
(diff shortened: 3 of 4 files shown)


pfrenssen:
The schema errors are possibly due to https://www.drupal.org/node/2870971.

og.module
```diff
@@ -289,10 +289,9 @@ function og_entity_type_alter(array &$entity_types) {
 function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
   // If group content is created or updated, invalidate the group content cache
   // tags for each of the groups this group content belongs to. This allows
-  // group listings to be cached effectively. Also invalidate the tags if the
-  // group itself changes. The cache tag format is
+  // group listings to be cached effectively. The cache tag format is
   // 'og-group-content:{group entity type}:{group entity id}'.
-  if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle()) || Og::isGroupContent($entity->getEntityTypeId(), $entity->bundle())) {
+  if (Og::isGroupContent($entity->getEntityTypeId(), $entity->bundle())) {
```
pfrenssen: This check to see whether the entity is a group is useless, in `MembershipManager::getGroupIds()` the queries are performed on group audience fields. If there are no group audience fields it won't return any results.

src/OgAccess.php
```diff
@@ -262,36 +262,37 @@ public function userAccessEntity($operation, EntityInterface $entity, AccountInt
     }
 
     $is_group_content = $this->groupAudienceHelper->hasGroupAudienceField($entity_type_id, $bundle);
-    $cache_tags = $entity_type->getListCacheTags();
-
-    // The entity might be a user or a non-user entity.
-    $groups = $entity->getEntityTypeId() == 'user' ? $this->membershipManager->getUserGroups($entity) : $this->membershipManager->getGroups($entity);
-
-    if ($is_group_content && $groups) {
```
pfrenssen, line 270: We already know at this point whether the entity is group content, so a small reshuffle of these two lines into the if statement is sufficient to not do an unneeded call to `MembershipManager::getGroups()`.

src/OgDeleteOrphansBase.php
```diff
@@ -135,10 +146,17 @@ protected function deleteOrphan($entity_type, $entity_id) {
       return;
     }
 
-    // Only delete content that is fully orphaned, i.e. it is no longer
+    // Only delete group content that is fully orphaned, i.e. it is no longer
     // associated with any groups.
-    $group_count = $this->membershipManager->getGroupCount($entity);
-    if ($group_count == 0) {
+    if ($this->groupAudienceHelper->hasGroupAudienceField($entity->getEntityTypeId(), $entity->bundle())) {
+      $group_count = $this->membershipManager->getGroupCount($entity);
```
pfrenssen, line 153: Only do a group count if the entity is actually group content.

config/schema/og.schema.yml
```diff
@@ -153,3 +153,7 @@ condition.plugin.og_group_type:
       type: sequence
       sequence:
         type: string
+
+entity_reference_selection.og:default:
+  type: entity_reference_selection.default
+  label: 'OG entity selection handler settings'
```
pfrenssen: Schema change required due to https://www.drupal.org/node/2870971 - I can split this off to a separate ticket if required.

og.module
```diff
@@ -317,8 +317,7 @@ function og_entity_type_alter(array &$entity_types) {
 function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
   // If group content is created or updated, invalidate the group content cache
   // tags for each of the groups this group content belongs to. This allows
-  // group listings to be cached effectively. Also invalidate the tags if the
-  // group itself changes. The cache tag format is
+  // group listings to be cached effectively. The cache tag format is
```
pfrenssen: This change should be undone, the code is now effectively checking if the group itself changes, this was added in #337.

config/schema/og.schema.yml
pfrenssen: Let's try leaving this out, this hasn't proven to be problematic in the last 10 months so it was maybe a temporary hiccup that is no longer needed.

pfrenssen:
Haha Github doesn't allow me to reject my own changes :grin:

pfrenssen:
The schema change can just be removed, it has already been fixed in #323 by @claudiu-cristea. It is now duplicating it. I will clean this up and then rebase to clean up the commit history.

pfrenssen:
OK this is looking good now. Some of the changes that were included here have been solved in other PRs in the meantime, so It is easier to read. It should be pretty easy to review now.

tatarbj:
For me it seems a bit over of the original scope, maybe renaming the PR would make sense. Apart of it as we've also discussed there are no other cases when deleteOrphan() will be executed so it seems safe to delete the entity when it's not a group content (else case).
It looks good to me, thanks @pfrenssen for the clear comments, helped me a lot of understand the cases, also the inline comments are clear and proper!

pfrenssen:
Do you mean this comment?

```
+    // If the entity is not group content (e.g. an OgMembership entity), just go
+    // ahead and delete it.
```

That is already what was happening before this PR, I just added some documentation here to explain what is going on.

I don't see where this is going out of scope? The PR adds the new exception that is being thrown if the `getGroupIds()` method is called with a wrong argument. There were two instances where this was called with the wrong argument in our own code, and these two are now fixed by first checking that the entity is group content before calling the method. And a test is added. There is nothing else in the PR?

src/MembershipManager.php
```diff
@@ -214,6 +214,11 @@ public function getGroupIds(EntityInterface $entity, $group_type_id = NULL, $gro
       throw new \InvalidArgumentException('\Drupal\og\MembershipManager::getGroupIds() cannot be used for user entities. Use \Drupal\og\MembershipManager::getUserGroups() instead.');
     }
 
+    // This should only be called on group content types.
+    if (!$this->groupAudienceHelper->hasGroupAudienceField($entity->getEntityTypeId(), $entity->bundle())) {
+      throw new \InvalidArgumentException('Can only retrieve group IDs for group content entities.');
```
MPParsley, line 219: Can config entities belong to a group? If so I'd go for:

```suggestion
      throw new \InvalidArgumentException('Can only retrieve group IDs for group entities.');
```

A bit further (and most of the time we just refer to **group content**):

>     // Only delete group content that is fully orphaned, i.e. it is no longer
```suggestion
      throw new \InvalidArgumentException('Can only retrieve group IDs for group content.');
```

Remains this question: a user doesn't have an og_audience field but don't we consider it "group content"?

src/OgAccess.php
```diff
@@ -262,36 +262,37 @@ public function userAccessEntity($operation, EntityInterface $entity, AccountInt
...
-          // Check if the operation matches a group level operation such as
-          // 'subscribe without approval'.
-          $user_access = $this->userAccess($group, $operation, $user);
-          if ($user_access->isAllowed()) {
-            return $user_access->addCacheTags($cache_tags);
+    if ($is_group_content) {
+      $cache_tags = $entity_type->getListCacheTags();
+
+      // The entity might be a user or a non-user entity.
+      $groups = $entity->getEntityTypeId() == 'user' ? $this->membershipManager->getUserGroups($entity) : $this->membershipManager->getGroups($entity);
```
MPParsley, line 269: Or we could test for Drupal\user\UserInterface?

Also use === of you want strict typing

Also might have been useful to have this check within getGroups().

src/OgDeleteOrphansBase.php
MPParsley:
```suggestion
      // Only do a group count if the entity is actually group content.
      $group_count = $this->membershipManager->getGroupCount($entity);
```

MPParsley (approved):

src/MembershipManager.php
pfrenssen: Since the group is referenced on a entity reference field, all group content needs to be a content entity (since config entities are not fieldable). Also the name "group content" implies that it is a content entity :grin: 

Users are not considered to be group content, they are considered to be group members, and are managed not through a reference field, but through the `OgMembership` entity tables.

src/OgAccess.php
pfrenssen: I agree with this proposal but these lines of code are not in fact being changed here, this PR just introduces this single line and moves the related code in it:

```
if ($is_group_content) {
  // All code in here is in fact unchanged, but shows up in Github as a huge chunk of changed code.
}
```

I will create a separate PR to implement this suggestion, so this can stay on scope.

src/MembershipManager.php
MPParsley: Note that webform is a bundlable config entity and has plans to make it fieldable (not that I would suggest doing this).
See https://www.drupal.org/project/webform/issues/3049023#comment-13073385

PR #312, opened 2017-09-18
`GroupMembershipManagerTest::testStaticCache()` doesn't have the right `@covers` annotation

pfrenssen:
Very small fix, while reading through the `GroupMembershipManagerTest` I noticed that `::testStaticCache()` has an annotation declaring that it is testing the `::getGroupIds()` method, but actually it is calling `::getGroups()` in the test.


amitaibu:
MWG! (`Merge When Green`, in Gizra lingo 😉 )

pfrenssen:
Nice that the RFR label is also green :D

Issue #313, opened 2017-09-18
BundleFormAlter crashes on bundle forms whose content entities implement bundleFieldDefinitions()


joachim-n:
I am getting this crash with Commerce's form to add an Order Item Type:

```
Drupal\commerce_order\Entity\OrderItem::bundleFieldDefinitions(Object, NULL, Array) (Line: 331)
Drupal\Core\Entity\EntityFieldManager->buildBundleFieldDefinitions('commerce_order_item', NULL, Array) (Line: 298)
Drupal\Core\Entity\EntityFieldManager->getFieldDefinitions('commerce_order_item', NULL) (Line: 68)
Drupal\og\OgGroupAudienceHelper->getAllGroupAudienceFields('commerce_order_item', NULL) (Line: 46)
Drupal\og\OgGroupAudienceHelper->hasGroupAudienceField('commerce_order_item', NULL) (Line: 269)
Drupal\og\Og::isGroupContent('commerce_order_item', NULL) (Line: 152)
Drupal\og_ui\BundleFormAlter->addGroupContent(Array, Object) (Line: 73)
Drupal\og_ui\BundleFormAlter->formAlter(Array, Object) (Line: 24)
og_ui_form_alter(Array, Object, 'commerce_order_item_type_add_form') (Line: 501)
Drupal\Core\Extension\ModuleHandler->alter('form', Array, Object, 'commerce_order_item_type_add_form') (Line: 819)
Drupal\Core\Form\FormBuilder->prepareForm('commerce_order_item_type_add_form', Array, Object) (Line: 272)
```

This seems to be because there is this call:

```
      '#default_value' => Og::isGroupContent($this->entityTypeId, $this->bundle),
```

which doesn't check to see whether $this->bundle is NULL. On a form to create a new bundle entity, there's no entity ID yet, and thus so is $this->bundle which is the bundle being edited and set from the entity ID.

PR #242, opened 2017-03-06
[WIP] OG Access port

zerolab:
This is the start for the OG Access module port.

Took a simple approach to group content visibility. It mirrors that of the group visibility (public, private). If any of the groups the content is assigned is private, then we mark it as private.

Any feedback at this stage would be most welcome.

Tasks:
- [x] Grants and node access
- [ ] Create batch to process changes to group visibility
- [ ] Tests

PR #314, opened 2017-09-18
Fixed crash on bundle entity add forms

joachim-n:
See https://github.com/Gizra/og/issues/313


joachim-n:
Argh, doesn't fully fix the problem.

Will work on this some more in a few days.

joachim-n:
OK, it *does* fix the problem.

The other crash was due to code in https://github.com/Gizra/og/pull/242.

pfrenssen:
This looks good. I'm OK with this not having a dedicated test btw, this is a trivial fix.

Thanks!

PR #323, opened 2017-10-15
Schema for OG audience handler settings is wrong

claudiu-cristea:
I'm getting this error while running a kernel test:

```
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors
for field.field.node.article.og_audience with the following errors:
field.field.node.article.og_audience:settings.handler_settings.target_bundles
missing schema
```

and, yes, the schema is missing for that part because, right now, the schema is:
```yaml
    handler_settings:
      type: entity_reference_selection.[%parent.handler]
```
which resolves to `entity_reference_selection.og:default`. This value is not falling back to the generic `entity_reference_selection.default:*` but to `entity_reference_selection.*`. And the last one doesn't define any schema for `handler_settings`.

Solution: Add a schema for `entity_reference_selection.og:default` that simply extends `entity_reference_selection.default`.

### DIFF
--- a/config/schema/og.schema.yml
+++ b/config/schema/og.schema.yml
@@ -42,10 +42,10 @@ field.field_settings.og_standard_reference:
       type: boolean
       label: 'Access Override'
 
-#entity_reference_selection.og:default:
-#  type: entity_reference_selection.default
-#  label: 'The OG selection handler settings'
-#
+entity_reference_selection.og:default:
+  type: entity_reference_selection.default
+  label: 'The OG selection handler settings'
+
 og.settings:
   type: config_object
   label: 'Organic Groups settings'
--- a/config/schema/og.schema.yml
+++ b/config/schema/og.schema.yml
@@ -42,6 +42,10 @@ field.field_settings.og_standard_reference:
       type: boolean
       label: 'Access Override'
 
+#entity_reference_selection.og:default:
+#  type: entity_reference_selection.default
+#  label: 'The OG selection handler settings'
+#
 og.settings:
   type: config_object
   label: 'Organic Groups settings'
--- a/tests/src/Kernel/SelectionHandlerSettingsSchemaTest.php
+++ b/tests/src/Kernel/SelectionHandlerSettingsSchemaTest.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Drupal\Tests\og\Kernel;
+
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\node\Entity\NodeType;
+
+/**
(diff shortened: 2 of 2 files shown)


amitaibu:
Thanks :)

Just to confirm, and see Travis catches those errors. If you'd comment out your code in the PR, will Travis return red?

claudiu-cristea:
@amitaibu I really don't understand how this is passing. Locally this test fails. Have no idea.

scripts/travis-ci/run-test.sh
```diff
@@ -14,12 +14,12 @@ case "$1" in
...
         ln -s $MODULE_DIR $DRUPAL_DIR/modules/og
         cd $DRUPAL_DIR
-        ./vendor/bin/phpunit -c ./core/phpunit.xml.dist $MODULE_DIR/tests
+        ./vendor/bin/phpunit -c ./core/phpunit.xml.dist $MODULE_DIR/
         exit $?
         ;;
     *)
         ln -s $MODULE_DIR $DRUPAL_DIR/modules/og
         cd $DRUPAL_DIR
-        ./vendor/bin/phpunit -c ./core/phpunit.xml.dist --exclude-group=console $MODULE_DIR/tests
```
claudiu-cristea: @amitaibu, It seems that OG UI tests never ran

claudiu-cristea: Dammit! The existing test that never ran is failing :)

claudiu-cristea:
I don't know why so many tests are running? You can see the failure in:

* PHP 7.1 & Drupal 8.4.x
* PHP 7 & Drupal 8.4.x

But I cannot explain the other failures. Those tests should not run.

scripts/travis-ci/run-test.sh
amitaibu: Oh my, it seems we've opened a pandora case! 😜

claudiu-cristea: @amitaibu so my patch to `scripts/travis-ci/run-test.sh` runs also other tests. Something is not OK.

claudiu-cristea:
@amitaibu I consider the OG UI tests not running out of scope, so I've opened #324 to handle that issue. Let's not mix them.

claudiu-cristea:
Now we can see the failure with the last commit. I wonder why it fails only with 8.4.x (but not with 8.3.x).

claudiu-cristea:
So commit e7ae52d reveals the bug and ad01008 is fixing the bug. I'm done here unless some changes are required.

amitaibu:
Merged, thanks!

Issue #331, opened 2017-10-26
add ways to set which Membership Type is used when creating a membership


joachim-n:
The various API functions for creating group memberships have a parameter for setting the membership type that is to be used for the new membership:

eg in MembershipManagerInterface:

```
  public function createMembership(EntityInterface $group, AccountInterface $user, $membership_type = OgMembershipInterface::TYPE_DEFAULT);
```

The route for subscribing to a group can take this too:

```
og.subscribe:
  path: 'group/{entity_type_id}/{group}/subscribe/{membership_type}'
  defaults:
    _controller: '\Drupal\og\Controller\SubscriptionController::subscribe'
    _title: 'Join Group'
    membership_type: default
```

However, all the code that calls this API or uses this route lets the default value of 'default' be used. Eg in GroupSubscribeFormatter:

```
        $parameters = [
          'entity_type_id' => $group->getEntityTypeId(),
          'group' => $group->id(),
        ];

        $url = Url::fromRoute('og.subscribe', $parameters);
``` 

There does not seem to be any way to set which type to use, either in the UI, or in code, short of writing a custom field formatter.

There are several ways this could be addressed.

Firstly, I think that group types should define a membership type to use by default, in a setting in the bundle form (BundleFormAlter). That ensures consistency in all code that creates memberships, which would need to look this up rather than just have a default value in a method parameter.

We could borrow the pattern from Commerce of the Order Type Resolver, which does a similar job of figuring out which bundle to use for a new entity, but that might be overkill here.  We do though need one of:

- MembershipManager::createMembership() will consult the group type and create the right type of membership, unless the parameter is specified to force it. (My preferred option, less work for the developer)
- an API to look up the membership type that's set by a group type, so you can get the right parameter to pass to MembershipManager::createMembership() 

Second, we could consider adding an option to the GroupSubscribeFormatter's settings to set the membership type that the formatter specifies in the URL it creates. That should have an option for 'use group type setting', which falls back to the setting for the group type. I'm not sure if there's a use case for that though, so I'd suggest leaving that to a follow-up.

joachim-n:
Working on a PR for this.

Rough outline:

- adding a form element to BundleFormAlter
- saving the membership type into a new property of og.settings, as: "$entity_type_id:$bundle_id" => membership type

WIP pushed to a branch -- https://github.com/Gizra/og/compare/8.x-1.x...joachim-n:sandbox-331?expand=1

joachim-n:
[...]

PR #333, opened 2017-10-27
331 set membership type for group type

joachim-n:
PR for #331.

### DIFF
--- a/src/GroupTypeManagerInterface.php
+++ b/src/GroupTypeManagerInterface.php
@@ -205,5 +205,5 @@ interface GroupTypeManagerInterface {
    *   The group map.
    */
   public function getGroupMap();
-  
+
 }
--- a/src/GroupTypeManagerInterface.php
+++ b/src/GroupTypeManagerInterface.php
@@ -2,9 +2,6 @@
 
 namespace Drupal\og;
 
-use Drupal\Core\Entity\EntityInterface;
-use Drupal\Core\Session\AccountInterface;
-
 /**
  * Group type manager interface.
  */
--- a/src/GroupTypeManagerInterface.php
+++ b/src/GroupTypeManagerInterface.php
@@ -1,0 +1,212 @@
+<?php
+
+namespace Drupal\og;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Group type manager interface.
+ */
+interface GroupTypeManagerInterface {
+
+  /**
+   * Determines whether an entity type ID and bundle ID are group enabled.
+   *
+   * @param string $entity_type_id
+   *   The entity type name.
+   * @param string $bundle
+   *   The bundle name.
+   *
+   * @return bool
+   *   TRUE if a bundle is a group.
+   */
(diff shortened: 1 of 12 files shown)


amitaibu:
Thanks! Can you provide gif/ screenshot(s) as-well - so it's easier to get the context.

joachim-n:
This shows the extra form element on the bundle form:

og - pr 333 screen shot

It uses states so it's only shown when the 'Group' checkbox is selected.

Just to be clear, I've not implemented the 2nd thing I suggested on #331, which was to add a setting to the GroupSubscribeFormatter. I'm not sure there's a use case for it now I think about it, as since the group settings are per-bundle, there is a 1-1 correspondence between the group bundle settings and the formatter's settings. Therefore having a setting in the formatter doesn't provide any finer-grained control. So overall, it's best to have this setting somewhere where the API uses it, rather than in the UI.

joachim-n:
Fixed the unit test errors that the CI was reporting.

Locally I am getting 1 fail with:

> 1) Drupal\Tests\og\Kernel\Console\DrupalConsoleAddFieldTest::testNewGroup
Error: Class 'Drupal\Console\Core\Style\DrupalStyle' not found

but I expect that's because my local copy of Drupal Console is out of date.

joachim-n:
Fixed the things the codesniffer was complaining about.

zerolab:
This works quite well on our build.

A couple of remarks:
1. it would be great to limit membership types per group type.
2. Had to have a "glue" patch for this + #231 + #181 - https://gist.github.com/zerolab/da4899abf6683180e1f07fc859b06707

joachim-n:
> it would be great to limit membership types per group type.

Not sure what you mean.

Do you mean that when you create a membership type, eg Membership Foo, you get to say in its form that it's only available for Group bundles X and Y?

That would presumably have to be enforced at the API level, since you can use the API to create a membership in any type, not just the one set for the group bundle.

zerolab:
> Do you mean that when you create a membership type, eg Membership Foo, you get to say in its form that it's only available for Group bundles X and Y?

Yes. If I have Group Type A and Group Type B and I need a special Membership Type for just B, I don't want to add a new member to a group in A and see the special membership type that is meant just for B.

joachim-n:
It's not my call, but I'd suggest that's left for a follow-up, as it's a fair bit more work.

This would need:

- a new property on the Group Membership type entity for the group bundles that it allows
- that *possibly* means adding a config dependency to the GM type, as it is referencing a bundle entity
- limiting the options shown on the group type form to the eligible GM types, and probably adding an API method to get those
- enforcing this setting in the API methods that create a group membership, so you can't pass in a GM type ID that's not allowed

staminna:
Drupal 7 backport or module?

src/GroupTypeManagerInterface.php
```diff
@@ -0,0 +1,209 @@
+<?php
+
+namespace Drupal\og;
```
MPParsley, line 3:
```suggestion
declare(strict_types = 1);

namespace Drupal\og;
```

claudiu-cristea:

PR #337, opened 2017-10-31
On audience change, cache is not properly invalidated

claudiu-cristea:
If the entity audience is changing the old group content OG tag is not invalidated. For example, if the old group is a node with nid 123 and the new group is nid 789 then the tag `og-group-content:node:123` is not invalidated. The test reveals the bug.

Additionally, the cache tags `og-group-content:{group entity type}:{group entity id}` are not correctly built because the code is making the assumption that a custom entity is using the core `\Drupal\Core\Entity\Entity::getCacheTagsToInvalidate()` method. But a custom entity may want to extend this method. For example in our project we hash the entity ID because the ID is a very long string that might break the `{cachetags}` table.

### DIFF
--- a/og.module
+++ b/og.module
@@ -340,7 +340,7 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
         }
       }
       if ($audience_has_changed) {
-        // If at least group has changed, we clear tha static cache of the
+        // If at least a group has changed, we clear tha static cache of the
         // membership manager which stores such relations. This is important
         // when the membership manager was used previously in the same request.
         $membership_manager->reset();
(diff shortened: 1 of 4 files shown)


og.module
```diff
@@ -312,13 +312,48 @@ function og_invalidate_group_content_cache_tags(EntityInterface $entity) {
...
     $tags = [];
-    foreach (\Drupal::service('og.membership_manager')->getGroupIds($entity) as $entity_type => $ids) {
-      foreach ($ids as $id) {
-        $tags[] = "$entity_type:$id";
+
+    // If the entity is a group content and we came here as an effect of an
+    // update, check if any of the OG audience fields has been changed (AKA 'the
+    // group of the entity has changed') and add also the tags of the old group.
+    /** @var \Drupal\Core\Entity\FieldableEntityInterface $entity */
+    if ($is_group_content && $original = $entity->original) {
```
pfrenssen: The `&&` operator has higher precedence than the `=` operator but this still works (probably because the operation doesn't make sense otherwise). Let's move this out of the `if` statement to save others some confusion while reading this.

pfrenssen:
Looks good, thanks!

PR #340, opened 2017-11-08
[D7] Check group existence before returning context

jorrit:
This prevents the following error when deleting the current group:
Notice: Trying to get property of non-object in entity_metadata_no_hook_node_access()

Also see https://www.drupal.org/node/2764017

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -415,6 +415,9 @@ function og_context_determine_context($group_type, $item = NULL, $check_access =
     $item = menu_get_item();
   }
 
+  $session_gid = !empty($_SESSION['og_context']['gid']) ? $_SESSION['og_context']['gid'] : NULL;
+  $session_type = !empty($_SESSION['og_context']['group_type']) ? $_SESSION['og_context']['group_type'] : NULL;
+
   $providers = og_context_negotiation_info();
   foreach ($enabled_providers as $name => $ignore) {
     if (empty($providers[$name])) {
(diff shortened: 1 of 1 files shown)


og_context/og_context.module
```diff
@@ -415,6 +415,9 @@ function og_context_determine_context($group_type, $item = NULL, $check_access =
     $item = menu_get_item();
   }
 
+  $sessiongid = !empty($_SESSION['og_context']['gid']) ? intval($_SESSION['og_context']['gid']) : NULL;
```
amitaibu: `sessiongid` => `session_gid`

og_context/og_context.module
```diff
@@ -436,26 +439,26 @@ function og_context_determine_context($group_type, $item = NULL, $check_access =
...
+      // and if so give it priority over the other IDs.
       $gids = $contexts[$group_type];
-      if (!empty($_SESSION['og_context']['group_type']) && $_SESSION['og_context']['group_type'] == $group_type && in_array($_SESSION['og_context']['gid'], $gids)) {
-        $check_gid = $_SESSION['og_context']['gid'];
-      }
-      else {
-        // Grab the first group ID.
-        $check_gid = reset($gids);
+      if ($sessiontype === $group_type && in_array($sessiongid, $gids) && $gids[0] != $sessiongid) {
+        array_unshift($gids, $_SESSION['og_context']['gid']);
```
amitaibu: Needs a comment what we do here.

amitaibu (changes requested):
Thanks.  I've added a few comments, but I'm not sure we have to take this path. I think that if after we load the `$group` it doesn't exist, we can just return early.

It sounds like an age case of not having the group in $_SESSION available, no?

Also, if in the end we do choose to take this path, this would need a simpleTest.

og_context/og_context.module
jorrit: Will do

og_context/og_context.module
jorrit: The comment explaining this is actually present but not directly above this code. I will move it around a bit.

jorrit:
Thanks for your quick response, @amitaibu, I really appreciate that!

I have updated this pull request according to your comments and I understand your comment that this is a bit over engineered. I also contemplated just doing this:

```diff
- if (!$check_access || entity_access('view', $group_type, $group)) {
+ if ($group && (!$check_access || entity_access('view', $group_type, $group))) {
```

but the reason I made a more comprehensive change is that I think the remaining values in `$gids` should also be taken into account.

The problem described is not just related to gids in `$_SESSION`.
When `og_context_handler_node()` is configured with sufficient priority, a path like `node/123/delete` (where 123 is a group node) causes `og_context_handler_node()` to return `[123]`, even though it may have just been deleted in that same request.

jorrit:
I will add the necessary tests in case you like to commit these changes.

claudiu-cristea:

PR #341, opened 2017-11-09
Infinite loop with og_context()

jorrit:
(copy from https://www.drupal.org/node/2717489)

There have been cases in the past where og_context() can cause an infinite loop depending on what submodules are doing. In the latest OG 2.9 I ran into a very specific case:

1) Enable the RealName module
2) Enable the Variables module and create a hook_variable_info() that needs to set the options of a variable based on og_vocab (so calls og_vocab_get_accessible_vocabs()).

What happens to an anonymous user is:
1) og_context() gets called by some hook_init() function
2) og_context calls user_load at the beginning to get the current user
3) user_load needs to load tokens
4) RealName module adds a token
5) Variables module adds tokens, so calls all hook_variable_info() hooks
6) A site specific module implements hook_variable_info() and calls og_vocab_get_accessible_vocabs() to get og_vocab data
7) og_vocab_get_accessible_vocabs() calls og_context()
--> goto 1, infinite loop.

Now, granted this is a bit obscure. But you can see other instances where this can happen. Doing any sort of user_load or node_load during og_context processing (such as in an og_context plugin) is just asking for complex situations like this.

It looks like the og_context_is_init() might have been intended to prevent infinite loops, but it is not used anywhere I can see.

So here is a simple patch to once-and-for-all prevent og_context() from getting into an infinite loop.

### DIFF
--- a/og_context/og_context.module
+++ b/og_context/og_context.module
@@ -217,77 +217,97 @@ function og_context_is_init($called = FALSE) {
  *   was found.
  */
 function og_context($group_type = 'node', $group = NULL, $account = NULL, $check_access = TRUE) {
-  global $user;
-  // User account is sent.
-  $account = $account ? $account : user_load($user->uid);
+  static $in_og_context = FALSE;
 
-  $id = FALSE;
-  if ($group) {
-    list($id) = entity_extract_ids($group_type, $group);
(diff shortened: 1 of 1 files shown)


claudiu-cristea:

PR #343, opened 2017-11-15
Delay OgRole storage loading in the Role manager

### LINKED ISSUE #342: Circular reference with content_moderation in 8.4.x when overriding 'content_moderation.state_transition_validation' to make it OG-aware
On our project we are swapping the `content_moderation.state_transition_validation` service with our own OG-aware implementation. To make it aware, we need to pass a reference to the `og.access` service, which in turn get GroupTypeManager which gets the OG Role manager in the constructor:

```
Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException: Circular reference detected for service "content_moderation.state_transition_validation", path: "content_moderation.state_transition_validation [error]
-> og.access -> og.group_type_manager -> og.role_manager". in core/lib/Drupal/Component/DependencyInjection/Container.php:141
```

According to https://www.drupal.org/node/2896536, this is because `OgRoleManager` calls the role storage too early. 

`GroupTypeManager` also loads the role storage, leading to the circular reference. And does nothing with it.

zerolab:
Fixes #342

### DIFF
--- a/src/GroupTypeManager.php
+++ b/src/GroupTypeManager.php
@@ -43,13 +43,6 @@ class GroupTypeManager {
   protected $configFactory;
 
   /**
-   * The entity storage for OgRole entities.
-   *
-   * @var \Drupal\Core\Entity\EntityStorageInterface
-   */
-  protected $ogRoleStorage;
-
-  /**
    * The service providing information about bundles.
    *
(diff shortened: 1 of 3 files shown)


zerolab:
This is ready, I believe.

pfrenssen:
Looks good, thanks!

PR #344, opened 2017-11-21
Skip creation of action plugins while in sync mode (aka config-import)

webflo:
Config import fails if og creates addional config on import.

```
The import failed due to the following reasons:   

  Deleted and replaced configuration entity "system.action.og_membership_add_  
  single_role_action.administrator"                                            
  Deleted and replaced configuration entity "system.action.og_membership_remo  
  ve_single_role_action.administrator"           

```

### DIFF
--- a/og.module
+++ b/og.module
@@ -8,6 +8,7 @@
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Config\Entity\ConfigEntityBundleBase;
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
@@ -370,7 +371,7 @@ function og_og_role_insert(OgRoleInterface $role) {
     return;
   }
 
-  if ($role instanceof \Drupal\Core\Config\Entity\ConfigEntityInterface) {
+  if ($role instanceof ConfigEntityInterface) {
     if ($role->isSyncing()) {
       return;
     }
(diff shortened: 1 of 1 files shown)


webflo:
``user_user_role_insert`` has a similar check for the same reason.

og.module
```diff
@@ -370,6 +370,12 @@ function og_og_role_insert(OgRoleInterface $role) {
     return;
   }
 
+  if ($role instanceof \Drupal\Core\Config\Entity\ConfigEntityInterface) {
```
zerolab: `OgRole` extends `Role` which, in turn, extends `ConfigEntityInterface`. This is in `og_og_role_insert()` which is `OgRole` specific, so this check is superfluous.

webflo: OgRoles extends Role. Thats corrent. But the type-hint of the function is OgRoleInterface. Should we change it to OgRole because the entity hook is related to the entity type and its base class.

zerolab: There is no need for it. `og_og_role_insert()` only fires when new `OgRole`s are added. So we just need to check that this is not during config sync.

Happy to stand corrected

og.module
```diff
@@ -370,6 +371,12 @@ function og_og_role_insert(OgRoleInterface $role) {
     return;
   }
 
+  if ($role instanceof ConfigEntityInterface) {
```
```suggestion
  // Skip creation of action plugins while importing configuration.
  if ($role instanceof ConfigEntityInterface) {
```

og.module
```diff
@@ -370,6 +371,13 @@ function og_og_role_insert(OgRoleInterface $role) {
     return;
   }
 
+  // Skip creation of action plugins while importing configuration.
+  if ($role instanceof ConfigEntityInterface) {
```
MPParsley: Can't we use this:

```suggestion
  if (\Drupal::isConfigSyncing()) {
```

og.module
MPParsley: How about using `\Drupal::isConfigSyncing()`?

MPParsley:
This was fixed in #443.

PR #346, opened 2017-11-24
Fixed fatal error when fetching group context when the route object has not been set.

### LINKED ISSUE #345: Call to a member function getPath() on null when group context is requested and no route has been set
I bumping into this issue when writing some tests.
I'm trying to fetch the group context quite early on in the request and `$this->routeMatch->getRouteObject()` is returning `NULL` which results in the following error:

```
Error: Call to a member function getPath() on null
Drupal\og\OgRouteGroupResolverBase->getContentEntity()() (Line: 92)
```

geertvd:
Fixed #345

### DIFF
--- a/src/OgRouteGroupResolverBase.php
+++ b/src/OgRouteGroupResolverBase.php
@@ -88,8 +88,12 @@ abstract class OgRouteGroupResolverBase extends OgGroupResolverBase implements C
    *   The entity, or NULL if we are not on a content entity path.
    */
   protected function getContentEntity() {
+    $route = $this->routeMatch->getRouteObject();
+    if (!$route) {
+      return NULL;
+    }
     // Check if we are on a content entity path.
-    $path = $this->routeMatch->getRouteObject()->getPath();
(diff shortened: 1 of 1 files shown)


src/OgRouteGroupResolverBase.php
```diff
@@ -88,6 +88,10 @@ public static function create(ContainerInterface $container, array $configuratio
    *   The entity, or NULL if we are not on a content entity path.
    */
   protected function getContentEntity() {
+    $route = $this->routeMatch->getRouteObject();
+    if (!$route) {
+      return NULL;
+    }
     // Check if we are on a content entity path.
     $path = $this->routeMatch->getRouteObject()->getPath();
```
zerolab: should this not become `$route->getPath()` ?

geertvd: Yes, fixed that

zerolab:
LGTM

amitaibu:
Thanks guys

PR #349, opened 2017-12-06
Return AccessResult::neutral() by default

webflo:
I am currently in the process of updating from an older version of OG. I think `::forbidden` as default value is too strict. Because it takes precedence over all the return values from other ``hook_entity_access`` implementations.

### DIFF
--- a/og.module
+++ b/og.module
@@ -166,7 +166,7 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface 
     return AccessResult::forbiddenIf($node_access_strict);
   }
 
-  return AccessResult::forbidden();
+  return AccessResult::neutral();
 }
 
 /**


og.module
```diff
@@ -166,7 +166,7 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
     return AccessResult::forbiddenIf($node_access_strict);
   }
 
-  return AccessResult::forbidden();
+  return AccessResult::neutral();
```
amitaibu, line 169: Maybe we should adapt the above `$node_access_strict = \Drupal::config('og.settings')->get('node_access_strict');` to be a general `$entity_access_strict = \Drupal::config('og.settings')->get('entity_access_strict');`

pfrenssen:
I think this is probably no longer needed. We now have a number of early checks that verify that the content is actually being handled by OG and these are returning `AccessResult::neutral()`. See for example  #374.

I'm closing this. If it is still a problem under certain circumstances we can revisit this.

PR #357, opened 2018-01-29
Fix fatal error when retrieving orphaned / non-existing roles from a membership

### LINKED ISSUE #356: Call to a member function bundle() on null
I am seeing the following error occurring in one of my Behat tests (ref https://continuousphp.com/git-hub/ec-europa/joinup-dev/build/a97410b0-0aba-4e3e-b414-4b9c7605428e/output):

```
Error: Call to a member function bundle() on null in web/modules/contrib/og/src/Entity/OgMembership.php on line 224 #0 web/modules/contrib/og/src/Cache/Context/OgRoleCacheContext.php(93): Drupal\og\Entity\OgMembership->getRoles() 
#1 web/core/lib/Drupal/Core/Cache/Context/CacheContextsManager.php(118): Drupal\og\Cache\Context\OgRoleCacheContext->getContext(NULL) 
```

In my test suite this is probably happening because a membership has been deleted in a previous test step but a reference to the membership is still present in the render cache or in a static cache.

The bug seems quite clear. `OgMembership::getRoles()` contains a call `$this->getGroup()->bundle()` but this doesn't account for the fact that `::getGroup()` might return `NULL` if there is no group set on the membership yet. This might for example happen if the membership has been newly instantiated and has no group yet. It should still be possible to call `::getRoles()` on it, and in this case an empty array should be returned.

This also means that in the calling code `OgRoleCacheContext::getContext()` we should take into account that an empty role list might be returned.

pfrenssen:
Fixes #356.

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -219,10 +219,14 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
    * {@inheritdoc}
    */
   public function getRoles() {
+    $roles = [];
+
     // Add the member role.
-    $roles = [
-      OgRole::getRole($this->getGroupEntityType(), $this->getGroup()->bundle(), OgRoleInterface::AUTHENTICATED),
-    ];
+    if ($group = $this->getGroup()) {
+      $roles = [
+        OgRole::getRole($this->getGroupEntityType(), $group->bundle(), OgRoleInterface::AUTHENTICATED),
+      ];
+    }
     $roles = array_merge($roles, $this->get('roles')->referencedEntities());
     return $roles;
   }
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -221,7 +221,8 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   public function getRoles() {
     $roles = [];
 
-    // Add the member role.
+    // Add the member role. This is only possible if a group has been set on the
+    // membership.
     if ($group = $this->getGroup()) {
       $roles = [
         OgRole::getRole($this->getGroupEntityType(), $group->bundle(), OgRoleInterface::AUTHENTICATED),
--- a/tests/src/Kernel/Entity/OgMembershipTest.php
+++ b/tests/src/Kernel/Entity/OgMembershipTest.php
@@ -490,6 +490,21 @@ class OgMembershipTest extends KernelTestBase {
   }
 
   /**
+   * Tests that we can retrieve the (empty) roles list from a new membership.
+   *
+   * If a membership is newly created and doesn't have a group associated with
+   * it yet, it should still be possible to get the (empty) list of roles
+   * without getting any errors.
+   *
+   * @covers ::getRoles
+   */
+  public function testGetRolesFromMembershipWithoutGroup() {
+    $membership = OgMembership::create();
+    $roles = $membership->getRoles();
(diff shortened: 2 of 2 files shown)


tests/src/Kernel/Entity/OgMembershipTest.php
```diff
@@ -489,6 +489,21 @@ public function testMemberRole() {
     $this->assertEquals(OgRoleInterface::AUTHENTICATED, $role->getName());
   }
 
+  /**
+   * Tests that we can retrieve the (empty) roles list from a new membership.
+   *
+   * If a membership is newly created and doesn't have a group associated with
```
amitaibu, line 495: I think we should prohibit this case if possible. That is, upon create - already pass the group and user.

pfrenssen: I'm doing this on an empty entity because it is the simplest way to prove the fatal error. In real life this occurs because the group entity is asynchronously deleted and memberships that are statically cached still reference it. This is not something we can prevent by simply requiring a group to be passed on instantiation. If you look at `OgMembership::getGroup()`, it gets the group entity from the entity storage, and this can always be deleted outside of the control of the membership.

```
// Create membership with a valid group.
$group = EntityTest::create();
$membership = Og::createMembership($group, $this->user);
$membership->save();

// Delete the group.
$group->delete();

// Fatal error.
$membership->getGroup()->bundle();
```

I had a look if it was feasible to prevent to create memberships without groups. I'm afraid it would have negative consequences because we would essentially be throwing an exception in `EntityStorageBase::create()` during the `::preCreate()` phase. It is legal to call `EntityStorageBase::create()` without passing any `$values`, so it is likely that there is a lot of code out there in the wild that does this.

I was looking for some examples in core and contrib where this pattern is used and it seems we are even doing it ourselves! In `MembershipManager::createMembership()` we create an empty entity, populate it and then save it. It looks like pretty standard Drupal code, I can imagine this type of code being very widespread:

```
  public function createMembership(EntityInterface $group, AccountInterface $user, $membership_type = OgMembershipInterface::TYPE_DEFAULT) {
    $membership = OgMembership::create(['type' => $membership_type]);
    $membership
      ->setOwner($user)
      ->setGroup($group);
    return $membership;
  }
```

So I had a further look, to see if the memberships have adequate protection for making sure they have correct groups and roles associated with it. It seems pretty good. We are validating them on presave, so we are basically guaranteeing that when the memberships are saved they have correct references.

amitaibu: Understood. Thanks

src/Entity/OgMembership.php
```diff
@@ -219,10 +219,14 @@ public function revokeRoleById($role_id) {
    * {@inheritdoc}
    */
   public function getRoles() {
+    $roles = [];
+
     // Add the member role.
-    $roles = [
-      OgRole::getRole($this->getGroupEntityType(), $this->getGroup()->bundle(), OgRoleInterface::AUTHENTICATED),
-    ];
+    if ($group = $this->getGroup()) {
```
amitaibu: 👍 Worth a comment to explain why we do this check

pfrenssen: Sure!

sandervd:
👍 RTBC  ;)

amitaibu:
Thanks!

amitaibu:
Thanks!

PR #358, opened 2018-01-29
Don't generate cache keys for users that have lost their membership role

pfrenssen:
Fixes the secondary problem from #356.

When a user has a certain role in a group but then the role is deleted from config, the membership entity still exists but contains an orphaned reference to role that was deleted. In this case no cache context keys should be generated for the user any more.


sandervd:
👍 RTBC ;)

amitaibu:
Thanks!

PR #360, opened 2018-03-08
Fix tests: core-quick-drupal not available in Drush 9

jacobembree:


amitaibu:
Great, seems @AronNovak that @jacobembree has already fixed the problem. Thanks folks!

alesrebec:
Yes, Travis is fixed with this PR which is accepted in 7.x-2.x now.
We just need to close and reopen PRs because "Restart build" doesn't take the updated configuration.

amitaibu:
Yeah, you'll need to merge 2.x-dev back to those PRs

PR #361, opened 2018-03-19
Change node_access_strict to entity_access_strict

zerolab:
Our project has a global/public side and various "networks" powered by OG.
We want to allow Media items to be OG content. Currently `og_entity_access()` makes it impossible to allow global editors to edit any content (that is non-node) that does not belong to a group.

Given `node_access_strict` is a D7 remnant, here's a PR that changes that to entity_access_strict to allow other entities do their magic when no in strict mode


og.module
```diff
@@ -159,14 +159,10 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
...
 
-  if ($entity_type_id == 'node') {
-    $node_access_strict = \Drupal::config('og.settings')->get('node_access_strict');
+  $entity_access_strict = \Drupal::config('og.settings')->get('entity_access_strict');
 
-    // Otherwise, ignore or deny based on whether strict node access is set.
-    return AccessResult::forbiddenIf($node_access_strict);
-  }
-
-  return AccessResult::forbidden();
```
pfrenssen, line 172: This makes sense to me, I don't see the point of special casing nodes, and providing a blanket ban on any other entity types.

og.module
```diff
@@ -159,14 +159,10 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
...
+  // Otherwise, ignore or deny based on whether strict node access is set.
+  return AccessResult::forbiddenIf($entity_access_strict);
```
pfrenssen, line 168: This comment should be updated, it's no longer applicable.

og.module
```diff
@@ -185,10 +181,19 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
...
+    if ($entity_type_id == 'node' && $account->hasPermission("create $bundle content")) {
+      // The user has the core permission and strict entity access is not set.
+      return AccessResult::neutral();
+    }
+
+    // Handle other entities.
+    if ($account->hasPermission("create $entity_type_id") ||
+        $account->hasPermission("create {$bundle} {$entity_type_id}")) {
+      // The user has the core permission and strict entity access is not set.
+      return AccessResult::neutral();
```
pfrenssen, line 199: This is problematic :( The permissions are just arbitrary strings so we cannot guarantee this to work.

This will work for a number of cases, but not all. I did a quick search in core, and it wouldn't work with the Book module for example (permission is `create new books`).

Until we have a structured way of handling permissions we'll have to allow modules to provide their own data. It seems this is being worked on in [Issue #2809177: Introduce entity permission providers and access handlers]()https://www.drupal.org/project/drupal/issues/2809177) but that is not in yet.

I think a good way to tackle this is to provide an event for this, and subscribe to it ourselves to return the list of permissions provided by the core modules. Developers can then implement a subscriber to provide any missing permissions.

pfrenssen (changes requested):
We can also use some test coverage

og.module
heddn: +1 on the event idea.

amitaibu: Seems we already have a better handling of entity permissions in OG -- https://github.com/Gizra/og/issues/510#issuecomment-496440783

og.module
```diff
@@ -160,14 +160,10 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
...
-  if ($entity_type_id == 'node') {
-    $node_access_strict = \Drupal::config('og.settings')->get('node_access_strict');
+  $entity_access_strict = \Drupal::config('og.settings')->get('entity_access_strict');
 
-    // Otherwise, ignore or deny based on whether strict node access is set.
-    return AccessResult::forbiddenIf($node_access_strict);
-  }
-
-  return AccessResult::forbidden();
+  // Otherwise, ignore or deny based on whether strict node access is set.
```
```suggestion
  // Otherwise, ignore or deny based on whether strict entity access is set.
```

claudiu-cristea:

PR #365, opened 2018-04-24
Use core private tempstore.

### LINKED ISSUE #364: Replace deprecated classes
There are a few uses of `\Drupal\user\PrivateTempStore` that need to be replaced with the non-deprecated version.

jhedstrom:
- Fixes #364

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ php:
   - 7.1
 
 env:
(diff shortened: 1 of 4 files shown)


pfrenssen:
There were some more problems related to recent changes in Drupal core. These have in the meantime been resolved in #395. That PR also included the private temp store.

Thanks!

PR #368, opened 2018-05-16
Og field access bybass

emerham:
We needed a way to grant roles a permission that then could bypass all field access permissions, kinda like how user_1 is super user.

### DIFF
--- a/og_field_access/og_field_access.module
+++ b/og_field_access/og_field_access.module
@@ -25,6 +25,10 @@ function og_field_access_field_access($op, $field, $entity_type, $entity, $accou
   if ($entity_type == 'user' && !$id) {
     return;
   }
+  // Check if the user has special permission and content is group content.
+  if (user_access('bypass og field access') && og_is_group_content_type($entity_type, $bundle)) {
+    return TRUE;
+  }
   if (!$id && $op == 'edit' && (og_is_group($entity_type, $entity) || og_is_group_content_type($entity_type, $bundle))) {
     // This is create form of a non-saved entity, so we check
     // permissions to access the field, for all the groups the user is a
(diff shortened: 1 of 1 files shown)


og_field_access/og_field_access.module
```diff
@@ -25,6 +25,10 @@ function og_field_access_field_access($op, $field, $entity_type, $entity, $accou
   if ($entity_type == 'user' && !$id) {
     return;
   }
+  // Check if the user has special permission and content is group content.
+  if (user_access('bypass og field access') && og_is_group_content_type($entity_type, $bundle)) {
```
alesrebec: You have 
```
if (user_access('administer group')) {
     return TRUE;
   }
```
above. Why would you need another bypass permission?

alesrebec:
Can you specify why would OG need this "bypass" permission if you can use "administer group" permission that is implemented few lines above?

```
function og_field_access_field_access($op, $field, $entity_type, $entity, $account) {
  ...
  if (user_access('administer group')) {
    return TRUE;
  }
  ...

// Check if the user has special permission and content is group content.
  if (user_access('bypass og field access') && og_is_group_content_type($entity_type, $bundle)) {
    return TRUE;
  }
...
```

emerham:
For our use case, we didn't want to put the the user or users in every group as a manager, nor give them mange group permissions because they are only site authors. We have about 200 groups, and they cannot be site admins but needed access to edit all fields that are in their site. It is really a unique use case for this situation, so I have a patch up on d.o to use drush make patching.

If it is too unique for general use I'm fine letting it live on as a patch that gets added into the distro, but figured I'd at least let others have a look see.

claudiu-cristea:

Issue #369, opened 2018-05-17
OgRole::isAdmin() should return TRUE for administrators


pfrenssen:
`OgRole` inherits the `::isAdmin()` method from the standard core `RoleInterface`.

Unfortunately we never implemented this method, so currently is always returns `FALSE`. The expected result is that it returns `TRUE` if the role represents an administrator.

pfrenssen:
This is actually working as designed, I had the wrong config exported in my project since our config predates the implementation of the administrator access checks.

PR #370, opened 2018-05-17
Provide tests for `OgRole::isAdmin()` and `OgRole::setIsAdmin()`

pfrenssen:
Fix for #369.

Edit: I was mistaken, the `OgRole::isAdmin()` and `OgRole::setIsAdmin()` methods were not malfunctioning due a bug but due to a misconfiguration in my project.

I've now repurposed this PR to provide some additional test coverage that demonstrates the use of these methods.

### DIFF
--- a/tests/src/Unit/OgRoleTest.php
+++ b/tests/src/Unit/OgRoleTest.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Drupal\Tests\og\Unit;
+
+use Drupal\og\Entity\OgRole;
+use Drupal\Tests\UnitTestCase;
+
+/**
+ * Unit tests for the OgRole config entity.
+ *
+ * @group og
+ * @coversDefaultClass \Drupal\og\Entity\OgRole
+ */
+class OgRoleTest extends UnitTestCase {
+
+  /**
+   * Tests getting and setting the admin role through the inherited methods.
+   *
+   * @param bool $value
+   *   A boolean value whether or not the admin role will be set.
+   *
+   * @covers ::isAdmin
+   * @covers ::setIsAdmin
+   *
+   * @dataProvider booleanProvider
+   */
+  public function testIsAdmin($value) {
+    $role = new OgRole([]);
+    $role->setIsAdmin($value);
+    $this->assertEquals($value, $role->isAdmin());
+  }
+
+  /**
+   * Provides boolean data.
+   */
+  public function booleanProvider() {
+    return [[TRUE], [FALSE]];
+  }
+
+}
--- a/tests/src/Kernel/Access/OgEntityAccessTest.php
+++ b/tests/src/Kernel/Access/OgEntityAccessTest.php
@@ -65,6 +65,13 @@ class OgEntityAccessTest extends KernelTestBase {
   protected $adminUser;
 
   /**
+   * A second administrator which has an alternative administration role.
+   *
+   * @var \Drupal\user\Entity\User
+   */
+  protected $alternativeAdminUser;
+
+  /**
    * A group entity.
    *
    * @var \Drupal\entity_test\Entity\EntityTest
@@ -121,6 +128,13 @@ class OgEntityAccessTest extends KernelTestBase {
   protected $ogAdminRole;
 
   /**
+   * A custom OG admin role.
+   *
+   * @var \Drupal\og\Entity\OgRole
+   */
+  protected $ogAlternativeAdminRole;
+
+  /**
    * {@inheritdoc}
    */
   protected function setUp() {
@@ -159,6 +173,10 @@ class OgEntityAccessTest extends KernelTestBase {
     // Admin user.
     $this->adminUser = User::create(['name' => $this->randomString()]);
     $this->adminUser->save();
+
+    // Second admin user which uses an alternative administration role.
+    $this->alternativeAdminUser = User::create(['name' => $this->randomString()]);
+    $this->alternativeAdminUser->save();
 
     // Declare the test entity as being a group.
     Og::groupTypeManager()->addGroup('entity_test', $this->groupBundle);
@@ -218,8 +236,14 @@ class OgEntityAccessTest extends KernelTestBase {
       ->grantPermission($this->randomMachineName())
       ->save();
 
-    $this->ogAdminRole = OgRole::create();
-    $this->ogAdminRole
+    // The administrator role is added automatically when the group is created.
+    // @see \Drupal\og\EventSubscriber\OgEventSubscriber::provideDefaultRoles()
+    $this->ogAdminRole = OgRole::loadByGroupAndName($this->group1, OgRoleInterface::ADMINISTRATOR);
+
+    // Create a second administration role, since this is a supported use case.
+    // It is possible to have multiple administration roles.
+    $this->ogAlternativeAdminRole = OgRole::create();
+    $this->ogAlternativeAdminRole
       ->setName($this->randomMachineName())
       ->setLabel($this->randomString())
       ->setGroupType($this->group1->getEntityTypeId())
@@ -254,6 +278,11 @@ class OgEntityAccessTest extends KernelTestBase {
     $membership = Og::createMembership($this->group1, $this->adminUser);
     $membership
       ->addRole($this->ogAdminRole)
+      ->save();
+
+    $membership = Og::createMembership($this->group1, $this->alternativeAdminUser);
+    $membership
+      ->addRole($this->ogAlternativeAdminRole)
       ->save();
   }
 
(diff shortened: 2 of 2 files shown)


pfrenssen:
Travis tests are running but the results don't seem to be reported back.

Here is the failing test: https://travis-ci.org/Gizra/og/builds/380149838
And the fix that turns the test green: https://travis-ci.org/Gizra/og/builds/380152585

tests/src/Unit/OgRoleTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Drupal\Tests\og\Unit;
+
+use Drupal\og\Entity\OgRole;
+use Drupal\og\OgRoleInterface;
+use Drupal\Tests\UnitTestCase;
+
+/**
+ * Tests the OgContext context provider.
```
amitaibu: Wrong comment

amitaibu (commented):
Looks good, just a tiny comment.

pfrenssen:
This apparently breaks something, `OgEntityAccessTest::testAccess()` is failing.

pfrenssen:
It's the test that is wrong, instead of setting the role name to `OgRoleInterface::ADMINISTRATOR`, it sets the role name to a random value, and it calls `OgRole::setIsAdmin()` instead.

This points out that we also have to implement `::setIsAdmin()`. We can easily set it (by assigning the value `OgRoleInterface::ADMINISTRATOR` to the role name), but _unsetting_ the admin role by calling `::setIsAdmin(FALSE)` is problematic. We don't know what name to set it to.

I think a good way to handle this case is to set the role name to `OgRoleInterface::AUTHENTICATED` and emitting a PHP warning alongside it, to alert developers that they should instead set the role name directly.

Edit: ~~even better would be to retain the underlying mechanic, and set the value dynamically when the role name is set to `OgRoleInterface::ADMINISTRATOR`.~~ Edit: no, this could put us in the awkward position that a role would be called 'administrator' but is actually not an administrator.

pfrenssen:
@amitaibu could it be possible that the Travis CI integration was disabled when the Gizra organisation on Github enabled OAuth app restrictions recently?

amitaibu:
I don't think thats the case

apps_and_integrations_

@AronNovak have you seen travis not working in other Gizra repos? (I haven't)

pfrenssen:
Ah yeah that makes sense, otherwise the tests wouldn't even start I suppose. I just find it strange that the tests results are not reported back.

amitaibu:
I suspect it's a travis-ci hickup

AronNovak:
@amitaibu :+1: on this, i did not experience any issues at Travis projects. However those are closed projects, I'll check at Elm Starter ASAP.

AronNovak:
Hm, actually I used Travis at Elm Starter yesterday, then it must be a hiccup, almost certainly.

pfrenssen:
Thanks for having a look @AronNovak !

PR looks good now, tests are passing. What was changed:

* Dropped the storage of the `is_admin` configuration, it wasn't working before, and we don't need it since we can infer it from the role name.
* Implemented `::setIsAdmin()` while throwing a warning to alert developers that this has the unexpected side effect of changing the role name. I also added a test for this implementation.
* Fixed the `OgEntityAccessTest`. The setup of the test was indeed wrong, in reality the admin role is already created when a group is created, so we should use that one.

pfrenssen:
Here is the test result of the last commit: https://travis-ci.org/Gizra/og/builds/380237297

src/Entity/OgRole.php
```diff
@@ -269,6 +268,33 @@ public function isRequired() {
...
+    // This method has been inherited from the core RoleInterface but has the
+    // unexpected side effect that it will change the role name. The recommended
+    // way of defining the administrator role is by setting the role name to
+    // OgRoleInterface::ADMINISTRATOR, for example:
+    // @code
+    //   $role->setName(OgRoleInterface::ADMINISTRATOR);
+    // @endcode
+    trigger_error('Using ::setIsAdmin() affects the role name.', E_USER_WARNING);
+
+    $this->setName($is_admin ? OgRoleInterface::ADMINISTRATOR : OgRoleInterface::AUTHENTICATED);
```
amitaibu: This doesn't feel right. We say and `admin` is defined by having a role named `admin`, but if it's not an admin, the role name can be whatever we want.

Maybe we should just throw an Exception, if this method is used?

pfrenssen: OK, that's also a good solution, alerting the developer using an exception.

@amitaibu to be 100% clear, it is the intention that a group admin has all permissions in the group, right? Because that is wat `::isAdmin()` promises:

```
  /**
   * Indicates that a role has all available permissions.
   *
   * @return bool
   *   TRUE if the role has all permissions.
   */
  public function isAdmin();
```

Reason I am asking is that in my project we have 1 test failing with this PR applied: we have a use case where the group admin is not allowed to delete content, because the content needs to be retained for archival reasons. With the PR the admin gets this permission.

So my understanding is that this is intentional, and on our end we should implement `hook_og_user_access_alter()` to take away the permission from the group admin?

idimopoulos: ahh, hello :) There is an explicit `is_admin` flag that controls this in core. We can use this. We don't need to restrict it in the administrator role.
The og_roles have inherited this flag and are already in exported configurations. Core also bases this in this flag.

*Edit:*
As I discussed with Pieter in the chat, here are some food for thoughts from our chat.
Currently we are using the `OgRoleInterface::ADMINISTRATOR` constant to define if the user is an administrator. This was the corresponding `user 1` of D7 for groups. My comment was not about the current implementation being wrong. This is merely a design decision which is working fine. My comment was that since core has decided to include an extra flag that makes a role an administrator (and not some global `grant all privileges; flush privileges` permission), by keeping the same administrator named role design, it is still valid, but we are deviating from core. We could use the flag in order to determine the power of the role since it is offered by core and then someone who is accustomed in core and just enters og will find it quite familiar.

pfrenssen: Thanks for the input @idimopoulos! I can see the merit of your proposal. It would simplify this PR, and we would stay closer in spirit to the core implementation, while still retaining the full flexibility.

So to be clear, is this what you are suggesting?

* We keep the `is_admin` flag in exported config. When creating a new group type this flag will be automatically set when the admin role is auto-created.
* We keep the core mechanic of determining the admin role solely by this flag, so we will ignore the role name.

Advantages:

* We don't need to override `::isAdmin()` and `::setIsAdmin()` or prevent their usage. They will work as designed.
* The intended behavior remains the same: when a new group type is created, we will have an admin role that has the role name "administrator" and it will have the flag set.
* Developers have the flexibility to create multiple admin roles, just like it can be done in core. I don't see many use cases for this, but there might be.
* Developers have the flexibility to revoke admin rights from the pre-created admin role, just like they can do in core, by calling `$role->setIsAdmin(FALSE)->save()`.

Disadvantages:

* It is possible to have a role that is called "administrator" but is in fact a regular role. This is a minor drawback, it matches the core functionality. It also requires a manual action from a developer to get into this state.

pfrenssen:
So I found this snippet in `OgAccess::userAccess()`:

```
      if ($membership = $this->membershipManager->getMembership($group, $user)) {
        foreach ($membership->getRoles() as $role) {
          // Check for the is_admin flag.
          /** @var \Drupal\og\Entity\OgRole $role */
          if ($role->isAdmin()) { // <<<< here it calls the inherited method
            $user_is_group_admin = TRUE;
            break;
          }

          $permissions = array_merge($permissions, $role->getPermissions());
        }
      }
```

It looks like this IS in fact implemented correctly. I also see that the flag is being set when the admin role is being created in `OgEventSubscriber::provideDefaultRoles()`:

```
    $role = $this->entityTypeManager->getStorage('og_role')->create([
      'name' => OgRoleInterface::ADMINISTRATOR,
      'label' => 'Administrator',
      'is_admin' => TRUE, // <<<<< Here it is being set.
    ]);
```

It also appears that I am the person who implemented this back in 2016 :confounded:

pfrenssen:
OK I found the source of my confusion for this PR. It appears that in our project we have exported our initial group types in May 2016, which is 1 month before this admin role functionality was implemented in OG. We simply have the wrong data exported.

Sorry for the noise. I will repurpose this PR to just update the test coverage for this.

pfrenssen:
I have written the additional test coverage that demonstrates that these methods actually work as designed. I've rebased the PR so this is just a single commit without the noise of the earlier implementation.

Test is running here: https://travis-ci.org/Gizra/og/builds/380706308

pfrenssen:
Test is green, up for review!

Ohmdesbois:
Bravo! and thank you !

pfrenssen:
You're welcome :)

amitaibu:
Looks good, thanks!

Issue #373, opened 2018-05-28
When an entity is not part of a group but the type could be part of it, the access is blocked


DavidHernandez:
The `og_entity_access` hook checks if a entity type and bundle can be content of a group. But it does not check if the actual entity is content of a group. This causes some entities with an audience field marked as not required be blocked by the hook implementation.

Here are some steps to replicate the issue:
- Create a group entity.
- Create a new content type with the audience field, mark it as not required.
- Use an user that is not admin, to create a new node of that content type but do not fill the audience field.
- Try to edit the node. It will return an access denied.

MPParsley:
It seems this has been fixed in #374.

PR #374, opened 2018-05-28
Adds new access check to `og_entity_access`

DavidHernandez:
This adds new function to verify if an entity is actually part of a group. Updates og_entity_access to make sure the module does not block access to entities that can be part of a group but they are not.

This should fix the error reported on #373


DavidHernandez:
I'm aware this needs a test, I will add it as soon as I can.

src/Og.php
```diff
@@ -269,6 +269,36 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+   * This works by checking if the bundle has at least one group audience field
+   * filled.
+   *
+   * @param object $entity
+   *   The entity object.
+   *
+   * @return bool
+   *   True or false if the given entity is group content.
+   */
+  public static function isEntityGroupContent($entity) {
```
mariano-dagostino: I'm not sure if this is an appropriated name for this method. What are you trying to do? Asserting some content is referencing some group?

Also, Maybe we could return $groups instead of TRUE and FALSE, that way you can decide if this is relevant to you by inspecting the array. And you get a useful method (I remember seeing something similar in #67 )

mariano-dagostino (changes requested):

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
     return \Drupal::service('og.group_audience_helper')->hasGroupAudienceField($entity_type_id, $bundle_id);
   }
 
+  /**
+   * Check if the given entity is a group content.
```
mariano-dagostino: Docs seems not up to date.

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+   * This works by checking if the bundle has at least one group audience field
+   * filled.
+   *
+   * @param object $entity
+   *   The entity object.
+   *
+   * @return array
+   *   List of groups where the entity belongs.
+   */
+  public static function getEntityGroups($entity) {
```
mariano-dagostino: You could use `EntityInterface $entity` as parammeter.

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+    $can_be_group_content = \Drupal::service('og.group_audience_helper')->hasGroupAudienceField($entity->type, $entity->bundle);
```
mariano-dagostino: You could save the service into a variable to avoid call it twice. `\Drupal::service('og.group_audience_helper')`

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
     return \Drupal::service('og.group_audience_helper')->hasGroupAudienceField($entity_type_id, $bundle_id);
   }
 
+  /**
+   * Check if the given entity is a group content.
+   *
+   * This works by checking if the bundle has at least one group audience field
+   * filled.
+   *
+   * @param object $entity
```
mariano-dagostino: Instead of object, this could be  `EntityInterface`

mariano-dagostino (changes requested):

og.module
```diff
@@ -147,6 +147,10 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
     return AccessResult::neutral();
   }
 
+  if (!empty(Og::getEntityGroups($entity))) {
```
amitaibu: => `if (empty`

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity object.
+   *
+   * @return array
+   *   List of groups where the entity belongs.
+   */
+  public static function getEntityGroups(EntityInterface $entity) {
+    $can_be_group_content = \Drupal::service('og.group_audience_helper')->hasGroupAudienceField($entity->getEntityTypeId(), $entity->bundle());
+    if (!$can_be_group_content) {
+      return FALSE;
```
amitaibu: I think it should be either an empty array - but maybe even better throw an Exception?

Not sure, which one would be better. @pfrenssen any thoughts?

tests/src/Functional/GroupContentAccessTest.php
```diff
@@ -0,0 +1,113 @@
...
+use Drupal\simpletest\BrowserTestBase;
+use Drupal\simpletest\ContentTypeCreationTrait;
+use Drupal\simpletest\NodeCreationTrait;
+
+/**
+ * Tests the access to content that belongs to a group.
+ *
+ * @group og
+ */
+class GroupContentAccessTest extends BrowserTestBase {
```
amitaibu: Kernel tests are going to be quicker, and are fitting this scenario.

tests/src/Unit/OgAccessEntityTestBase.php
```diff
@@ -48,13 +48,19 @@ public function setup() {
     $this->groupContentEntity->isNew()->willReturn(FALSE);
     $this->groupContentEntity->getEntityType()->willReturn($entity_type->reveal());
     $this->groupContentEntity->getEntityTypeId()->willReturn($entity_type_id);
+    // We should be returning a field and mocking the referencedEntities there, but I do not know how to do it.
```
amitaibu: @mariano-dagostino can you have a look please (or ping if you get stuck)

mariano-dagostino: @DavidHernandez you could do something like this. However, I'm not really sure why would you try to mock everything if you are writing a Kernel test to test the actual hook.

```diff
diff --git a/tests/src/Unit/OgAccessEntityTestBase.php b/tests/src/Unit/OgAccessEntityTestBase.php
index 71270f5..f18fb1a 100644
--- a/tests/src/Unit/OgAccessEntityTestBase.php
+++ b/tests/src/Unit/OgAccessEntityTestBase.php
@@ -8,6 +8,7 @@ use Drupal\Core\Entity\EntityStorageInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\FieldableEntityInterface;
+use Drupal\Core\Field\EntityReferenceFieldItemList;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\og\OgGroupAudienceHelperInterface;
@@ -42,6 +43,15 @@ class OgAccessEntityTestBase extends OgAccessTestBase {
     $entity_type->isSubclassOf(FieldableEntityInterface::class)->willReturn(TRUE);
     $entity_type->id()->willReturn($entity_type_id);
 
+    // Some mocked groups that are referenced by the content entity.
+    $groups = [
+      $this->prophesize(ContentEntityInterface::class),
+      $this->prophesize(ContentEntityInterface::class),
+    ];
+
+    $og_audience = $this->prophesize(EntityReferenceFieldItemList::class);
+    $og_audience->referencedEntities()->willReturn($groups);
+
     $this->groupContentEntity = $this->prophesize(ContentEntityInterface::class);
     $this->groupContentEntity->id()->willReturn($entity_id);
     $this->groupContentEntity->bundle()->willReturn($bundle);
@@ -49,8 +59,7 @@ class OgAccessEntityTestBase extends OgAccessTestBase {
     $this->groupContentEntity->getEntityType()->willReturn($entity_type->reveal());
     $this->groupContentEntity->getEntityTypeId()->willReturn($entity_type_id);
     // We should be returning a field and mocking the referencedEntities there, but I do not know how to do it.
-    $this->groupContentEntity->get('test')->willReturn($this->groupContentEntity);
-    $this->groupContentEntity->referencedEntities()->willReturn(['test']);
+    $this->groupContentEntity->get('og_audience')->willReturn($og_audience);
     $this->addCache($this->groupContentEntity);
 
     // If the group audience helper is asked if the group content entity has any
@@ -59,7 +68,7 @@ class OgAccessEntityTestBase extends OgAccessTestBase {
       ->willReturn(TRUE);
 
     $this->groupAudienceHelper->getAllGroupAudienceFields($entity_type_id, $bundle)
-      ->willReturn(['test' => 'test']);
+      ->willReturn(['og_audience' => 'og_audience']);
 
     // It is expected that a list of entity operation permissions is retrieved
     // from the permission manager so that the passed in permission can be

```

src/Og.php
```diff
@@ -269,6 +269,34 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+   * filled.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity object.
+   *
+   * @return array
+   *   List of groups where the entity belongs.
+   */
+  public static function getEntityGroups(EntityInterface $entity) {
+    $can_be_group_content = \Drupal::service('og.group_audience_helper')->hasGroupAudienceField($entity->getEntityTypeId(), $entity->bundle());
```
mariano-dagostino: Could you save the service into a variable, so you don't need to invoke `Drupal::service` twice?

tests/src/Unit/OgAccessEntityTestBase.php
```diff
@@ -42,19 +43,32 @@ public function setup() {
...
+    $og_audience = $this->prophesize(EntityReferenceFieldItemList::class);
+    $og_audience->referencedEntities()->willReturn($groups);
+
     $this->groupContentEntity = $this->prophesize(ContentEntityInterface::class);
     $this->groupContentEntity->id()->willReturn($entity_id);
     $this->groupContentEntity->bundle()->willReturn($bundle);
     $this->groupContentEntity->isNew()->willReturn(FALSE);
     $this->groupContentEntity->getEntityType()->willReturn($entity_type->reveal());
     $this->groupContentEntity->getEntityTypeId()->willReturn($entity_type_id);
+    $this->groupContentEntity->get('og_audience')->willReturn($og_audience);
```
mariano-dagostino: This og_audience can be replaced by `OgGroupAudienceHelperInterface::DEFAULT_FIELD`

src/Og.php
DavidHernandez: Those functions are static, so they are invoked without the object being initialized. We could not start the service on the constructor. So I think this is the best solution that comes to my mind.

amitaibu: no, I believe he means 

```php
$audience_helper = \Drupal::service('og.group_audience_helper');
$audience_fields = audience_helper->getAllGroupAudienc ...
```

mariano-dagostino (approved):
Looks good to me now.

src/Og.php
```diff
@@ -269,6 +269,36 @@ public static function isGroupContent($entity_type_id, $bundle_id) {
...
+   * filled.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity object.
+   *
+   * @return array
+   *   List of groups where the entity belongs.
+   */
+  public static function getEntityGroups(EntityInterface $entity) {
+    $audience_helper = \Drupal::service('og.group_audience_helper');
```
pfrenssen: This looks like it duplicates `MembershipManager::getGroups()`.

og.module
```diff
@@ -142,8 +142,13 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
...
   $entity_type_id = $entity->getEntityTypeId();
   $bundle_id = $entity->bundle();
+  $is_group_content = Og::isGroupContent($entity_type_id, $bundle_id);
 
-  if (!Og::isGroup($entity_type_id, $bundle_id) && !Og::isGroupContent($entity_type_id, $bundle_id)) {
+  if (!Og::isGroup($entity_type_id, $bundle_id) && !$is_group_content) {
+    return AccessResult::neutral();
+  }
+
+  if (empty(Og::getEntityGroups($entity)) && $is_group_content) {
```
pfrenssen: If you just need to know whether an entity has _any_ groups then it is better to do this:

```
$group_count = \Drupal::service('og_membership_manager')->getGroupCount($entity);
if ($group_count && !$is_group_content) {}
```

This has a few advantages: it caches the results to speed up multiple lookups, it will check against the database that there are no stale references left behind by the Entity Reference module causing false positives, and it will avoid loading the full entity objects.

pfrenssen:
@DavidHernandez I'm sorry you are experiencing these failures. I had a quick look at the failing unit tests in `\Drupal\Tests\og\Unit\OgAccessHookTest`.

These tests are not correctly implemented, they should at some point be refactored into Kernel tests. For properly testing hooks we have to bootstrap Drupal, so Kernel tests are the way to go. Also in general it is a bad idea to use unit tests for testing procedural code since it is typically not well structured. For now we can update the unit test but this would mean to replicate the entire logic of the hook implementation into the test. As I said I am sorry you are hitting this in this PR, you should not have to deal with this mess at this point :(

I am willing to assist you with this but I won't have time right away. In 2 weeks I will be sprinting at Drupal Dev Days and I can dive into this. I don't know if this functionality is urgent for your project?

DavidHernandez:
@pfrenssen I was able to fix the unit tests. I will try to invest a bit of more time in order to fix the functional tests too. Ideally, we would like to have this ready during this sprint, that ends this Friday, but is not really necessary. It can wait after the DDD.

pfrenssen:
@DavidHernandez that's great! I have made a followup to convert this test into a Kernel test so we will be able to work on the access hook in the future without breaking this brittle test: #391

DavidHernandez:
@pfrenssen tests are green now. Could you review this when you have the chance?

pfrenssen:
@DavidHernandez Unfortunately I have no availability right now but I will have time for this next week!

pfrenssen:
I reviewed this, it looks good. I just added a few lines of documentation in the access hook to explain what we are doing.

If tests are green this can be merged.

pfrenssen:
Tests are green, great!

Thanks all for working on this!

Issue #377, opened 2018-05-31
Make sure OG UI tests are executed even if a failure occurs in OG tests


pfrenssen:
Starting in #375 we are now also testing OG UI properly. The OG tests are now being run first, followed by the OG UI tests.

However if an error occurs in the first test run (OG), then the test script will exit with error code 1, and the second test run (OG UI) will not be executed:

```
for i in ${TEST_DIRS[@]}; do
  echo " > Executing tests from $i"
  ./vendor/bin/phpunit -c ./core/phpunit.xml.dist $i || exit 1
done
```

PR #378, opened 2018-05-31
Make sure the OG UI test still runs even if the OG test fails

pfrenssen:
Fix for #377


AronNovak (approved):
Looks great! There are two approaches, it can be argued that early interruption of a failure is a good idea as it decreases the execution time. On the other hand, if after a failure, we continue, the full feedback is given at the first run, both have its advantages, it's equally fine.

pfrenssen:
Thanks Aron!

PR #389, opened 2018-06-19
Use correct class name in exception message

pfrenssen:
When we renamed `GroupMembership` to `MembershipManager` it appears we had forgotten to update it in this exception message.


amitaibu:
Looks good. Merge when green

pfrenssen:
Thanks for the quick review!

PR #390, opened 2018-06-20
OG 7.x 2.10 release

### LINKED ISSUE #120: New member state is pending, should be active when Admin adds user via profile widget

### LINKED ISSUE #125: Behavior for og_membership_delete_action

alesrebec:
**Commits between 7.x-2.9 and 7.x-2.10-rc1**

 * #2057025 OG subscribe / join page not working with field permissions on group field.
 * #1865852 og_form_group_manager_validate() assumes the entity form has a language.
 * #2831715 Unsubscribe confirm firm uses 'Remove' as button text.
 * #2338665 Remove unnecessary files[] from .info file.
 * #2542856 og_is_member() documentation incorrectly inform about  parameter.
 * #2575833 OG entityreference field breaks Views when the field is attached to multiple entity types..
 * #1997274 Return docs for og_get_entity_groups() unclear.
 * #1865860 Issues with multiple entity types as groups.
 * #2158477 hook_og_permission() does not document its @return.
 * Issue #2792031: Undefined property in og_entity_update when updating group
 * Fix tests: core-quick-drupal not available in Drush 9
 * Fix return value of og_get_groups_by_user() when $group_type is provided and results are empty
 * Issue #2569471 by robertom, jacob.embree: og_get_groups_by_user don't respect @return documentation
 * Remove duplicate checks in og_ui_user_access_group
 * Optimize some calls to og_is_group()
 * Revert "Revert "temporarily downgrade Travis image to confirm the problem""
 * Revert "line-by-line variables"
 * line-by-line variables
 * Revert "temporarily downgrade Travis image to confirm the problem"
 * temporarily downgrade Travis image to confirm the problem
 * Group owner membership is activated on every group save
 * og_ui edit_membership_form disable checkboxes if member is not active.
 * Issue #2380967 by rbayliss, criznach: Edit Group membership of a user and changing the status
 * Wrong references for og_ui_admin_roles validate and submit functions.
 * Issue #121: Changed form element type for check_access from select to checkbox and changed description to match the change.
 * og_ui: fix non-existing group test
 * og_ui.test: aside 403 group admin page tests, add a regular 404 response test too
 * Revert "og_ui/og_ui.test: temporary test break to test Travis error detection"
 * og_ui/og_ui.test: temporary test break to test Travis error detection
 * .travis.yml: quote bash commands to preserve special characters
 * .travis.yml: negate the result of the egrep simply when checking the result of simpletest
 * .travis.yml: fix error checking
 * .travis.yml: fix mysql server has gone away, from https://github.com/piwik/piwik/pull/6963/files
 * removed unreachable code in OgBehaviorHandler.class.php . If the $user has group admin permission, we should not set new membership to pending. But if he is not an admin, we exit early from the membership Crud. Adapted the test to test admin permission check of the Crud itself
 * og_ui/og_ui.test: fixed failing access test. Drush + built-in PHP webserver causes whatever/10.00/group path a 404, it does not reach Drupal. Instead of patching Drush, rewrote the test to check similar, but a bit different scenario
 * .travis.yml: fix sendmail stubbing for newer Drush
 * .travis.yml: fix Drush path, add PHP7 to the matrix
 * .travis.yml: switch to Composer-based Drush due to https://github.com/drush-ops/drush/issues/2090
 * .travis.yml: move the launch of the webserver to the proper location
 * .travis.yml: restore local webserver
 * .travis.yml: use testing script's --php version, path to php is not trivial due to phpenv
 * .travis.yml: assume that before_script change directory applies to script too
 * .travis.yml: restore sendmail stubbing
 * .travis.yml: download Drupal 7.x, not 8.x
 * .travis.yml: do not alter sendmail_path for site install
 * .travis.yml: initial attempt to fix the test logic for 7.x
 * Issue #2451857 by justanothermark: Extend og_user_access_entity to support skip_alter and ignore_admin
 * Issue #2411041 by hefox, mpotter, filsterjisah: Group manager always given default default role as og_is_member arguments wrong
 * Issue #2143733 by czigor, Angry Dan, dobe: EntityMalformedException with select og_audience widget if group content has no other fields
 * Issue #2471581 by Chewie, shelane: Missed asterisk for required Group audience field
 * Issue #1781386 by robertom, milos.kroulik: Context only set on primary node view page, nothing else
 * Fix tests by adding a missing new Drupal core permission
 * Fixed: og_help function in og.module refers now README.md.
 * Issue #2250177 by joelpittet, Devin Carlson, ezra-g: Improve performance of og_is_group()
 * Issue #121: Added option_definition for ['check_access'].
 * Issue #121: Added option in og_context_plugin_argument_default_group_context.inc to accompany the permission rewrite in og_context().
 * Fixed #125: Behavior for og_membership_delete_action
 * Fixed #120: New member state is pending, should be active when Admin adds user via profile widget
 * Issue #2151015 by hass, tmckeown, iaminawe: Migrate warnings

### DIFF
--- a/plugins/entityreference/selection/OgSelectionHandler.class.php
+++ b/plugins/entityreference/selection/OgSelectionHandler.class.php
@@ -105,7 +105,6 @@ class OgSelectionHandler extends EntityReference_SelectionHandler_Generic {
 
     $field_mode = $this->instance['field_mode'];
     $user_groups = og_get_groups_by_user(NULL, $group_type);
-    $user_groups = $user_groups ? $user_groups : array();
     $user_groups = array_merge($user_groups, $this->getGidsForCreate());
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_script:
   - cd ../..
 
   # Install Drush
-  - composer global require drush/drush
+  - composer global require drush/drush:8
   - export PATH=$PATH:~/.composer/vendor/bin
 
   # Install php packages required for running Drupal
--- a/og.api.php
+++ b/og.api.php
@@ -12,7 +12,28 @@
  */
 
 /**
- * Add group permissions.
+ * Define permissions specific to organic groups.
+ *
+ * This hook can supply permissions that the module defines, so that they can be
+ * selected on the group permissions page and used to grant or restrict access
+ * to actions pertaining to groups.
+ *
+ * Permissions are checked using og_user_access().
+ *
+ * @return array
+ *   An array whose keys are permission names and whose corresponding values are
+ *   arrays containg the following key-value pairs:
+ *   - title: The human-readable name of the permission, to be shown on group
+ *     permissions pages. This should be wrapped in the t() function so it can
+ *     be translated.
+ *   - description: (optional) A description of what the permission does. This
+ *     should be wrapped in the t() function so it can be translated.
+ *   - roles: (optional) An array of OG roles to which it is possible to assign
+ *     this permission. The default value is an array containg OG_ANONYMOUS_ROLE
+ *     and OG_AUTHENTICATED_ROLE.
+ *   - default role: (optional) An array of OG roles that should be assigned
+ *     this permission by default. If omitted then no role will receive the
+ *     permission by default.
  */
 function hook_og_permission() {
   return array(
--- a/og.info
+++ b/og.info
@@ -24,7 +24,6 @@ files[] = includes/views/handlers/og_plugin_argument_validate_group.inc
 files[] = includes/views/handlers/og_plugin_argument_default_user_groups.inc
 
 ; Views fields
-files[] = includes/views/handlers/og_handler_field_group_member_count.inc
 files[] = includes/views/handlers/og_handler_field_group_audience_state.inc
 files[] = includes/views/handlers/og_handler_field_prerender_list.inc
 files[] = includes/views/handlers/og_handler_field_user_roles.inc
--- a/og.module
+++ b/og.module
@@ -81,8 +81,8 @@ function og_help($path, $arg) {
   switch ($path) {
     case 'admin/help#og':
       $path = drupal_get_path('module', 'og');
-      $output  = '<p>' . t("Read the <a href='@url'>README.txt</a> file in the Organic groups module directory.", array('@url' => "/$path/README.txt")) . '</p>';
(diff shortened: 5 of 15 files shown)


plugins/entityreference/selection/OgSelectionHandler.class.php
```diff
@@ -105,7 +105,6 @@ public function buildEntityFieldQuery($match = NULL, $match_operator = 'CONTAINS
 
     $field_mode = $this->instance['field_mode'];
     $user_groups = og_get_groups_by_user(NULL, $group_type);
-    $user_groups = $user_groups ? $user_groups : array();
```
amitaibu, line 108: this change seems wrong, or at least we should change `og_get_groups_by_user` to always return `[]`

https://github.com/Gizra/og/blob/81cd92d0651ed97e8d20627a9ebba161673ca6a8/og.module#L3566

amitaibu: Sorry, I see this is already the case in this PR.

amitaibu:
Thanks @alesrebec for the work! If you have also tested this locally - please go ahead and merge & create a release.

PR #393, opened 2018-06-26
Added note on contributing for D8

MPParsley:
Added links to issue queue, pr and alpha blockers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,13 @@
-Contributing
-------------
+# Contributing
+--------------
 
-If you like to contribute use the github issue queue to add issues
-and PRs
+If you like to contribute use [the github issue queue](https://github.com/gizra/og/issues) to add issues
+and pull requests.
+
+## Drupal 8
+
+If you want to help get OG released, please check the open issues for the Drupal 8 version and the open pull requests, especially the ones that are marked as alpha blockers.
+
+- [open issues](https://github.com/gizra/og/issues?q=is%3Aissue+is%3Aopen+label%3A%22Drupal+8%22)
+- [pull requests](https://github.com/Gizra/og/pulls?q=is%3Apr+is%3Aopen+label%3A"Drupal+8")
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,5 +1,13 @@
-Contributing
-------------
+# Contributing
+--------------
```
zerolab: No need for the underline here since you are using `#` to mark "Contributing" as heading.
Markdown renders the `-------` as an `<hr>` which looks odd imho

MPParsley: Well, I agree,  but that line was already there :-)

MPParsley:
@zerolab, there seems to be an issue with Travis.

Also, some not related question:
- "Note: Development and issues are handled on GitHub, and code is synced back to Drupal.org" => It seems that drupal.org/og Gizra/og have diverged, how does the sync happen?
- Are you on drupal.slack.com?

CONTRIBUTING.md
zerolab: The line was there because Markdown accepts both `#` and the underlines to mark headings - https://daringfireball.net/projects/markdown/syntax#header  - so either leave the underline or replace it wholly with the hashtag :)

zerolab:
phpcs complains about line lengths, so Travis is working fine.

I am on on drupal.slack.com as Dan Braghis

> Note: Development and issues are handled on GitHub, and code is synced back to Drupal.org
That is something that @amitaibu or @pfrenssen do once in a while. I see the last sync was about a month ago which is not bad.

MPParsley:
@zerolab thanks for the tips!

I updated the MD file to 80 lines but the links themselves are longer than 80 characters.
Either I shorten them using https://goo.gl/ or we need to exclude in code sniffer the links for markdown files.

MPParsley:
Appareantly Github has its own url shortener service, using that one :-)

See https://blog.github.com/2011-11-10-git-io-github-url-shortener/

amitaibu:
Great, thanks.

PR #395, opened 2018-06-27
Test on current versions of Drupal and PHP.

pfrenssen:
All our tests are running on deprecated versions of Drupal core. This PR updates the test matrix to the currently supported version 8.5.x and the upcoming 8.6.x.

I also enabled testing on PHP 7.2, curious how our code will run on this version.

To do:

- [x] PHPUnit 6 requires php ^7.0, use PHPUnit 4 when testing on PHP 5.6.
- [x] Error: Class 'Drupal\Console\Core\Style\DrupalStyle' not found.
- [x] ServiceNotFoundException: You have requested a non-existent service "entity_type.repository".
- [x] PrivateTempStoreFactory is scheduled for removal in Drupal 9.0.0. Use \Drupal\Core\TempStore\PrivateTempStoreFactory instead.
- [x] PrivateTempStore is scheduled for removal in Drupal 9.0.0. Use \Drupal\Core\TempStore\PrivateTempStore instead.
- [x] The "user.private_tempstore" service is deprecated. You should use the 'tempstore.private' service instead.
- [x] PHPUnit testing framework version 6 or greater is required when running on PHP 7.2 or greater.
- [x] The each() function is deprecated in PHP 7.2.
- [x] Fatal error:  Allowed memory size of 1610612736 bytes exhausted when installing Drupal console.
- [ ] Warning: call_user_func() expects parameter 1 to be a valid callback, class 'Drupal\dynamic_page_cache\EventSubscriber\DynamicPageCacheSubscriber' does not have a method 'onRouteMatch' -> related to https://www.drupal.org/node/2575841 ?
- [x] SafeMarkup::format() is scheduled for removal in Drupal 9.0.0. Use \Drupal\Component\Render\FormattableMarkup.
- [ ] The $memory_cache parameter was added in Drupal 8.6.x and will be required in 9.0.0.
- [ ] \Drupal\Component\Utility\Unicode::strlen() is deprecated in Drupal 8.6.0 and will be removed before Drupal 9.0.0. Use mb_strlen() instead.
- [x] \Drupal\Component\Utility\Unicode::strtolower() is deprecated in Drupal 8.6.0 and will be removed before Drupal 9.0.0. Use mb_strtolower() instead.
- [ ] The Drupal\Core\Entity\Query\QueryFactory class is deprecated in Drupal 8.3.0, will be removed before Drupal 9.0.0. Use \Drupal\Core\Entity\EntityStorageInterface::getQuery() or \Drupal\Core\Entity\EntityStorageInterface::getAggregateQuery() instead.
- [x] The "entity.query" service is deprecated. Use the 'entity_type.manager' service to get an entity type's storage object and then call \Drupal\Core\Entity\EntityStorageInterface::getQuery() or \Drupal\Core\Entity\EntityStorageInterface::getAggregateQuery() instead.

### DIFF
--- a/src/Form/GroupSubscribeForm.php
+++ b/src/Form/GroupSubscribeForm.php
@@ -42,8 +42,11 @@ class GroupSubscribeForm extends ContentEntityForm {
    * @param \Drupal\Component\Datetime\TimeInterface $time
    *   The time service.
    *
-   * @todo Set the correct type hint on the second argument when Drupal 8.6.0 is
-   *   released. It is currently omitted to preserve backwards compatibility.
+   * @todo Set the `EntityRepositoryInterface` type hint on the second argument
+   *   once Drupal 8.6.0 is released. It is currently omitted to preserve
+   *   backwards compatibility with Drupal 8.5.x and earlier.
+   *
+   * @see https://github.com/Gizra/og/issues/397
    */
   public function __construct(OgAccessInterface $og_access, $entity_repository, EntityTypeBundleInfoInterface $entity_type_bundle_info = NULL, TimeInterface $time = NULL) {
     parent::__construct($entity_repository, $entity_type_bundle_info, $time);
--- a/src/Form/GroupSubscribeForm.php
+++ b/src/Form/GroupSubscribeForm.php
@@ -2,10 +2,13 @@
 
 namespace Drupal\og\Form;
 
+use Drupal\Component\Datetime\TimeInterface;
 use Drupal\Core\Entity\ContentEntityForm;
+use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
-use Drupal\og\OgAccess;
+use Drupal\og\OgAccessInterface;
 use Drupal\og\OgMembershipInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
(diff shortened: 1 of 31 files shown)


pfrenssen:
A variety of failures :) I will take care of these during the sprint at Drupal Dev Days next week.

pfrenssen:
I brought this as far as I could. There are still some deprecated warnings being thrown on Drupal 8.6 which are not in our code base. There are also failures due to `DynamicPageCacheSubscriber::onRouteMatch()` which was removed in https://www.drupal.org/node/2575841 but this doesn't seem to be called by OG, and I was unable to replicate this bug locally.

Since all of these problems are occurring on Drupal 8.6.x which is still pre-alpha, I have marked them as being allowed to fail. We can have another look at these problems once Drupal 8.6 hits alpha.

src/Form/GroupSubscribeForm.php
```diff
@@ -23,23 +26,50 @@ class GroupSubscribeForm extends ContentEntityForm {
...
+   * @param \Drupal\og\OgAccessInterface $og_access
+   *   The OG access service.
+   * @param \Drupal\Core\Entity\EntityRepositoryInterface|\Drupal\Core\Entity\EntityManagerInterface $entity_repository
+   *   The entity repository service.
+   * @param \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info
+   *   The entity type bundle service.
+   * @param \Drupal\Component\Datetime\TimeInterface $time
+   *   The time service.
+   *
+   * @todo Set the correct type hint on the second argument when Drupal 8.6.0 is
```
amitaibu: if we'll do that, then it means OG will support only 8.6.0 - or you mean that 8.6.0 will be responsible for the BC?

amitaibu:
Just a question above, otherwise can be merged. Thanks!

src/Form/GroupSubscribeForm.php
pfrenssen: Yes, we have to leave out the type hint now to make it work for both 8.5 and 8.6. Once 8.5 is deprecated we don't need this B/C workaround any more and can add the `EntityRepositoryInterface` type hint. But in doing so it indeed means that OG will only support Drupal 8.6 from that point onwards, so this should be reflected in `composer.json`.

To clarify this I will create the followup issue to handle this task and link it here in the `@todo`.

Issue #372, opened 2018-05-28
Give some <3 to the GroupTypeManager


pfrenssen:
I was working with the `GroupTypeManager` today and encountered a few niggles. We can make a number of small improvements that will make this easier to understand and use:

- [x] ~~Provide a `GroupTypeManagerInterface` so that developers can override this service easily with a custom implementation, while still adhering to a well defined API.~~ This already exists!
- [x] Let's check if every time we refer to the group type manager in the code, we are referring to the `GroupTypeManagerInterface` and not to the class itself.
- [ ] The `::getAllGroupBundles()` method can be removed. It is incompletely documented and passing an argument to it completely changes the nature of the data that is returned. Moreover, if this method is called with no arguments, then it is an exact duplicate of `::getGroupMap()`, and if it is called with an argument it duplicates `::getGroupsForEntityType()`. So we can safely remove it after changing the calling code to the two other methods.
- [x] `::getGroupsForEntityType()` is badly named, and has incorrect documentation. This doesn't return any groups, but it returns group bundle IDs. So to be consistent with the sister method for group content, we should call this `::getGroupBundleIdsByEntityType()`.
- [x] `::getAllGroupContentBundles()` returns bundle IDs, but the method name implies that it returns objects of type `EntityTypeInterface`. Let's rename it to `::getAllGroupContentBundleIds()`.
- [ ] `::getAllGroupContentBundlesByEntityType()` is misleading in two ways: first of all it doesn't return "all" bundles, but only a subset, and secondly it returns bundle IDs and not `EntityTypeInterface` objects. So it can best be renamed to `::getGroupContentBundleIdsByEntityType()`. This will also be nicely consistent with the other methods.
- [x] We can update the generic `@return array` type hints to more precise data types like `@return string[]` and `@return string[][]`. This will help developers understand at a glance what they are dealing with.
- [ ] `::getGroupMap()` is very nondescriptive, it is not clear to a developer what we mean with the group map. So to be consistent with the other methods I would propose to rename this to `::getAllGroupBundleIds()`.
- [ ] `::isGroupContent()` is missing test coverage.
- [ ] `::getAllGroupContentBundleIds()` is missing test coverage.
- [ ] `::getAllGroupContentBundlesByEntityType()` is missing test coverage (or if it is already renamed, `::getGroupContentBundleIdsByEntityType()`).
- [ ] `::getGroupMap()` is missing test coverage (or if it is already renamed, `::getAllGroupBundleIds()`).
[...]

PR #402, opened 2018-07-03
Remove getAllGroupBundles and replace with getGroupMap

mariano-dagostino:
#372 

> The ::getAllGroupBundles() method can be removed. It is incompletely documented and passing an argument to it completely changes the nature of the data that is returned. Moreover, if this method is called with no arguments, then it is an exact duplicate of ::getGroupMap(), and if it is called with an argument it duplicates ::getGroupsForEntityType(). So we can safely remove it after changing the calling code to the two other methods.


src/Plugin/EntityReferenceSelection/OgSelection.php
```diff
@@ -68,13 +68,13 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
     $definition = \Drupal::entityTypeManager()->getDefinition($target_type);
 
     if ($bundle_key = $definition->getKey('bundle')) {
-      $bundles = Og::groupTypeManager()->getAllGroupBundles($target_type);
+      $map = Og::groupTypeManager()->getGroupMap();
```
pfrenssen: We can replace this with a call to `GroupTypeManager::getGroupsForEntityType()`, this can be passed the `$target_type` parameter directly, so we don't need to retrieve the full map and can save some memory.

pfrenssen (changes requested):
Looking good, just one small remark. Thanks!

pfrenssen:
Oh, something went wrong in the merge here, the `getAllGroupBundles()` method has reappeared in the `GroupTypeManager`, and in the `GroupTypeManagerInterface`. Can you remove them again?

pfrenssen:
Now we have some failures due to `::getGroupsForEntityType()` being renamed to `::getGroupBundleIdsByEntityType()`. This happened in #404 which was merged recently.

We'll get there eventually :grimacing:

src/Plugin/EntityReferenceSelection/OgSelection.php
```diff
@@ -70,7 +70,7 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
     $definition = \Drupal::entityTypeManager()->getDefinition($target_type);
 
     if ($bundle_key = $definition->getKey('bundle')) {
-      $bundles = Og::groupTypeManager()->getAllGroupBundles($target_type);
+      $bundles = Og::groupTypeManager()->getGroupsForEntityType($target_type);
```
MPParsley: In #404 ::getGroupsForEntityType() was renamed to ::getGroupBundleIdsByEntityType():

```suggestion
      $bundles = Og::groupTypeManager()->getGroupBundleIdsByEntityType($target_type);
```

src/GroupTypeManagerInterface.php
```diff
@@ -50,7 +50,7 @@ public function getGroupBundleIdsByEntityType($entity_type_id);
   /**
    * Get all group bundles keyed by entity type.
    *
-   * @return string[][]
+   * @return array
```
MPParsley: Revert return type
```suggestion
   * @return string[][]
```

pfrenssen (approved):

PR #403, opened 2018-07-03
Make referenec to GroupTypeManagerInterface and not the class.

mariano-dagostino:
#372 

> Let's check if every time we refer to the group type manager in the code, we are referring to the GroupTypeManagerInterface and not to the class itself.

I searched for `egrep -unr 'GroupTypeManager[^I]' .` to see what needed to be changed,


pfrenssen (approved):
Looking good, thanks!

PR #404, opened 2018-07-03
Rename getGroupsForEntityType into getGroupBundleIdsByEntityType

mariano-dagostino:
#372 

> ::getGroupsForEntityType() is badly named, and has incorrect documentation. This doesn't return any groups, but it returns group bundle IDs. So to be consistent with the sister method for group content, we should call this ::getGroupBundleIdsByEntityType().

### DIFF
--- a/tests/src/Unit/GroupTypeManagerTest.php
+++ b/tests/src/Unit/GroupTypeManagerTest.php
@@ -204,7 +204,7 @@ class GroupTypeManagerTest extends UnitTestCase {
   /**
    * Tests getting all the groups of an entity type.
    *
-   * @covers ::getGroupsForEntityType
+   * @covers ::getGroupBundleIdsByEntityType
    */
   public function testGetGroupsForEntityType() {
     // It is expected that the group map will be retrieved from config.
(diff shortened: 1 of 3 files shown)


tests/src/Unit/GroupTypeManagerTest.php
```diff
@@ -204,7 +204,7 @@ public function providerTestIsGroup() {
   /**
    * Tests getting all the groups of an entity type.
    *
-   * @covers ::getGroupsForEntityType
+   * @covers ::getGroupBundleIdsByEntityType
    */
   public function testGetGroupsForEntityType() {
```
pfrenssen: We can now also rename the test to `testGetGroupBundleIdsByEntityType()`. The documentation of this test can also be updated, it says that we are "getting all the groups" but we are in fact getting all the group bundle IDs.

pfrenssen (changes requested):
Looking good, just for completion a test method can be renamed and its documentation updated.

Thanks for working on this!

pfrenssen (approved):
Great, thanks, looking good!

PR #405, opened 2018-07-03
Rename getAllGroupContentBundles into getAllGroupContentBundleIds

mariano-dagostino:
#372

> ::getAllGroupContentBundles() returns bundle IDs, but the method name implies that it returns objects of type EntityTypeInterface. Let's rename it to ::getAllGroupContentBundleIds().


pfrenssen (approved):
Looks great, thanks!

PR #408, opened 2018-07-04
We can update the generic @return array type hints to more precise da…

fazni:
…ta types like @return string[] and @return string[][]. This will help developers understand at a glance what they are dealing with.

### DIFF
--- a/src/GroupTypeManagerInterface.php
+++ b/src/GroupTypeManagerInterface.php
@@ -50,7 +50,7 @@ interface GroupTypeManagerInterface {
   /**
    * Get all group bundles keyed by entity type.
    *
-   * @return array
+   * @return string[]
    *   An associative array, keyed by entity type, each value an indexed array
    *   of bundle IDs.
    */
@@ -63,7 +63,7 @@ interface GroupTypeManagerInterface {
    * information about the relations between groups and group content bundles
    * then use getGroupRelationMap() instead.
    *
-   * @return array
+   * @return string[]
    *   An associative array of group content bundle IDs, keyed by entity type
    *   ID.
    *
@@ -81,7 +81,7 @@ interface GroupTypeManagerInterface {
    * @param string $entity_type_id
    *   Entity type ID to filter the bundles by.
    *
-   * @return array
+   * @return string[]
    *   An array of group content bundle IDs.
    *
    * @throws \InvalidArgumentException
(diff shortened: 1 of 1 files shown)


pfrenssen:
This is fixing one of the points of #372.

src/GroupTypeManagerInterface.php
```diff
@@ -50,7 +50,7 @@ public function getGroupsForEntityType($entity_type_id);
   /**
    * Get all group bundles keyed by entity type.
    *
-   * @return array
+   * @return string[]
```
pfrenssen: This is a nested array, so the type hint should be `string[][]`.

src/GroupTypeManagerInterface.php
```diff
@@ -164,7 +164,7 @@ public function resetGroupRelationMap();
   /**
    * Returns the group map.
    *
-   * @return array
+   * @return string[]
```
pfrenssen: This array has this format:

```
[
  $entity_type_id => [
    $bundle_ids,
  ],
]
```

So the correct return type is `string[][]`.

src/GroupTypeManagerInterface.php
```diff
@@ -63,7 +63,7 @@ public function getAllGroupBundles($entity_type = NULL);
    * information about the relations between groups and group content bundles
    * then use getGroupRelationMap() instead.
    *
-   * @return array
+   * @return string[]
```
pfrenssen: This is also a nested array, so this one is also `string[][]`.

src/GroupTypeManagerInterface.php
```diff
@@ -102,7 +102,7 @@ public function getAllGroupContentBundlesByEntityType($entity_type_id);
    *   The bundle ID of the group content type for which to return associated
    *   group bundle IDs.
    *
-   * @return array
+   * @return string[]
```
pfrenssen: This is a nested array which is keyed by entity type ID, so this is a `string[][]`.

src/GroupTypeManagerInterface.php
```diff
@@ -117,7 +117,7 @@ public function getGroupBundleIdsByGroupContentBundle($group_content_entity_type
    *   The bundle ID of the group type for which to return associated group
    *   content bundle IDs.
    *
-   * @return array
+   * @return string[]
```
pfrenssen: Same story here, it is a `string[][]`.

pfrenssen (changes requested):
Looks good, but in a few cases a nested array is returned. If these are fixed this is OK to merge.

pfrenssen (approved):
Excellent, thanks!

Issue #352, opened 2018-01-15
Fix for being able to relate content from the logged in user ( drupal 8.3 )


beerbaitandfishing:
Add the following to the file - modules/contrib/og/og.views.inc

-----Add ------

$data['og_membership']['node']['relationship'] = [
    'real field' => 'entity_id',
    'base' => 'node',
    'base field' => 'nid',
    'label' => t('OG Membership Content'),
    'title' => t('OG Membership Content'),
    'id' => 'standard',
  ];

Add that under the existing relation between user and og_membership.

Then you should be able to go from user -> og membership -> content ( the group that they're a member of ).

MPParsley:
Hi @beerbaitandfishing, I created a PR in #415.

This relation might be useful but isn't this issue a duplicate for #254? You'd want to use a contextual filter to list all the content for the current user.

meecect:
I thought #254 was more about getting content from groups that a user is a part of.  I think this issue is just about being about to get the group node from the og membership entity.

I found this issue while I was trying to create a view that showed the groups a user belonged to.  The familiar relationship from the the 7.x branch aren't here anymore so I'm a bit lost as to how to create the view. 

When I add this  code, I do see the new 'OG Membership Content'  Relation and I have added it, however, I do not get many (any?) additional fields to add to my view.  I was expecting to see things like 'title' and the rest of the fields on the group node bundle that I could specify to use the relationship, but they aren't there.  

I can chose 'rendered entity' but then I have to choose a view mode.  I think it would be useful to just render individual fields.

meecect:
anyway, for my case, I thought I might as well add another relation, relating the group to the og membership, then the membership to the user, then filtering on the user.  I did that by adding this:

  $data['node']['og_membership']['relationship'] = [
    'real field' => 'nid',
    'base' => 'og_membership',
    'base field' => 'entity_id',
    'label' => t('OG Membership'),
    'title' => t('OG Membership'),
    'id' => 'standard',
  ];



It seems to work well.

semiaddict:
I believe this would indeed be a better and more generic option than the default argument proposed in #254.

cofihouse:
[...]

PR #415, opened 2018-07-23
Added relation to content from the logged in user.

MPParsley:
PR for patch in #352.


pfrenssen:
This still needs a test and an example use case.

PR #420, opened 2018-08-03
Performance: Store the group bundle in the membership entity

pfrenssen:
We got a report from a user that was experiencing very slow page loads throughout the entire website. This user is very active and is a member of 547 groups. She was experiencing page loading times of around 15+ seconds throughout the entire website.

After profiling it turned out that over 95% of the process time was spent in this line of code in `OgMembership`:

```
  public function getRoles() {
    $roles = [];

    // Add the member role. This is only possible if a group has been set on the
    // membership.
    if ($group = $this->getGroup()) { // <<< 95% of page load is spent executing this line.
      $roles = [
        OgRole::getRole($this->getGroupEntityType(), $group->bundle(), OgRoleInterface::AUTHENTICATED),
      ];
    }
    $roles = array_merge($roles, $this->get('roles')->referencedEntities());
    return $roles;
  }
```

This code gets executed through `OgRoleCacheContext::getContext()` which loads all the user memberships, and calls `::getRoles()` on them to compile a hash that identifies this user's memberships. Through this method call this was also doing full entity loads of all the groups this user is a member of.

In our projects we have two group types and both are very complex entities with many fields, and it turns out that loading 547 of them has a huge impact on the page load.

Now it is not even necessary for us to dynamically load the full group entity to be able to determine the roles of the membership. We only need to know the bundle.

This PR stores the group bundle in the membership table, so it sits along the entity type and ID which we were already storing. By doing this we can avoid the full entity loads.

The result is a huge performance increase. Here are the results from loading a page containing a simple configuration form for a user that is a member of 547 groups:

* without this PR: 15.56 seconds
* with this PR: 0.21 seconds (74x faster!)

### DIFF
--- a/og.install
+++ b/og.install
@@ -5,10 +5,60 @@
  * Install, update, and uninstall functions for the Organic groups module.
  */
 
+use Drupal\Core\Field\BaseFieldDefinition;
+
 /**
  * Implements hook_uninstall().
  */
 function og_uninstall() {
   \Drupal::queue('og_orphaned_group_content')->deleteQueue();
   \Drupal::queue('og_orphaned_group_content_cron')->deleteQueue();
+}
+
+/**
+ * Add a base field to store the group bundle in memberships.
+ */
+function og_update_8001(&$sandbox) {
+  $storage = \Drupal::entityTypeManager()->getStorage('og_membership');
+
+  if (!isset($sandbox['total'])) {
+    $storage_definition = BaseFieldDefinition::create('string')
+      ->setLabel(t('Group bundle id'))
+      ->setDescription(t("The bundle ID of the group."));
+    \Drupal::entityDefinitionUpdateManager()->installFieldStorageDefinition('entity_bundle', 'og_membership', 'og', $storage_definition);
+
+    $sandbox['#finished'] = 0;
+    $sandbox['current'] = 0;
+    $sandbox['total'] = $storage->getQuery()->count()->execute();
+    $sandbox['batch_size'] = 500;
+  }
+
+  // Update the existing memberships to include the group bundle ID.
+  $membership_ids = $storage->getQuery()
+    ->range($sandbox['current'], $sandbox['batch_size'])
+    ->sort('id')
+    ->execute();
+
+  /** @var \Drupal\og\Entity\OgMembership $membership */
+  foreach ($storage->loadMultiple($membership_ids) as $membership) {
+    $group = $membership->getGroup();
+    if (!empty($group)) {
+      $membership->set('entity_bundle', $group->bundle());
+      $membership->save();
+    }
+  }
+
+  $sandbox['current'] += $sandbox['batch_size'];
+  if ($sandbox['current'] >= $sandbox['total']) {
+    $sandbox['current'] = $sandbox['total'];
+  }
+  $sandbox['#finished'] = $sandbox['current'] / $sandbox['total'];
+
+  $message = t('Processed @current of @total memberships (@percentage% complete)', [
+    '@current' => $sandbox['current'],
+    '@total' => $sandbox['total'],
+    '@percentage' => number_format($sandbox['#finished'] * 100, 2),
+  ]);
+
+  return $message;
 }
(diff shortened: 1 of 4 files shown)


idimopoulos:
This has indeed boosted up performance quite a few times up. The tests are covering it fine though I am not sure about the new Interface method breaking BC. However, this is a very good feature.

og.install
```diff
@@ -5,10 +5,60 @@
...
+
+/**
+ * Add a base field to store the group bundle in memberships.
+ */
+function og_update_8001(&$sandbox) {
+  $storage = \Drupal::entityTypeManager()->getStorage('og_membership');
+
+  if (!isset($sandbox['total'])) {
+    $storage_definition = BaseFieldDefinition::create('string')
+      ->setLabel(t('Group bundle id'))
```
amitaibu: => Group bundle ID

og.install
```diff
@@ -5,10 +5,60 @@
...
+      ->setDescription(t("The bundle ID of the group."));
```
amitaibu: `"` => `'`

pfrenssen:
Thanks for the review! We should be fine regarding B/C, it's just a new method being added to the interface. We're also not in beta yet so it is still allowed to change stuff.

og.install
```diff
@@ -5,10 +5,60 @@
...
+  // Update the existing memberships to include the group bundle ID.
+  $membership_ids = $storage->getQuery()
+    ->range($sandbox['current'], $sandbox['batch_size'])
+    ->sort('id')
+    ->execute();
+
+  /** @var \Drupal\og\Entity\OgMembership $membership */
+  foreach ($storage->loadMultiple($membership_ids) as $membership) {
+    $group = $membership->getGroup();
+    if (!empty($group)) {
```
amitaibu, line 45: As part of the update, we should probably already have a group (as we're not creating a new membership). Anyway, we can keep as is.

src/Entity/OgMembership.php
```diff
@@ -137,13 +138,35 @@ public function getGroupEntityType() {
...
     return $this->get('entity_id')->value;
   }
 
+  /**
+   * Checks if a group has already been populated on the membership.
+   *
+   * The group is required for a membership, so it is always present if a
+   * membership has been saved. This is intended for internal use to verify if
+   * a group is present when methods are called on a membership that is possibly
+   * still under construction.
```
amitaibu, line 161: Maybe add a comment about performance, something like. ```In order to avoid loading the group just for this purpose, we access the `entity_type` property that should give us the same indication, but with a lower performance cost.```

src/Entity/OgMembership.php
```diff
@@ -335,6 +358,10 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setLabel(t('Group entity type'))
       ->setDescription(t('The entity type of the group.'));
 
+    $fields['entity_bundle'] = BaseFieldDefinition::create('string')
+      ->setLabel(t('Group bundle id'))
```
amitaibu: id => ID

amitaibu (commented):
Looks good, and wonderful to see such a performance improvement. Some nitpick comments, and then you can go ahead an merge

og.install
pfrenssen: Very good comment. I was not sure about this one. I think there are cases where there might be orphaned data left when the update hook is executed. For example if the project has configured orphans to be cleaned up on cron. I don't want the update to fail on potential missing groups, since that could be a very stressful situation for a devops engineer during a deployment.

I think it might probably be a good idea to remove the membership at this point, since a membership for a group that no longer exists is orphaned and can safely be removed. What do you think?

This comment also makes me realize there is a potential bug in the protected `OgMembership::hasGroup()` method that is being added here. If this update hook finds that the group no longer exists, then it will not update the membership, and the `entity_bundle` field will contain `NULL`. This will then blow up in `OgMembership::getRoles()`, since it will call `::hasGroup()` which doesn't currently check the presence of the bundle, and then it will try to retrieve a non-existing role.

I will address both: I will clean up the orphaned memberships as part of the update hook, and will update `::hasGroup()` to also check if the bundle is populated.

src/Entity/OgMembership.php
pfrenssen: I also fixed another instance here where `ID` was incorrectly capitalized, and used single quotes instead of double quotes.

amitaibu:
Thanks!

PR #424, opened 2018-08-15
Avoid database lock timeouts when cleaning up orphans in long running Behat tests

pfrenssen:
I'm working on a large project that is based on OG and uses Behat for validating user scenarios. Our current Behat test suite has grown to 9000+ steps in 500+ scenarios and takes around 90 minutes to complete.

Lately we are experiencing around a 20% failure rate in our suite due to database lock timeouts occurring when OG is cleaning up orphaned group content. We are using the default "simple" plugin for cleaning up orphans which works fine in a normal Drupal page request, but unfortunately during this 1.5 hour long test the "page request" remains active and accumulates a large number of orphans. Cleaning these up takes so long that database lock timeouts occur.

Here is an example of such a failing test: https://app.continuousphp.com/git-hub/ec-europa/joinup-dev/build/ea8af49d-8c49-4ff5-ba79-8c451bd4aa5c/logs/5945b8bb-d395-4aa1-b081-6046c9e2c29b#line-1678

This PR implements the `@AfterScenario` Behat hook and cleans up orphans after every test scenario rather than after the entire test suite.

### DIFF
--- a/og.behat.inc
+++ b/og.behat.inc
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * @file
+ * Contains \OgSubContext.
+ */
+
+use Drupal\DrupalExtension\Context\DrupalSubContextBase;
+use Drupal\DrupalExtension\Context\DrupalSubContextInterface;
+use Drupal\og\OgDeleteOrphansInterface;
+
+/**
+ * Behat step definitions and hook implementations for Organic Groups.
+ */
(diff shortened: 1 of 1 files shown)


pfrenssen:
This is not yet ready for review. I first would like to run a number of tests against our test suite to see if this indeed fixes the problem. Our current failure rate is 4/15 tests. I plan on running 15 tests which will take several days to complete.

pfrenssen:
We had another failure, this doesn't seem to be the right solution. Closing.

PR #428, opened 2018-09-16
7.x 2.x og fix head tests for various versions of php

joejoseph00:
Currently head on drupal.org is passing only on php 5.3.x but is failing on everything else.  The fix is very simple.  I am raising this issue because the next release of core (7.x-dev already supports it) will support php 7.2.x .  I've been running tests against contrib modules that I maintain and others, so far it's looking great.  For og (organic groups) we only need to make this small change.
For followup on drupal.org please see https://www.drupal.org/project/og/issues/3000030

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/Gizra/og.svg?branch=7.x-2.x)](https://travis-ci.org/Gizra/og)
+[![Build Status](https://travis-ci.org/Gizra/og.svg?branch=7.x-2.x)](https://travis-ci.org/Gizra/og/branches)
 
 DESCRIPTION
 --------------------------
--- a/og.test
+++ b/og.test
@@ -2166,7 +2166,10 @@ class OgNonMembersPublishingContentTestCase extends DrupalWebTestCase {
 
     $this->assertTrue(!empty($result['node']), 'The node was added to the group.');
 
-    $nid = reset(array_keys($result['node']));
+    // Avoid strict errors in php 5.4, 5.5, 5.6, 7.0, 7.1, 7.2, see https://drupal.org/node/3000030 patch 2.
+    $result_node = $result['node'];
+    $node_array_keys = array_keys($result_node);
+    $nid = reset($node_array_keys);
 
     $this->drupalLogin($this->adminUser);
 
--- a/og.test
+++ b/og.test
@@ -2166,7 +2166,6 @@ class OgNonMembersPublishingContentTestCase extends DrupalWebTestCase {
 
     $this->assertTrue(!empty($result['node']), 'The node was added to the group.');
 
-    // Avoid strict errors in php 5.4, 5.5, 5.6, 7.0, 7.1, 7.2, see https://drupal.org/node/3000030 patch 2.
     $result_node = $result['node'];
     $node_array_keys = array_keys($result_node);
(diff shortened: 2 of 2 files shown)


joejoseph00:
This is to fix:

> 
> exception: [Run-time notice] Line 2169 of sites/all/modules/og/og.test:
> Only variables should be passed by reference

as seen in these test results:
https://www.drupal.org/project/og/issues/2299353#comment-11809313
AND
https://www.drupal.org/project/og/issues/2998637#comment-12776149

joejoseph00:
Woohoo! passes, more importantly however, it passes drupal.org simpletest automation for organic groups.  See the passing results here:
https://www.drupal.org/project/og/issues/3000030#comment-12776181

for the before patch I can show you those if you need more proof.  When this is merged please also update the branch commit and issue branch tests to add tests for php 5.6, php 7.0, php 7.1 and php 7.2 on drupal.org , here is the link to do that: https://www.drupal.org/node/13446/qa only maintainers have access to add branch tests and commit tests.

og.test
```diff
@@ -2166,7 +2166,10 @@ class OgNonMembersPublishingContentTestCase extends DrupalWebTestCase {
 
     $this->assertTrue(!empty($result['node']), 'The node was added to the group.');
 
-    $nid = reset(array_keys($result['node']));
+    // Avoid strict errors in php 5.4, 5.5, 5.6, 7.0, 7.1, 7.2, see https://drupal.org/node/3000030 patch 2.
```
amitaibu: This comment can be removed.

README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/Gizra/og.svg?branch=7.x-2.x)](https://travis-ci.org/Gizra/og)
+[![Build Status](https://travis-ci.org/Gizra/og.svg?branch=7.x-2.x)](https://travis-ci.org/Gizra/og/branches)
```
amitaibu, line 1: Why is this change needed?

amitaibu (changes requested):
@joejoseph00 Thank you for the PR! 

@alesrebec care to take it from here? 😄

README.md
joejoseph00: This change is needed because on the 7.x-2.x branch clicking that link takes us to the 8.x build results.  It should go to the 7.x-2.x build results but all I could find is /branches which at least shows all the build results for all the branches (which is better than just the latest default branch result which is what is used prior to this change)

joejoseph00: However this change isn't critical to the pull request, only the change to og.test is critical.  This change is merely an unrelated small improvement imho.  It's important that the og.test gets fixed so I'm willing to drop this change to the README.md if you disagree with it.

joejoseph00:
@alesrebec , @amitaibu , 
would be really nice to get this in asap so that automated tests work for php 5.4.x , 5.5.x, 5.6.x , 7.0.x, 7.1.x and 7.2.x
see the patch test results on drupal.org, it fixes the test so that it passes on all of these newer versions!
The default should be set to php 7.2.x
https://www.drupal.org/project/og/issues/3000030#comment-12776181
Drupal core the next release will support php 7.2.x , lets be proactive here and push this in to be ready!

Thanks

alesrebec:
Looking good. Merged into 7.x-2.x now.
Thanks @joejoseph00 .

PR #429, opened 2018-10-08
Deleted groups causes EntityMalformedException in og_is_group().

MegaChriz:
When a group is deleted, other entities can still hold a reference to this group. For example: group content. This can cause exceptions of type EntityMalformedException in various situations. One of them - which is addressed here - is when you have a single group content assigned to multiple groups and then one of the groups is deleted. When you then try to update the group content an EntityMalformedException is thrown.

Steps to reproduce:

1. Enable modules og, og_ui, og_access.
2. Create a node type used as group.
3. Create a node type used as group content.
4. Create a role with permissions to create, edit and delete own group and group content nodes.
5. Log in as an user with this role.
6. Create two group nodes.
7. Create group content node, assign to both groups.
8. Delete the first created group.
9. Go to the node edit page of the group content.
10. Click "Save".

An other way to reproduce the issue:

1. Enable modules og, og_ui, og_access.
2. On the OG settings page (/admin/config/group/settings), enable the "Use queue" option.
3. Create a node type used as group.
4. Create a node type used as group content.
5. Create a role with permissions to create, edit and delete own group and group content nodes.
6. Log in as an user with this role.
7. Create two group nodes.
8. Create group content node, assign to both groups.
9. Delete the first created group.
10. Do something that checks if the current user may update the group content, using a node_access() call. For example a View with your own nodes that includes a Views edit link (views_handler_field_node_link_edit).

See also this issue on drupal.org to see that the included tests fail without the fix:
https://www.drupal.org/project/og/issues/2900273

### DIFF
--- a/og.module
+++ b/og.module
@@ -1830,6 +1830,10 @@ function og_is_group($entity_type, $entity) {
   if (is_numeric($entity)) {
     $entity = entity_load_single($entity_type, $entity);
   }
+  if (!$entity) {
+    // The entity does not exist.
+    return FALSE;
+  }
   list(,, $bundle) = entity_extract_ids($entity_type, $entity);
   if (!field_info_instance($entity_type, OG_GROUP_FIELD, $bundle)) {
     return variable_get("og_is_group__{$entity_type}__{$bundle}", FALSE);
--- a/og.test
+++ b/og.test
@@ -2083,6 +2083,82 @@ class OgDeleteOrphansTestCase extends DrupalWebTestCase {
 
     $this->assertEqual($gid, $second_group->nid, 'The group content moved to another group.');
   }
+
+  /**
+   * Tests editing content that belongs to a deleted group.
+   *
+   * Tests if content belonging to multiple groups can be edited without issues
+   * after deleting one of the groups.
+   */
+  public function testEditGroupContentAfterDeletingGroup() {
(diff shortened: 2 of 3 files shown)


jannakha:
Please, merge it in and release a patch :-) the fix works for me 💯 
thanks for the fix!!! 👍

jannakha:
Please, merge it in and release a patch :-) the fix works for me 💯 
thanks for the fix!!! 👍

jannakha:
Please, merge it in and release a patch :-) the fix works for me 💯 
thanks for the fix!!! 👍

jacobembree:
Nice work, @MegaChriz.

alesrebec (approved):
Review the code: ✔
> Clean as a whistle.

Do the steps to reproduce the bug: ✔
> group content edit after one of two groups is deleted

Apply the patch and check if the bug is fixed: ✔
> Bug is gone after patch is applied.

Run automated tests: ✔
> All green! 
screenshot from 2019-01-13 18-56-54

alesrebec:
All looking good after the review.
Merged to 7.x-2.x.

Thanks @MegaChriz.

jacobembree:
This commit triggered a build that failed on PHP 7.3, but I suspect that this commit itself is okay. It's passing on PHP 5.6.  I tested on PHP 7.3, and that passed.  Tests passed for @alesrebec.

PR #430, opened 2018-10-09
Add an index to improve speed of og_roles()

jorrit:
`og_roles()` queries the og_role table with conditions on the group_type, group_bundle and gid columns. None of these columns has an index at the moment.

This pull request adds an index to speed up `og_roles()`.

### DIFF
--- a/og.install
+++ b/og.install
@@ -169,6 +169,9 @@ function og_schema() {
       ),
     ),
     'primary key' => array('rid'),
+    'indexes' => array(
+      'lookup' => array('group_type', 'group_bundle', 'gid'),
+    ),
   );
 
   $schema['og_users_roles'] = array(
@@ -1186,4 +1189,11 @@ function og_update_7204() {
  */
 function og_update_7205() {
   db_add_index('og_membership', 'group_type', array('group_type'));
+}
+
+/**
+ * Add an index to {og_role} on group_type, group_bundle and gid to improve speed of og_roles().
+ */
+function og_update_7206() {
+  db_add_index('og_role', 'lookup', array('group_type', 'group_bundle', 'gid'));
 }


amitaibu:
Cool, thanks. Do you happen to have some stats to back this PR?

MPParsley (commented):
We should probably use the new array notation [] instead of array().

amitaibu:
As this Pr is for 7.x, we still have the old coding standard of `array()`

MPParsley:
Whoops, my bad. I just realized it was for 7.x

jorrit:
What kind of stats do you want? Query execution time or improvement in the execution plan? The execution time greatly depends on the amount of roles in og_roles and the amount of work the SQL server is also doing at that time. In the execution plan you can see that it changes from 839 rows examined (all rows) with "Using where" to 3 rows examined (the only rows needed) with "Using index condition; Using where". Examining less rows is always faster. The query is a simple one with 3 constant conditions. With this index, these conditions can be read entirely from the index.

If you want query timing statistics I can try to provide some with a lot of caveats.

amitaibu:
> If you want query timing statistics I can try to provide some with a lot of caveats.

That would be great, For sure it will have caveats, but I think it's still important to see if the "effort" is worth it, and a extra step to make sure we indeed improve the speed.

jorrit:
I have 862 rows in `og_role`.

Test script which I executed via /devel/php:

```php
function test($gid) {
  $t = microtime(true);
  for($i = 0; $i < 1000; $i++) {
    og_roles('node', 'oa_space', $gid);
  }
  dpm(microtime(true) - $t);
}

db_drop_index('og_role', 'lookup');
test(0);
test(8);
test(3863);

db_add_index('og_role', 'lookup', array('group_type', 'group_bundle', 'gid'));
test(0);
test(8);
test(3863);
```

My output:

```
1.2727510929108
1.6513860225677
1.5563130378723
0.55390405654907
0.89380192756653
0.92062616348267
```

The first three lines are without the index, the last three are with the index.

Of course, the savings per query execution are minuscule, but that is the case for every index. My goal is to reduce the amount of full table scans where the amount of rows considered is much larger than the amount of rows returned, and this index, among others, helps with that.

Please note that the MySQL query cache is disabled, otherwise calling the same query right after the same query has executed before is very fast. In practice, `og_role()` will not be called immediately after another and it will be called with many different arguments. Also, using the query cache is usually discouraged because cache management adds overhead and in a highly dynamic site like Drupal the probability that a query already present in the cache is quite small.

amitaibu:
Thanks @jorrit . @alesrebec if looks ok to you, could you please merge?

alesrebec:
Nice find @jorrit!
Looks good and merged into 7.x-2.x now.

jorrit:
Thanks @amitaibu  and @alesrebec !

PR #437, opened 2018-11-19
Fix Travis CI integration

webflo:
Hi, i tried to fix the Travis CI Integration. You can see the debugging process in my commits. The additional ``composer install`` inside the og Repo caused some issue. The class loader in this folder is picked up by the test-runner oder Drupal?

It was not problem in the past, because the code inside the module and /drupal/vendor was similar. But the diff between multiple versions of Drupal gets bigger over time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,4 @@ before_script:
   # Export database variable for kernel tests.
   - export SIMPLETEST_DB=mysql://root:@127.0.0.1/og
 
-  # Disable all deprecations
-  - export SYMFONY_DEPRECATIONS_HELPER=disabled
-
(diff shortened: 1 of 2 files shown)


webflo:
It passed with all deprecations disabled. see https://travis-ci.org/ueberbit/og/builds/456878838

webflo:
Should we add SYMFONY_DEPRECATIONS_HELPER=disabled? And open an second PR to get rid of all deprecations?

// ping @pfrenssen

pfrenssen:
Oh I didn't see this, I have just updated the test matrix and solved those deprecations in #447.

Getting rid of that call to `composer install` would be great! Care to merge in latest 8.x-1.x and seeing if it turns green?

webflo:
@pfrenssen done

pfrenssen:
Thanks!

PR #444, opened 2018-12-11
Add alter hook to modify whether the 'Other groups' field is visible

jorrit:
I want to show the 'Other groups' field in a specific case. Adding an alter hook gives me the possibility to override the default behavior.

### DIFF
--- a/includes/og.field.inc
+++ b/includes/og.field.inc
@@ -50,15 +50,21 @@ function og_field_widget_form(&$form, &$form_state, $field, $instance, $langcode
   ctools_include('fields');
 
   $field_modes = array('default');
-  $has_admin = FALSE;
+  $has_admin = user_access('administer group');
+  $context = array(
+    'field' => $field,
+    'instance' => $instance,
+    'entity_type' => $entity_type,
+    'entity' => $entity,
+  );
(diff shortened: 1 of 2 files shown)


amitaibu:
Thanks. What about using https://api.drupal.org/api/drupal/modules!field!field.api.php/function/hook_field_widget_form_alter/7.x instead?

jorrit:
Thanks for your response. Can you explain how I can use that hook to implement what I need? I want to override the value of `$has_admin`.

amitaibu:
Maybe from the alter hook, you can call a custom

`my_og_field_widget_form` which is a copy/ paste of `og_field_widget_form`, but with your custom logic.

What that work for you?

jorrit:
That would work, but the downside is that I would need to keep my code in sync with yours in order to change just a little bit of functionality. Anyway, I am going to keep this as a patch in my make file for now. I understand it when you are reluctant to add any more hooks to og.

amitaibu:
> I would need to keep my code in sync with yours in order to change just a little bit of functionality

Theoretically yes, but I suspect OG7 isn't going to have too many changes. I'd be ok with a PR to moved the logic into API functions you could use, to reduce the copy/ paste.

claudiu-cristea:

PR #446, opened 2018-12-11
Avoid double caching of OgMembership entities.

### LINKED ISSUE #445: Membership entities are double cached in `MembershipManager`
`MembershipManager::getMemberships()` caches the fully loaded `OgMembership` entities, but this is pointless since they are already cached in `EntityStorageBase::loadMultiple()`.

Let's save some precious memory by only caching the entity IDs instead of the full entities.

pfrenssen:
Fixes #445

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -97,25 +97,20 @@ class MembershipManager implements MembershipManagerInterface {
     ];
     $identifier = implode(':', $identifier);
 
-    // Return cached result if it exists.
-    if (isset($this->cache[$identifier])) {
-      return $this->cache[$identifier];
+    // Use cached result if it exists.
+    if (!isset($this->cache[$identifier])) {
+      $query = $this->entityTypeManager
+        ->getStorage('og_membership')
+        ->getQuery()
+        ->condition('uid', $user->id())
+        ->condition('state', $states, 'IN');
+
+      $this->cache[$identifier] = $query->execute();
     }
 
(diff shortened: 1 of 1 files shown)


pfrenssen:
Rebased on top of #447 since the tests were failing due to Drupal 8.6 incompatibility which is fixed in that PR.

src/MembershipManager.php
```diff
@@ -97,25 +97,20 @@ public function getMemberships(AccountInterface $user, array $states = [OgMember
...
+    }
 
-    /** @var \Drupal\og\Entity\OgMembership[] $memberships */
-    $this->cache[$identifier] = $this->entityTypeManager
+    return $this->entityTypeManager
       ->getStorage('og_membership')
-      ->loadMultiple($results);
-
-    return $this->cache[$identifier];
+      ->loadMultiple($this->cache[$identifier]);
```
amitaibu, line 113: `$this->cache[$identifier]` can also be an empty array at this point, no?

pfrenssen: Yes good point. In the code before this PR this was also happening, but I will still put a switch in so we don't needlessly try to do a load.

pfrenssen: I fixed it here in #451 which already refactors this line: https://github.com/Gizra/og/pull/451/commits/84e5553ebc97a7e5763df52076f3ddc729a35345

This saves me a round of rebasing :)

amitaibu:
Thanks.

Issue #400, opened 2018-07-03
Ensure compatibility with Drupal 8.6.0


pfrenssen:
We are dealing with a number of test failures on the Drupal 8.6.x development branch. These need to be solved before Drupal 8.6.0 is released.

- [ ] `Exception: Warning: call_user_func() expects parameter 1 to be a valid callback, class 'Drupal\dynamic_page_cache\EventSubscriber\DynamicPageCacheSubscriber' does not have a method 'onRouteMatch'`. This seems to be related to [Issue #2575841: Use proper method name in DynamicPageCacheSubscriber](https://www.drupal.org/node/2575841) in which this method was removed.
- [ ] `SafeMarkup::format() is scheduled for removal in Drupal 9.0.0. Use \Drupal\Component\Render\FormattableMarkup.`. All instances in OG have been removed, but this is deprecation warning is still thrown in other code. Possibly 
- [ ] `The $memory_cache parameter was added in Drupal 8.6.x and will be required in 9.0.0`. This is thrown in `EntityStorageBase::__construct()`.
- [ ] `\Drupal\Component\Utility\Unicode::strlen() is deprecated in Drupal 8.6.0 and will be removed before Drupal 9.0.0. Use mb_strlen() instead.` OG is not calling this deprecated method, but it is still thrown in code we depend on.

mariano-dagostino:
> Exception: Warning: call_user_func() expects parameter 1 to be a valid callback, class 'Drupal\dynamic_page_cache\EventSubscriber\DynamicPageCacheSubscriber' does not have a method 'onRouteMatch'.

This issue is really strange. I searched for 'onRouteMatch' for the entire codebase of drupal 8.6.x, og, coder, drupal/console and I was not able to find a single entry of that event.

> \Drupal\Component\Utility\Unicode::strlen() is deprecated
>  \SafeMarkup::format() is scheduled for removal in Drupal 9.0.0

Same applies for Unicode::strlen() and SafeMarkup::format(), the only parts of the code that have those calls are the tests that check the deprecations messages are triggered.

pfrenssen:
This has been fixed in the meantime, we're even compatible with 8.7 now.

PR #447, opened 2018-12-11
Start testing Drupal 8.7 and PHP 7.3. Stop testing deprecated versions.

pfrenssen:
- Disable deprecated versions of PHP and Drupal core.
- Start testing on Drupal 8.7.x and PHP 7.3, but still allow failures.
- We are already testing on 8.6.x but these are failing because `drupal_set_message()` is deprecated. Fix this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,30 +2,27 @@ language: php
 sudo: true
 
 php:
-  - 5.6
-  - 7.0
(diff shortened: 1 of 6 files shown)


pfrenssen:
This also fixes #400.

amitaibu:
Thanks!

PR #448, opened 2018-12-11
Remove user_load call in og_get_user_roles

jorrit:
This prevents adding a partial user object to the entity cache because og_get_user_roles is called during user_save().

### DIFF
--- a/og.module
+++ b/og.module
@@ -2578,7 +2578,8 @@ function og_get_all_group_content_bundle() {
  * @param $entity_type
  *   The entity type.
  * @param $entity
- *   The entity object. If empty the current user will be used.
+ *   (optional) The entity object or entity ID. If empty, and $entity_type is
+ *   "user", the current user will be used.
  * @param $states
  *   (optional) Array with the membership states to check against. If omitted
  *   then only active (OG_STATE_ACTIVE) memberships will be checked.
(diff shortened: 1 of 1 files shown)


jorrit:
I forgot to mention: the effect of having a partial object in de user entity cache is that the realname will not be properly generated for email notifications.

jacobembree:
This makes sense to me.

og.module
```diff
@@ -2775,21 +2776,19 @@ function og_get_user_roles($group_type, $gid, $uid = NULL, $include = TRUE, $che
...
 
-  $account = user_load($uid);
-
   $identifier = implode(':', array($group_type, $gid, $uid, $include));
   if (isset($roles[$identifier])) {
     return $roles[$identifier];
   }
 
-  $is_blocked = og_is_member($group_type, $gid, 'user', $account, array(OG_STATE_BLOCKED));
+  $is_blocked = og_is_member($group_type, $gid, 'user', $uid, array(OG_STATE_BLOCKED));
```
MPParsley, line 2784: This doesn't seem right, the signature of og_is_member still expects an entity yet here you pass an integer.

It might work if og_get_user_roles was called with $uid=NULL but otherwise it'll fail.

jorrit: `og_is_member()` passes the `$entity` parameter to `og_get_entity_groups()`, it doesn't need the parameter for anything else. `og_get_entity_groups()` accepts either a user object or a user ID. In fact, when an object is passed only the ID is used:

```php
if (is_object($entity)) {
    // Get the entity ID.
    list($id) = entity_extract_ids($entity_type, $entity);
  }
```

The docblock of `og_is_member()` could be updated to reflect that.

claudiu-cristea:

Issue #336, opened 2017-10-30
static caching in MembershipManager::getMemberships() doesn't work properly with states


joachim-n:
```
  public function getMemberships(AccountInterface $user, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
    // Get a string identifier of the states, so we can retrieve it from cache.
    sort($states);
    $states_identifier = implode('|', array_unique($states));

    $identifier = [
      __METHOD__,
      'user',
      $user->id(),
      $states_identifier,
    ];
    $identifier = implode(':', $identifier);

    // Return cached result if it exists.
    if (isset($this->cache[$identifier])) {
      return $this->cache[$identifier];
    }
```

If I ask for the memberships for user X, with all states, I pass in an empty array for $states (though that's not documented - #335). The cache ID is then:

> "Drupal\og\MembershipManager::getMemberships:user:1:"

If I then ask for the memberships for user X which are active, the cache ID is:

> Drupal\og\MembershipManager::getMemberships:user:1:active"

That results in a cache miss, even though the data is loaded already.

pfrenssen:
Nice find. Together with the documentation issue in #335 I think we should introduce a new array constant that contains all possible states and document that this should be used.

We can keep the empty array passing too. We have a bunch of tests that verify that this works so this was clearly intended as such, but I think it would be better to just document that people should pass `OgMembershipInterface::STATES` since this will make it clearer what is going to be returned.

amitaibu:
> but I think it would be better to just document that people should pass OgMembershipInterface::STATES since this will make it clearer what is going to be returned.

I agree

pfrenssen:
This has been fixed in #449

PR #449, opened 2018-12-11
Clarify how to work with all membership states.

### LINKED ISSUE #335: MembershipManager::getMemberships() and getMembership() don't document how to get all states
```
   * @param array $states
   *   (optional) Array with the state to return. Defaults to active.
```

This doesn't tell me how to get all states.

Looking at the code, an empty array will do it.

pfrenssen:
Fixes #335 and #336.

### DIFF
--- a/src/OgMembershipInterface.php
+++ b/src/OgMembershipInterface.php
@@ -41,7 +41,11 @@ interface OgMembershipInterface extends ContentEntityInterface, EntityOwnerInter
   /**
    * An array containing all possible group membership states.
    */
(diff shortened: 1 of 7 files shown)


src/OgMembershipInterface.php
```diff
@@ -38,6 +38,11 @@
    */
   const STATE_BLOCKED = 'blocked';
 
+  /**
+   * An array containing all possible group membership states.
+   */
+  const STATES = [self::STATE_ACTIVE, self::STATE_PENDING, self::STATE_BLOCKED];
```
amitaibu: Should we call it `All_STATES`?

amitaibu:
Thanks,

PR #450, opened 2018-12-11
Use cache api to store the group relation map

### LINKED ISSUE #440: State API shouldn't be used for caching
In `GroupTypeManager::refreshGroupRelationMap()` state API is used to cache the map:

```
    // Cache the map.
    $this->state->set(self::GROUP_RELATION_MAP_CACHE_KEY, $this->groupRelationMap);
``` 

We shouldn't use state for caching. It can break things when another entity type has an og_reference field (e.g. og_menu).

A simple workaround is to clear the state manually:

``` 
drush state:delete og.group_manager.group_relation_map
```

MPParsley:
This should fix #440.

I used simple Cache API but it may be useful to add cache tags and invalidations here.

### DIFF
--- a/og.services.yml
+++ b/og.services.yml
@@ -42,7 +42,7 @@ services:
     arguments: ['@entity_type.manager', '@entity_field.manager']
   og.group_type_manager:
     class: Drupal\og\GroupTypeManager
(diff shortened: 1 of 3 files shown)


MPParsley:
@pfrenssen this is ready for review.

pfrenssen (approved):

PR #451, opened 2018-12-11
Provide a way to retrieve all group memberships by role

pfrenssen:
Programmatically getting a list of all group memberships filtered by a certain role can be non-trivial to implement (for example getting a list of all administrators of a group).

Because all memberships are getting the "member" role by default this data is not saved in the database. This is implemented similarly to how core deals with user roles, but it means a simple entity query that filters by the role reference will not return any "member" roles.

The trick is to omit the condition when filtering by role. This might sound easy enough, but this is not obvious if you are not familiar with the inner workings of OG. To illustrate this, in my project I found a method that solved it by retrieving the full list of memberships and then filtering it down to get the required role. For one large group this was loading 23000 memberships to find 1 administrator :sweat_smile:

Let's help out by adding a convenient method to the `MembershipManager` that retrieves the memberships efficiently and caches the results.

### DIFF
--- a/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
+++ b/tests/src/Kernel/Entity/GroupMembershipManagerTest.php
@@ -9,8 +9,12 @@ use Drupal\KernelTests\KernelTestBase;
 use Drupal\entity_test\Entity\EntityTest;
 use Drupal\node\Entity\Node;
 use Drupal\node\Entity\NodeType;
+use Drupal\og\Entity\OgRole;
 use Drupal\og\Og;
 use Drupal\og\OgGroupAudienceHelperInterface;
+use Drupal\og\OgMembershipInterface;
+use Drupal\og\OgRoleInterface;
+use Drupal\Tests\og\Traits\OgMembershipCreationTrait;
 use Drupal\user\Entity\User;
 
 /**
@@ -20,6 +24,8 @@ use Drupal\user\Entity\User;
  * @coversDefaultClass \Drupal\og\MembershipManager
  */
 class GroupMembershipManagerTest extends KernelTestBase {
+
+  use OgMembershipCreationTrait;
 
   /**
    * {@inheritdoc}
@@ -48,11 +54,25 @@ class GroupMembershipManagerTest extends KernelTestBase {
   protected $groupContent;
 
   /**
+   * Test users.
+   *
+   * @var \Drupal\user\UserInterface[]
+   */
+  protected $users;
+
+  /**
    * The entity type manager.
    *
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
   protected $entityTypeManager;
+
+  /**
+   * The membership manager. This is the system under test.
+   *
+   * @var \Drupal\og\MembershipManagerInterface
+   */
+  protected $membershipManager;
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


pfrenssen:
This depends on #446 and #447.

amitaibu:
Thanks - needs a re-roll

pfrenssen:
Rebased it on the latest 8.x-1.x.

tests/src/Kernel/Entity/GroupMembershipManagerTest.php
```diff
@@ -318,4 +329,315 @@ public function groupContentProvider() {
...
+        foreach ($test_cases as $test_case) {
+          $group = $this->groups[$entity_type_id][$group_key];
+          $role_names = $test_case['roles'];
+          $states = $test_case['states'];
+          $expected_memberships = $test_case['expected_memberships'];
+
+          /** @var \Drupal\og\OgMembershipInterface[] $actual_memberships */
+          $actual_memberships = $this->membershipManager->getGroupMembershipsByRoleNames($group, $role_names, $states);
+
+          $this->assertEquals(count($expected_memberships), count($actual_memberships));
```
brummbar: Small nit: can we use `assertCount()` in here?

sandervd:
Overall everything looks fine. And it is consistent with the rest of the code.
I am a bit conflicted over the use of the static caching of the query result, but I know the rest of the code uses the same pattern.
It looks like a micro optimization, which might even cause incorrect behavior when the method is called before and after an update has been made.

brummbar (changes requested):
Looks fine, just a small nit.

pfrenssen:
That cache is being cleared when an update is made in `og_invalidate_group_content_cache_tags()` which is called by `hook_entity_update()`. It is only clearing the cache when the relationships actually change, to maximise the usefulness of the cache.

tests/src/Kernel/Entity/GroupMembershipManagerTest.php
MPParsley: @brummbar, you can use the github suggestion widget to do suggestions like this.
```suggestion
          $this->assertCount(count($expected_memberships), $actual_memberships);
```

brummbar: Cool, didn't know this functionality.

brummbar: Or even `assertSameSize()`.

```suggestion
          $this->assertSameSize($expected_memberships, $actual_memberships);
```

pfrenssen: @MPParsley I didn't know this feature either, it looks very handy! Unfortunately the suggestion wasn't correct :-/

I went with the `$this->assertSameSize()`, that is the perfect match for our use case!

pfrenssen: This new suggestion feature seems to have some bugs regarding the CI pipelines, the build was started but immediately canceled. It's still in beta, I'm sure they'll iron this out soon enough.

MPParsley: @pfrenssen indeed, assertSameSize() is better.

Thanks for the update about the CI pipeline. I'm not sure how it's set up on this project but on our projects Travis is triggered (and runs) as if it were a regular commit.

This feature is going to be a great help for open source collaboration as the threshold is much lower to contribute (especially when compared to an archaic d.o patch workflow...).

brummbar (approved):
Good to go!

amitaibu:
Thank you @pfrenssen et al!

pfrenssen:
Great! Thanks @amitaibu @sandervd @brummbar @MPParsley for the reviews, it was a real pleasure getting such a response :)

@amitaibu are you OK if I tag a new release? Especially because the current alpha1 release is not compatible with Drupal 8.6.x right now.

amitaibu:
> @amitaibu are you OK if I tag a new release? 

Sure, go ahead :)

PR #452, opened 2018-12-17
Test against PHP 7.3 instead of 7.0

jacobembree:
PHP 7.0 is no longer supported.  It would be good to test against a more recent version.
https://secure.php.net/supported-versions.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: php
 
 php:
   - 5.6
-  - 7.0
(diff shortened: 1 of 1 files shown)


MPParsley:
This has been fixed in #447.

jacobembree:
#447 is for Drupal 8.  This is the backport for Drupal 7.

pfrenssen:
Thanks!

Issue #455, opened 2018-12-24
Upgrade hook infinite loop when no memberships


xurizaemon:
In the case where og_upgrade_8001() finds no memberships, it gets stuck in a loop outputting -

```text
 [warning] Division by zero og.install:71
 [notice] Processed 0 of 0 memberships (nan% complete)
 [notice] Update started: og_update_8001
 [warning] Division by zero og.install:71
 [notice] Processed 0 of 0 memberships (nan% complete)
 [notice] Update started: og_update_8001
 [warning] Division by zero og.install:71
 [notice] Processed 0 of 0 memberships (nan% complete)
 [notice] Update started: og_update_8001
```

xurizaemon:
Resolved in above commit

PR #456, opened 2018-12-24
Issue #455 - Early exit if no memberships to process

xurizaemon:
If we don't get any memberships to process, say so and set `$sandbox['finished'] = 1`.


amitaibu:
Thanks, and welcome to OG :)

PR #462, opened 2019-02-01
Add default access check

pebosi:
See: https://www.drupal.org/project/og/issues/2810795

### DIFF
--- a/config/optional/views.view.og_members_overview.yml
+++ b/config/optional/views.view.og_members_overview.yml
@@ -22,8 +22,9 @@ display:
     position: 0
     display_options:
       access:
-        type: none
-        options: {  }
+        type: perm
+        options:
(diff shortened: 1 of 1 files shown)


amitaibu:
Thanks. Can you please attach a screenshot demonstrating the change of access

amitaibu:
Actually, it looks good. Thanks!

pfrenssen:
I wonder why the `access content` permission was chosen instead of `access user profiles`? This view is displaying user profiles, not nodes.

amitaibu:
You might be right, but I think it's not a case that it's very clear. I wouldn't mind changing it to `access user profiles`.

MPParsley:
Closed https://www.drupal.org/project/og/issues/2810795

PR #463, opened 2019-02-20
Set bundle form value in group permissions form

jorrit:
`$form['bundle']['#value']` is now also set when the group permissions form is displayed for a single group.

I use this value in a form alter hook and with this patch I only need to look at `$form['bundle']['#value']` to handle both the generic permissions form and the group specific form.


alesrebec:
Looks like it nicely fills out the missing bit.
All clean.
Thanks @jorrit

jorrit:
Thanks for merging @alesrebec !

PR #467, opened 2019-03-14
Use full pager on member overview

MPParsley:
At the moment the mini pager is used for the member overview:
mini pager

Most Drupal admin overviews (e.g. content, people) use 50 items per page and a full pager:
full pager


amitaibu:
Thanks, tests seems to be red

amitaibu:
Thanks.

PR #468, opened 2019-03-20
Small optimization for adding/removing roles to/from memberships

pfrenssen:
The plugins that handle the addition and removal of roles do the following to retrieve the group bundle:

```
$membership->getGroup()->bundle();
```

However since the bundle is directly available on the membership it is not needed to load the full group entity for this. It can be changed to:

```
$membership->getGroupBundle();
```

### DIFF
--- a/src/Plugin/Action/AddSingleOgMembershipRole.php
+++ b/src/Plugin/Action/AddSingleOgMembershipRole.php
@@ -26,7 +26,7 @@ class AddSingleOgMembershipRole extends ChangeSingleOgMembershipRoleBase {
     $role_name = $this->configuration['role_name'];
     $role_id = implode('-', [
       $membership->getGroupEntityType(),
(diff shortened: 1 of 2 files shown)


MPParsley (approved):
Tests are green and looks good.

```
  /**
   * {@inheritdoc}
   */
  public function getGroup() {
    $entity_type = $this->get('entity_type')->value;
    $entity_id = $this->get('entity_id')->value;
    if (empty($entity_type) || empty($entity_id)) {
      return NULL;
    }
    return \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
  }

  /**
   * {@inheritdoc}
   */
  public function getGroupBundle() {
    return $this->get('entity_bundle')->value;
  }
```

amitaibu:
@MPParsley go ahead, use your commit power! :)

Issue #470, opened 2019-04-18
Unnecessarily gendered language in a comment


MPParsley:
```
FILE: ...ild/Gizra/og/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 177 | ERROR | Unnecessarily gendered language in a comment
--------------------------------------------------------------------------------
FILE: ...ome/travis/build/Gizra/og/tests/src/Kernel/Entity/SelectionHandlerTest.php
--------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 136 | ERROR | Unnecessarily gendered language in a comment
 143 | ERROR | Unnecessarily gendered language in a comment
--------------------------------------------------------------------------------
Time: 4.24 secs; Memory: 14MB
```

https://github.com/Gizra/og/blob/e8d32cbbb5452de8d19512d53945611c136f0a06/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php#L177

https://github.com/Gizra/og/blob/e8d32cbbb5452de8d19512d53945611c136f0a06/tests/src/Kernel/Entity/SelectionHandlerTest.php#L136

https://github.com/Gizra/og/blob/e8d32cbbb5452de8d19512d53945611c136f0a06/tests/src/Kernel/Entity/SelectionHandlerTest.php#L143

pfrenssen:
Fixed in #471. Thanks!

PR #471, opened 2019-04-18
Unnecessarily gendered language in a comment

MPParsley:
Solves #470

### DIFF
--- a/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
+++ b/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
@@ -174,7 +174,7 @@ class OgGroupContentOperationAccessTest extends KernelTestBase {
     }
 
     // Create a 'blocked' user. This user is identical to the normal
-    // 'authenticated' member, except that it has the 'blocked' state.
+    // 'authenticated' member, except that they have the 'blocked' state.
     $this->users['blocked'] = User::create(['name' => $this->randomString()]);
     $this->users['blocked']->save();
     $this->createOgMembership($this->group, $this->users['blocked'], NULL, OgMembershipInterface::STATE_BLOCKED);
--- a/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
+++ b/tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
@@ -174,7 +174,7 @@ class OgGroupContentOperationAccessTest extends KernelTestBase {
     }
 
     // Create a 'blocked' user. This user is identical to the normal
-    // 'authenticated' member, except that she has the 'blocked' state.
+    // 'authenticated' member, except that it has the 'blocked' state.
     $this->users['blocked'] = User::create(['name' => $this->randomString()]);
     $this->users['blocked']->save();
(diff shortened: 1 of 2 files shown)


tests/src/Kernel/Entity/SelectionHandlerTest.php
```diff
@@ -133,14 +133,14 @@ public function testSelectionHandler() {
...
-   * We need to verify that each user get the groups he own in the normal widget
+   * We verify that each user gets the groups it owns in the normal widget
    * and the other users group's in the other groups widget and vice versa.
    */
   public function testSelectionHandlerResults() {
     $user1_groups = $this->createGroups(2, $this->user1);
     $user2_groups = $this->createGroups(2, $this->user2);
 
-    // Checking that the user get the groups he mange.
+    // Check that the user gets the groups he manages.
```
mradcliffe: There's a "he" here as well, @MPParsley.

mradcliffe (commented):
Found another term.

tests/src/Kernel/Entity/SelectionHandlerTest.php
MPParsley:
```suggestion
    // Check that the user gets the groups it manages.
```

MPParsley: Yeah, so I noticed, updated it.

MPParsley: FYI, you can propose inline code suggestions with cmd-g, awesome feature :-)

mradcliffe (approved):
It's passing now. :+1:

zerolab:
I would suggest replacing he/she with "they", rather than "it"

tests/src/Kernel/Entity/SelectionHandlerTest.php
```diff
@@ -133,14 +133,14 @@ public function testSelectionHandler() {
...
-   * We need to verify that each user get the groups he own in the normal widget
+   * We verify that each user gets the groups it owns in the normal widget
    * and the other users group's in the other groups widget and vice versa.
    */
   public function testSelectionHandlerResults() {
     $user1_groups = $this->createGroups(2, $this->user1);
     $user2_groups = $this->createGroups(2, $this->user2);
 
-    // Checking that the user get the groups he mange.
+    // Check that the user gets the groups it manages.
```
MPParsley:
```suggestion
    // Check that users get the groups they manage.
```

tests/src/Kernel/Access/OgGroupContentOperationAccessTest.php
```diff
@@ -174,7 +174,7 @@ protected function setUp() {
     }
 
     // Create a 'blocked' user. This user is identical to the normal
-    // 'authenticated' member, except that she has the 'blocked' state.
+    // 'authenticated' member, except that it has the 'blocked' state.
```
amitaibu:
```suggestion
    // 'authenticated' member, except that they have the 'blocked' state.
```

(also change below it => they)

pfrenssen (approved):

PR #473, opened 2019-04-22
Add support for Drupal 8.7.x

pfrenssen:
The first release candidate of Drupal 8.7 has been released. This is a good moment to start supporting Drupal 8.7, and start experimental support of Drupal 8.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ env:
   global:
     - COMPOSER_MEMORY_LIMIT=2G
   matrix:
-    - TEST_SUITE=8.6.x
     - TEST_SUITE=8.7.x
     - TEST_SUITE=8.8.x
     - TEST_SUITE=PHP_CodeSniffer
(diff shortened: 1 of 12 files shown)


pfrenssen:
There were a whole bunch of deprecation warnings when upgrading to 8.7. Most of them were pretty easy to fix since they link to change records with relevant information.

One of the changes is incompatible with Drupal 8.6 and throws a deprecation warning on 8.7: https://www.drupal.org/node/3016699

We could exclude this deprecation warning from being checked, or we can declare a dependency on 8.7. I have opted for the latter, since this is consistent with how we did previous releases for OG.

This change will not be disruptive for end users, since they can still use the current 8.x-1.0-alpha2 release for as long as they are on Drupal 8.6, and will automatically update to the upcoming alpha3 release when they move to Drupal 8.7.

The release of 8.7 is planned for next week, and at that point 8.6 will be deprecated anyway. We could plan a release for OG shortly after Drupal 8.7 is released.

MPParsley (approved):

pfrenssen:
Thanks for review!

amitaibu:
> I have opted for the latter, since this is consistent with how we did previous releases for OG

I agree, good call.

> We could plan a release for OG shortly after Drupal 8.7 is released.

Sounds like a good plan.

Issue #472, opened 2019-04-22
Provide an `og-group-member-list` cache tag that is invalidated when a membership changes


pfrenssen:
In #216 we have added a `og-group-content` cache tag that can be used to tag lists of group content. It is invalidated whenever group content changes.

Similar to this we need a `og-group-member-list` cache tag that is invalidated whenever the memberships of a group change. This can be used to tag member lists, member counts etc. When a new member joins a group, or a member leaves, is blocked, or changes their role in the group we can invalidate this tag and the member lists and member counts will be recalculated.

As part of this work we should add some functionality to OG that uses this tag. The simplest use case would be to include a "Member count" block that simply shows the number of active members. We can then demonstrate that the block is correctly cached and invalidated whenever a member joins / leaves / is blocked / changes their role.

pfrenssen:
I have started working on this. I looked into using the "Members overview" view as a use case for demonstrating the cache tag, but this is not a good candidate since the view is currently uncacheable due to being a bulk operations form (ref. [Issue #2503009: Consider making the bulk operations form cacheable](https://www.drupal.org/node/2503009)).

I'll make a simple block that shows the user count.

PR #475, opened 2019-04-23
Provide a method to retrieve the membership IDs of a given group

pfrenssen:
For many of the methods in `MembershipManager` that return entity data we offer two versions for convenience: one that returns IDs, and one that returns fully loaded entities. See for example:

* `::getUserGroups()` vs `::getUserGroupIds()`
* `::getGroups()` vs `::getGroupIds()`

This PR adds a new method `::getGroupMembershipIdsByRoleNames()` which is equivalent to `::getGroupMembershipsByRoleNames()` but avoids to load the full entity objects.

I am needing this since I am implementing a block that shows the membership count of a group for #472 but it would be wasteful to load the full `OgMembership` entities just to get a count.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -178,7 +178,8 @@ class MembershipManager implements MembershipManagerInterface {
    * {@inheritdoc}
    */
   public function getGroupMembershipsByRoleNames(EntityInterface $group, array $role_names, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
-    return $this->loadMemberships($this->getGroupMembershipIdsByRoleNames($group, $role_names, $states));
+    $ids = $this->getGroupMembershipIdsByRoleNames($group, $role_names, $states);
+    return $this->loadMemberships($ids);
   }
 
   /**
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -124,7 +124,7 @@ class MembershipManager implements MembershipManagerInterface {
   /**
    * {@inheritdoc}
    */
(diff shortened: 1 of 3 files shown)


src/MembershipManager.php
```diff
@@ -171,7 +171,14 @@ public function getGroupMembershipsByRoleNames(EntityInterface $group, array $ro
...
 
-    return $this->loadMemberships($this->cache[$identifier]);
+    return $this->cache[$identifier];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getGroupMembershipsByRoleNames(EntityInterface $group, array $role_names, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
+    return $this->loadMemberships($this->getGroupMembershipIdsByRoleNames($group, $role_names, $states));
```
MPParsley:
```suggestion
    $ids = $this->getGroupMembershipIdsByRoleNames($group, $role_names, $states);
    return $this->loadMemberships($ids);
```

src/MembershipManagerInterface.php
```diff
@@ -88,13 +88,32 @@ public function getMemberships(AccountInterface $user, array $states = [OgMember
    */
   public function getMembership(EntityInterface $group, AccountInterface $user, array $states = [OgMembershipInterface::STATE_ACTIVE]);
 
+  /**
+   * Returns the membership IDs of the given group filtered by role name.
```
MPParsley:
```suggestion
   * Returns the membership IDs of the given group filtered by role names.
```

MPParsley (approved):

MPParsley (approved):

PR #479, opened 2019-04-25
MembershipManager returns memberships of the wrong group

pfrenssen:
There is a bug in `MembershipManager::getGroupMembershipIdsByRoleNames()` and `::getGroupMembershipsByRoleNames()`: when there are multiple groups with different entity types but the same entity ID then the results of the first group will be cached and will be returned for the other groups.

We should include the entity type in the cache ID.


MPParsley (approved):
Nice find!

pfrenssen:
Thanks!

PR #480, opened 2019-04-25
MembershipManager still returns a user's memberships after the user is deleted

pfrenssen:
`MembershipManager::getGroupMembershipsByRoleNames()` and `MembershipManager::getGroupMembershipIdsByRoleNames()` still return the memberships of a user from the static cache after a user is deleted.

There are two possible solutions for this:

* We can call `MembershipManager->reset()` in `OgMembership::delete()` just as it is now being done in `OgMembership::save()`.
* We can replace the internal static `$cache` property with core's static `MemoryBackend` cache backend which has full support for cache tag invalidation.

In this PR I am going to explore the second solution. By using `MemoryBackend` we can add the correct cache tags to each of our cache entries, and they will be invalidated automatically when needed. This should allow us to remove all calls to `MembershipManager::reset()` and clean up some code that does unnecessary cache clears.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -381,13 +381,6 @@ class MembershipManager implements MembershipManagerInterface {
   }
 
   /**
-   * {@inheritdoc}
-   */
-  public function reset() {
-    $this->staticCache->deleteAll();
-  }
-
-  /**
    * Prepares a conditional array for use in a cache identifier and query.
    *
    * This will filter out any duplicate values from the array and sort the
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -3,6 +3,9 @@
 namespace Drupal\og;
 
 use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Cache\Cache;
+use Drupal\Core\Cache\CacheBackendInterface;
+use Drupal\Core\Cache\CacheTagsInvalidatorInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
@@ -11,16 +14,16 @@ use Drupal\field\FieldStorageConfigInterface;
 use Drupal\og\Entity\OgMembership;
 
 /**
- * Membership manager.
+ * Service for managing memberships and group content.
  */
 class MembershipManager implements MembershipManagerInterface {
 
   /**
-   * Static cache of the memberships and group association.
+   * The static cache backend.
    *
-   * @var array
+   * @var \Drupal\Core\Cache\CacheBackendInterface
    */
-  protected $cache;
+  protected $staticCache;
 
   /**
    * The entity type manager.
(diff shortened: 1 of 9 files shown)


pfrenssen:
Started by adding a failing test, this demonstrates that `MembershipManager` still returns the memberships of the deleted user.

pfrenssen:
OK so using the memory backend that supports invalidation based on cache tags fixes the issue without requiring any changes to the code base.

I'm now removing all calls to `MembershipManager::reset()`. I'm expecting these to be no longer needed.

pfrenssen:
Great! Working fine, and the bug is fixed :)

src/MembershipManager.php
```diff
@@ -399,6 +406,22 @@ protected function prepareConditionArray(array $value, array $default = NULL) {
...
 
+  /**
+   * Stores the given list of membership IDs in the static cache backend.
+   *
+   * @param string $cid
+   *   The cache ID.
+   * @param array $membership_ids
+   *   The list of membership IDs to store in the static cache.
+   */
+  protected function cacheMembershipIds($cid, array $membership_ids) {
```
MPParsley, line 417: Should we add this to MembershipManagerInterface?

MPParsley (approved):

src/MembershipManager.php
pfrenssen: No I don't think this should be part of the public API, this is just an internal caching strategy. We should be able to switch to different caching without affecting any end user websites.

src/MembershipManager.php
```diff
@@ -11,16 +14,16 @@
...
 
   /**
-   * Static cache of the memberships and group association.
+   * The static cache backend.
    *
-   * @var array
+   * @var \Drupal\Core\Cache\CacheBackendInterface
    */
-  protected $cache;
+  protected $staticCache;
```
claudiu-cristea, line 26: @pfrenssen, I wonder why we're only doing memory caching here. This looks to me a good candidate to double the memory caching with persistent caching (database, Redis, etc.). All we need here is to declare a `\Drupal\Core\Cache\BackendChain` service and add a memory and a cache bin backends to it. Then replace the static cache service with this one. That would improve dramatically the performance.

src/MembershipManager.php
```diff
@@ -366,13 +380,6 @@ public function isMemberBlocked(EntityInterface $group, AccountInterface $user)
     return $this->isMember($group, $user, [OgMembershipInterface::STATE_BLOCKED]);
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  public function reset() {
```
claudiu-cristea, line 372: IMO, This should have been deprecated instead of fully removed. I know we're not offering yet BC but still, given the fact that so many sites are already using OG...

src/MembershipManager.php
claudiu-cristea: I've opened #485 to fix this.

src/MembershipManager.php
MPParsley: You're right, we could have deprecated this.
It broke tests in #181.

PR #482, opened 2019-04-25
Groups and roles by permissions

idimopoulos:
Hi! It is me again!
I was thinking that we could populate the helper methods a bit more since I have extra use cases that at least for us happen quite often.
In our case, we have a case where we need to query for all groups a user can share some content in. To do this, we need to query all groups that the user has a certain permission in. 
By looking into the og code, it seems that two things are missing from this:
* Retrieve all roles with a given permission.
* Retrieve all groups with a given role.

Obviously, a role id also include the entity type id and the bundle so my thoughts are:
* Provide a method to the `OgRoleManager` service that can load all roles with a given permission. This method can optionally filter by entity type id and bundle.
* Provide a method to the `MembershipManager` that loads all memberships given some roles and then retrieve the groups. For this method I am using the `MembershipManager::getMemberships` method so cache is already being taken care by existing methods.

I am attaching the two methods, the interface definitions and some tests related to them.

Please, let me know if these are two specific to be included in og so that I can include them explicitly in our project instead.


src/MembershipManager.php
```diff
@@ -128,6 +128,32 @@ public function getMembership(EntityInterface $group, AccountInterface $user, ar
...
+    });
+
+    $group_ids = [];
+    foreach ($memberships as $membership) {
+      $group_ids[$membership->getGroupEntityType()][] = $membership->getGroupId();
+    }
+
+    $groups = [];
+    foreach ($group_ids as $entity_type => $entity_ids) {
+      $groups[$entity_type] = $this->entityTypeManager->getStorage($entity_type)->loadMultiple($entity_ids);
```
pfrenssen: This is looping twice over what is basically the same data. I also see that this last loop is a duplication of the loop in `MembershipManager::getGroups()`. If possible:

* Split this method in two: `::getUserGroupsByRoles()` and `::getUserGroupIdsByRoles()` - it is only a matter of time before someone will require to just get the group IDs.
* Split off the loading of the groups into a separate protected method `::loadGroups()` which can then be called by both `::getUserGroupsByRoles()` and `::getGroups()`.

src/OgRoleManager.php
```diff
@@ -132,6 +132,24 @@ public function getRolesByBundle($entity_type_id, $bundle) {
     return $this->ogRoleStorage()->loadByProperties($properties);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getRolesByPermissions(array $permissions, $entity_type_id = NULL, $bundle = NULL) {
```
pfrenssen: This is such a helpful method! Great suggestion!

src/OgRoleManagerInterface.php
```diff
@@ -58,6 +58,24 @@ public function getRequiredDefaultRoles();
...
   public function getRolesByBundle($entity_type_id, $bundle);
 
+  /**
+   * Returns all the roles that have a specific permission.
+   *
+   * Optionally filter the roles by entity type id and bundle.
+   *
+   * @param array $permissions
+   *   An array of permissions that the roles must have. Any permission included
+   *   will return the role.
```
pfrenssen: This is not clear to me, does this mean that if I pass two permissions `['edit any article content', 'delete any article content']` that it will return all the roles that have one or both of these permissions?

I would assume that the more common case would be to only return the roles that have all of the passed permissions.

If you have a use case for matching any condition it is maybe an option to introduce a flag `$require_all_permissions = TRUE` that allows both?

tests/src/Kernel/Entity/GetUserGroupsTest.php
```diff
@@ -71,13 +73,6 @@ class GetUserGroupsTest extends KernelTestBase {
    */
   protected $groupBundle;
 
-  /**
-   * The machine name of the group content node type.
-   *
-   * @var string
-   */
-  protected $groupContentBundle;
-
```
pfrenssen, line 80: Ah this was indeed not used. Good catch!

tests/src/Kernel/Entity/GetUserGroupsTest.php
```diff
@@ -234,6 +228,32 @@ public function testIsMemberStates() {
     $this->assertFalse(Og::isMemberPending($this->group1, $this->user3));
   }
 
+  /**
+   * Tests retrieval of groups filtered by roles.
+   */
+  public function testGetGroupsByRoles() {
+    /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+    $membership_manager = \Drupal::service('og.membership_manager');
```
pfrenssen: I would say to store the membership manager in a property and populate it in the setup using `$this->membershipManager = $this->container->get('og.membership_manager');` but I see this is actually consistent with the other tests in this class. So it is fine as it is right now, but if you want to improve it feel free!

tests/src/Kernel/Entity/GetUserGroupsTest.php
```diff
@@ -234,6 +228,32 @@ public function testIsMemberStates() {
...
+      ->setGroupBundle($this->groupBundle)
+      ->save();
+
+    $member_role = OgRole::getRole('entity_test', $this->groupBundle, OgRoleInterface::AUTHENTICATED);
+
+    $this->createOgMembership($this->group2, $this->user1, [$og_role->getName()]);
+    $groups = $membership_manager->getUserGroupsByRoles($this->user1, [$member_role]);
+    $this->assertCount(2, $groups['entity_test']);
+    $groups = $membership_manager->getUserGroupsByRoles($this->user1, [$og_role]);
+    $this->assertCount(1, $groups['entity_test']);
```
pfrenssen, line 289: It's only testing basic use cases. The `$states` filter is not tested, and neither is passing multiple roles in the `$roles` array.

tests/src/Kernel/Entity/OgRoleTest.php
```diff
@@ -243,6 +243,47 @@ public function testRoleCreate() {
...
+    $og_role3 = OgRole::create();
+    $og_role3->setName(mb_strtolower($this->randomMachineName()))
+      ->setLabel($this->randomString())
+      ->setGroupType('entity_test')
+      ->setGroupBundle('group')
+      ->grantPermission('administer group')
+      ->save();
+
+    $roles = $role_manager->getRolesByPermissions(['access content']);
+    $this->assertCount(2, $roles);
```
pfrenssen: Can we also test retrieval of roles with multiple permissions, and the optional parameters `$entity_type_id` and `$bundle`?

pfrenssen (changes requested):

src/OgRoleManagerInterface.php
idimopoulos: I will refactor it with your latest suggestion, a flag to control this, as that might benefit a more wide variety of users.

idimopoulos:
@pfrenssen I have fixed all of your remarks and extended the tests to test a few more cases.
I have implemented the `$require_all` flag, sorry for not naming it `$require_all_permissions` but I am using it in 2 different places so I though we could keep it simpler to remember the usage by having the same name of the variable. Very small silly thought but it makes a bit of sense to me.

Speaking of which, I took your idea of having a "require all" flag for permissions and extended the other API methods in this PR, `getUserGroupsIdsByRoles` and `getUserGroupsByRoles` to also have such a flag. Now you can request groups of the user where he has _all_ of the requested roles. 
This could be useful in searches where let's say you want the groups where the user is the author *and* the publisher or something similar. 
Tests for the roles were also extended accordingly.

I did not use a provider for the tests because I wanted to have a test for the default values in the methods as well.

src/MembershipManager.php
```diff
@@ -128,6 +123,35 @@ public function getMembership(EntityInterface $group, AccountInterface $user, ar
     return NULL;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getUserGroupsIdsByRoles(AccountInterface $user, array $roles, bool $require_all = FALSE, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
```
pfrenssen: I think the `$require_all` flag should default to `TRUE`.

Roles in Drupal are additive and tailored to a specific set of actions that the user needs to perform in the site. In corporate sites a user typically has many roles. For example a HR manager might have access see payroll information in specific departments, and a business analyst might see production processes. If you are making a page that correlates payroll with business processes you would restrict it to people that have both roles.

The other case makes less sense. What are the use cases (except from a permission lookup) where you would need to get the groups where a user has one role, or another role, or maybe a third one?

Also the name `$require_all` is ambiguous. Both the roles and the states are arrays that filter on multiple values, so it is not clear to which this is referring. Let's call it `$require_all_roles`. And put it as the last parameter so the parameter ordering is consistent with the rest of the class.

src/MembershipManager.php
```diff
@@ -128,6 +123,35 @@ public function getMembership(EntityInterface $group, AccountInterface $user, ar
...
+    foreach ($memberships as $membership) {
+      $group_ids[$membership->getGroupEntityType()][] = $membership->getGroupId();
+    }
+    return $group_ids;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getUserGroupsByRoles(AccountInterface $user, array $roles, bool $require_all = FALSE, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
```
pfrenssen: Same here, let's rename the flag, make it default to `TRUE` and move it to the last position in the argument list.

src/MembershipManager.php
```diff
@@ -407,6 +426,24 @@ protected function prepareConditionArray(array $value, array $default = NULL) {
     return array_unique($value);
   }
 
+  /**
+   * Loads the entities of a structured array of entity ids.
```
pfrenssen: It is more common to use `associative array` instead of `structured array`.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
    */
   public function getUserGroups(AccountInterface $user, array $states = [OgMembershipInterface::STATE_ACTIVE]);
 
+  /**
+   * Returns an array of groups filtered by the og roles of the user.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $user
+   *   The user to get the groups for.
+   * @param array $roles
```
pfrenssen: Let's call this type hint `\Drupal\og\OgRoleInterface[]`.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
...
+   *   (optional) An array of states to filter the memberships by.
+   */
+  public function getUserGroupsByRoles(AccountInterface $user, array $roles, bool $require_all = FALSE, array $states = [OgMembershipInterface::STATE_ACTIVE]);
+
+  /**
+   * Returns an array of groups ids filtered by the og roles of the user.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $user
+   *   The user to get the groups for.
+   * @param array $roles
```
pfrenssen: `\Drupal\og\OgRoleInterface[]`.

src/OgRoleManager.php
```diff
@@ -132,6 +132,35 @@ public function getRolesByBundle($entity_type_id, $bundle) {
     return $this->ogRoleStorage()->loadByProperties($properties);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getRolesByPermissions(array $permissions, $entity_type_id = NULL, $bundle = NULL, $require_all = FALSE) {
```
pfrenssen: Let's default to `TRUE` here as well, this is more secure. If a developer forgets to set this flag at least it won't suddenly expose sensitive data to people who only have 1 permission out of an expected set of permissions.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
...
+   *   A list of og role objects to filter by.
+   * @param bool $require_all
+   *   (optional) If set to true, all requested roles must be present to return
+   *   the group.
+   * @param array $states
+   *   (optional) An array of states to filter the memberships by.
+   */
+  public function getUserGroupsIdsByRoles(AccountInterface $user, array $roles, bool $require_all = FALSE, array $states = [OgMembershipInterface::STATE_ACTIVE]);
```
pfrenssen: Let's rename this to `getUserGroupIdsByRoles()` to be consistent with the other methods in the class.

pfrenssen (changes requested):
It's getting there! I still have some remarks about the flags that are being added and some nitpicks.

I also made some additional comments while trying to understand what the test was doing. I have committed these in the branch.

src/MembershipManager.php
idimopoulos: The problem with that is that it is almost unavoidable to have conflicts between roles. Even with the current state, you might want to say, I want all users that can see node with ID 3.
This can be, the author (which does not matter in this example), UID 1 (also does not matter), users that can access content, users that can bypass content access, users that have the administer content permission.
So that would mean that we are requesting either an author, a moderator, an owner of the group, or a viewer.
I think most cases will derive anyway from the full PR which will be a scenario like "I want all users with either of those permissions". This will escalate to "I want all users with either of these roles that have one of the requested permissions". And frankly, most cases will be "Give me all users whose roles have this permission". So I still believe that having it default to false is still better. But if you insist, we can go on with your solution.

I agree with the renaming due to ambiguity.

src/OgRoleManager.php
idimopoulos: Please, check my comment in the previous similar remark. Again, if you insist, we will go with default to true. However, it is more of a common use case than a security issue. How many times you request a user that has more than one specific permission? Even the access handler has one operation when checking. 
In general, when you are requesting users based on permissions, the common case would be that you would want to request all users that can perform a single action, so either "access content" or "administer content" would not be a security leak when you are requesting users that can view the content.

src/MembershipManagerInterface.php
idimopoulos: You know I was struggling with it :P I changed it 3 times and all of them was adding and removing 's's around.. :P 
I could not think of which is the best case. The current name is get user groups ids. That must mean that you get the ID from each group, so from many groups, you get the ID. So many groups but what you get back is many IDs. Which hides behind it the `get user groups' ids`. So it gets down to `getUserGroupsIds[...]`. But ok, I sill change it.

src/MembershipManager.php
```diff
@@ -126,15 +126,15 @@ public function getMembership(EntityInterface $group, AccountInterface $user, ar
   /**
    * {@inheritdoc}
    */
-  public function getUserGroupsIdsByRoles(AccountInterface $user, array $roles, bool $require_all = FALSE, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
+  public function getUserGroupsIdsByRoles(AccountInterface $user, array $roles, array $states = [OgMembershipInterface::STATE_ACTIVE], bool $require_all_roles_roles = FALSE) {
```
zerolab: duplicate "roles" (`$require_all_roles_roles` ➜ `$require_all_roles`)

src/MembershipManagerInterface.php
pfrenssen: If you want to know why we chose this pattern, this is a collection of "Group IDs", not a collection of "IDs of groups" :)

Or: 1 group ID, 2 group IDs, ...

src/MembershipManager.php
idimopoulos: Ouch, sorry for that. Fixed.

src/MembershipManager.php
pfrenssen: I understand your point, but view it from this angle: in Drupal users are related to roles, not directly to permissions. User roles are intended to solve business use cases for people in certain positions. A role represents something like "The people in the HR department should be able to update the data of the employees". It is not just an intermediate storage for permissions.

Your argumentation is revolving around permissions and not roles. This is completely understandable as your main focus point right now is to solve your immediate use case, but you should considering the wider use cases of people consuming the OG API. This method will return all the group IDs that correspond to passed in roles. It has nothing to do with permissions. This method is `getGroupIdsByRoles()` not `getGroupIdsByPermissions()`.

If you imagine the use cases for retrieving the groups that are associated with a number of passed in roles then it makes sense that the most common and safer option would be to default to requiring all. If a developer wants to use this method to display sensitive data to users who have multiple roles in the organisation, then having this default to FALSE makes it a security risk. If the developer forgets to toggle the flag then the sensitive information is exposed to all users who have at least one of the roles. If we default to TRUE on the other hand and the developer wants to show information to users that have one or more of the passed in roles, then it is not a risk: if the developer forgets to toggle the flag then the info will be limited to only the users that have all the passed in roles.

src/OgRoleManager.php
pfrenssen: Yeah passing only a single permission in would also be a valid approach for this method. That is also safe. Multiple permissions which default to `AND` is also safe. However the proposed way of requesting multiple permissions which default to `OR` is NOT safe.

I will try to explain what I mean with the security risk using your own example: requesting roles by passing in both `access content` and `administer content`:

**Use case 1: `OR`**

We need to show a page with sensitive information to users that have either the `access content` OR the `administer content` permissions. This means we should set the flag `$require_all = FALSE`.

What happens if the default is `TRUE` and we forget to set the flag? The sensitive content is not shown to users which only have one of the permissions. The page is only shown to people that have both. This is an annoying bug but not a security risk: the sensitive content is not exposed unintentionally to unprivileged users.

**Use case 2: `AND`**

We need to show a page with sensitive information to users that have both the `access content` AND the `administer content` permissions. This means we should set the flag `$require_all = TRUE`.

What happens if the default is `FALSE` and we forget to set the flag? The sensitive content is now exposed unintentionally to the users that have only one of the permissions. This is a security issue.

src/MembershipManager.php
pfrenssen: Solved here but still a few instances of this in the parameter documentation below. I'll fix.

src/MembershipManager.php
```diff
@@ -413,6 +432,24 @@ protected function prepareConditionArray(array $value, array $default = NULL) {
     return array_unique($value);
   }
 
+  /**
+   * Loads the entities of a associative array of entity ids.
+   *
+   * @param array $group_ids
+   *   A structured array of entity keys indexed by their entity type id.
```
pfrenssen: Let's use the 'associative array' naming here too.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
...
+   *   The user to get the groups for.
+   * @param \Drupal\og\OgRoleInterface[] $roles
+   *   A list of og role objects to filter by.
+   * @param array $states
+   *   (optional) An array of states to filter the memberships by.
+   * @param bool $require_all_roles
+   *   (optional) If set to true, all requested roles must be present to return
+   *   the group.
+   */
+  public function getUserGroupsByRoles(AccountInterface $user, array $roles, array $states = [OgMembershipInterface::STATE_ACTIVE], bool $require_all_roles = FALSE);
```
pfrenssen: In the time since this PR was started we have decided to adopt PHP 7 type hints, so we can add the return type as well.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
...
+   *
+   * @param \Drupal\Core\Session\AccountInterface $user
+   *   The user to get the groups for.
+   * @param \Drupal\og\OgRoleInterface[] $roles
+   *   A list of og role objects to filter by.
+   * @param array $states
+   *   (optional) An array of states to filter the memberships by.
+   * @param bool $require_all_roles
+   *   (optional) If set to true, all requested roles must be present to return
+   *   the group.
```
pfrenssen: Return type is not documented.

src/MembershipManagerInterface.php
```diff
@@ -55,6 +55,36 @@ public function getUserGroupIds(AccountInterface $user, array $states = [OgMembe
...
+   */
+  public function getUserGroupIdsByRoles(AccountInterface $user, array $roles, array $states = [OgMembershipInterface::STATE_ACTIVE], bool $require_all_roles = FALSE);
```
pfrenssen: Here too, the return type is not documented and we can now add the PHP 7 return type hint.

tests/src/Kernel/Entity/OgRoleTest.php
```diff
@@ -243,6 +243,76 @@ public function testRoleCreate() {
     }
   }
 
+  /**
+   * Tests searching roles by a given permission list.
+   */
+  public function testLoadRoleByPermissions() {
```
pfrenssen: This is actually testing code on the `OgRoleManager` so this test should be moved to `OgRoleManagerTest`.

pfrenssen (changes requested):
This is looking close to completion. I just have some nitpicks and I insist on changing the flags to default to `TRUE` so that our public API is safe by default.

I will keep this assigned to me to make these small changes, then this is ready for merging after tests are green.

pfrenssen (approved):

pfrenssen:
I have switched the polarity of the flags as explained above, moved the test to the right class, and did an overhaul of the documentation. This is looking good now to me. Thanks a lot for this helpful addition!

The test on codeclimate has failed but this is still in experimental stage. All tests on Travis have succeeded.

PR #483, opened 2019-04-29
Add a cache tag to invalidate a group's user membership listings

### LINKED ISSUE #472: Provide an `og-group-member-list` cache tag that is invalidated when a membership changes

pfrenssen:
Fixes #472.

This also includes a member count block that demonstrates the use of the cache tag, but is fully themeable and usable in production sites to display member counts in group pages.

### DIFF
--- a/og.module
+++ b/og.module
@@ -304,6 +304,37 @@ function og_entity_type_alter(array &$entity_types) {
 }
 
 /**
+ * Implements hook_theme().
+ */
+function og_theme($existing, $type, $theme, $path) {
+  return [
+    'og_member_count' => [
+      'variables' => [
+        'count' => 0,
+        'group' => NULL,
+      ],
+    ],
+  ];
+}
+
+/**
+ * Prepares variables for member count templates.
+ *
+ * Default template: og-member-count.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - count: An integer representing the number of members in the group.
+ *   - group: The group, which is a content entity.
+ */
+function template_preprocess_og_member_count(array &$variables) {
+  /** @var \Drupal\Core\Entity\ContentEntityInterface $group */
+  $group = $variables['group'];
+
+  $variables['group_label'] = $group->label();
+}
+
+/**
  * Invalidates group content cache tags for the groups this entity belongs to.
  *
  * @param \Drupal\Core\Entity\EntityInterface $entity
--- a/og.module
+++ b/og.module
@@ -311,6 +311,7 @@ function og_theme($existing, $type, $theme, $path) {
     'og_member_count' => [
       'variables' => [
         'count' => 0,
+        'membership_states' => [],
         'group' => NULL,
       ],
     ],
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\og\Entity;
 
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityStorageInterface;
@@ -480,6 +481,37 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
     \Drupal::service('og.access')->reset();
 
     return $result;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function invalidateTagsOnSave($update) {
+    parent::invalidateTagsOnSave($update);
+
+    // A membership was created or updated: invalidate the membership list cache
+    // tags of its group. An updated membership may start to appear in a group's
+    // membership listings because it now meets those listings' filtering
+    // requirements. A newly created membership may start to appear in listings
+    // because it did not exist before.
+    $tags = Cache::buildTags('og-group-membership-list', $this->getGroup()->getCacheTagsToInvalidate());
+    Cache::invalidateTags($tags);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static function invalidateTagsOnDelete(EntityTypeInterface $entity_type, array $entities) {
+    parent::invalidateTagsOnDelete($entity_type, $entities);
+
+    // A membership was deleted: invalidate the membership list cache tags of
+    // its group membership lists, so that any lists that contain the membership
+    // will be recalculated.
+    $tags = [];
+    foreach ($entities as $entity) {
+      $tags = Cache::mergeTags(Cache::buildTags('og-group-membership-list', $entity->getGroup()->getCacheTagsToInvalidate()), $tags);
+    }
+    Cache::invalidateTags($tags);
   }
 
   /**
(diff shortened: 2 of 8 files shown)


src/Entity/OgMembership.php
```diff
@@ -494,8 +494,10 @@ protected function invalidateTagsOnSave($update) {
     // membership listings because it now meets those listings' filtering
     // requirements. A newly created membership may start to appear in listings
     // because it did not exist before.
-    $tags = Cache::buildTags('og-group-membership-list', $this->getGroup()->getCacheTagsToInvalidate());
-    Cache::invalidateTags($tags);
+    if ($group = $this->getGroup()) {
+      $tags = Cache::buildTags('og-group-membership-list', $group->getCacheTagsToInvalidate());
```
amitaibu: Maybe use a constant instead of `'og-group-membership-list'`?

src/Entity/OgMembership.php
```diff
@@ -482,6 +483,41 @@ public function save() {
...
+   */
+  protected function invalidateTagsOnSave($update) {
+    parent::invalidateTagsOnSave($update);
+
+    // A membership was created or updated: invalidate the membership list cache
+    // tags of its group. An updated membership may start to appear in a group's
+    // membership listings because it now meets those listings' filtering
+    // requirements. A newly created membership may start to appear in listings
+    // because it did not exist before.
+    if ($group = $this->getGroup()) {
```
MPParsley: There's no real consensus on this but I'm not a fan of assignments in conditionals
See https://www.drupal.org/project/coding_standards/issues/2907332 & https://softwareengineering.stackexchange.com/questions/334910/are-assignments-in-the-condition-part-of-conditionals-a-bad-practice

og.module
```diff
@@ -303,6 +303,37 @@ function og_entity_type_alter(array &$entity_types) {
   }
 }
 
+/**
+ * Implements hook_theme().
+ */
+function og_theme($existing, $type, $theme, $path) {
+  return [
+    'og_member_count' => [
+      'variables' => [
```
MPParsley, line 312: Definitely not needed but we _could_ also add the flags for the types we're counting to the template (i.c. blocked pending users).

MPParsley (approved):

src/Entity/OgMembership.php
pfrenssen: OK! I will move the assignment out of the `if` statement.

og.module
pfrenssen: OK this could be handy to generate some CSS classes for group administrators. Will add them to the data that is passed to the theme layer.

src/Plugin/Block/MemberCountBlock.php
```diff
@@ -0,0 +1,166 @@
...
+      $container->get('og.context'),
+      $container->get('og.membership_manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
```
claudiu-cristea, line 81: We always have to account the parent return value, even in this case that returns an empty array. This is because the parent might return something in the future or at some point we switch to other base class:

```
return [
  ...
] + parent:: defaultConfiguration();
```

og.module
```diff
@@ -303,6 +303,38 @@ function og_entity_type_alter(array &$entity_types) {
...
+  ];
+}
+
+/**
+ * Prepares variables for member count templates.
+ *
+ * Default template: og-member-count.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
```
claudiu-cristea, line 327: The array contains also the `membership_states` key but this is not detailed below.

og.module
```diff
@@ -303,6 +303,38 @@ function og_entity_type_alter(array &$entity_types) {
...
+ *   - count: An integer representing the number of members in the group.
+ *   - group: The group, which is a content entity.
+ */
+function template_preprocess_og_member_count(array &$variables) {
```
claudiu-cristea, line 331: I don't see the reason of this preprocess. We can pass the label directly from the block plugin.

templates/og-member-count.html.twig
```diff
@@ -0,0 +1,22 @@
+{#
+/**
+ * @file
+ * Default theme implementation to show the member count of a group.
+ *
+ * Available variables:
+ * - count: The number of members in the group.
+ * - membership_states: An array of membership states included in the count.
+ * - group_label: The group label.
```
claudiu-cristea, line 9: The `group` variable is also available as we don't unset it in `template_preprocess_og_member_count()`. If we don't need this variable, we probably want to pass the label from the beginning, in the block plugin and remove the template preprocess function. Or we add here the `group` too.

tests/src/Kernel/Plugin/Block/MemberCountBlockTest.php
```diff
@@ -0,0 +1,372 @@
...
+   * Checks that the block shows the correct member count for the given group.
+   *
+   * @param int $group_key
+   *   The key of the group for which to check the block.
+   * @param int $expected_count
+   *   The number of members that are expected to be shown in the block.
+   */
+  protected function assertMemberCount($group_key, $expected_count) {
+    $expected_string = (string) $this->formatPlural($expected_count, '@label has 1 member.', '@label has @count members', ['@label' => $this->groups[$group_key]->label()]);
+    $this->assertTrue(strpos($this->renderBlock($group_key), $expected_string) !== FALSE);
```
claudiu-cristea, line 276: This can be written in more meaningful way:

```
$this->assertContains($expected_string, (string) $this->renderBlock($group_key));
```

tests/src/Kernel/Plugin/Block/MemberCountBlockTest.php
```diff
@@ -0,0 +1,372 @@
...
+    $this->activeGroup = $original_active_group;
+  }
+
+  /**
+   * Callback providing the active group to be returned by the mocked OgContext.
+   *
+   * @return \Drupal\entity_test\Entity\EntityTest
+   *   The active group.
+   */
+  public function getActiveGroup() {
```
claudiu-cristea, line 368: ~This method is not used at all.~ EDIT: Sorry, I see now where is used.

claudiu-cristea (commented):
I know this is already merged. However, I still detected some points

MPParsley:
@claudiu-cristea, please review this PR: https://github.com/Gizra/og/pull/486

claudiu-cristea:
@MPParsley, thank you. I just did a review. I see only one aspect.

og.module
pfrenssen: This is intended to demonstrate how and end user would change this block into something useful for them.

PR #484, opened 2019-05-01
Optimize performance of og_get_all_group_entity()

jorrit:
For sites with many entity types and bundles `og_get_all_group_entity()` becomes unnecessarily slow because of the loops over entity types and bundles.
The function uses `og_is_group_type()` to determine whether a bundle is a group type. This function checks the bundles of field `OG_GROUP_FIELD`.
With this change `og_get_all_group_entity()` reads the bundles of `OG_GROUP_FIELD` directly, which is much faster.

### DIFF
--- a/og.module
+++ b/og.module
@@ -2490,15 +2490,17 @@ function og_is_group_content_type($entity_type, $bundle_name) {
 function og_get_all_group_entity() {
   $return = array();
 
-  foreach (entity_get_info() as $entity_type => $entity_value) {
-    foreach ($entity_value['bundles'] as $bundle => $bundle_value) {
-      if (og_is_group_type($entity_type, $bundle)) {
-        $return[$entity_type] = check_plain($entity_value['label']);
-        // At least one bundle of the entity can be a group, so break.
-        break;
-      }
-    }
+  $field = field_info_field(OG_GROUP_FIELD);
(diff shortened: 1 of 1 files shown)


MPParsley:
Makes sense but there seems to be an issue with the tests:
`The command "! egrep -i "([1-9]+ fail)|(Fatal error)|([1-9]+ exception)" /tmp/simpletest-result.txt" exited with 1.
`

jorrit:
Thanks for your comments @MPParsley. I have fixed the exceptions by adding a check if the `OG_GROUP_FIELD` field exists.

og.module
```diff
@@ -2490,15 +2490,18 @@ function og_is_group_content_type($entity_type, $bundle_name) {
...
-        break;
-      }
-    }
+  $field = field_info_field(OG_GROUP_FIELD);
+  if (empty($field['bundles'])) {
+    return $return;
+  }
+
+  $entity_types = array_keys($field['bundles']);
+
```
```suggestion
```

MPParsley (approved):
Look fine but this would need review by someone using of for D7, @amitaibu ?

og.module
jorrit: I have removed the blank line.

amitaibu (approved):
:+1:

PR #486, opened 2019-05-02
472 Fixes

MPParsley:
Added feedback from #483.

### DIFF
--- a/og.module
+++ b/og.module
@@ -313,6 +313,7 @@ function og_theme($existing, $type, $theme, $path) {
         'count' => 0,
         'membership_states' => [],
         'group' => NULL,
+        'group_label' => NULL,
       ],
     ],
   ];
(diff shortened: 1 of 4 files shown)


og.module
```diff
@@ -326,6 +326,7 @@ function og_theme($existing, $type, $theme, $path) {
  * @param array $variables
  *   An associative array containing:
  *   - count: An integer representing the number of members in the group.
+ *   - membership_states: The membership states that are taken into account.
  *   - group: The group, which is a content entity.
  */
 function template_preprocess_og_member_count(array &$variables) {
```
claudiu-cristea: We don't need the `template_preprocess_og_member_count()` anymore. We already have the `group_label` passed in the block plugin.

EDIT: And we should add the `group_label` also in the `hook_theme()` implementation.

claudiu-cristea (changes requested):

og.module
MPParsley: Indeed, I was going to remove the preprocess but forgot.
PS: you can add inline code suggestions, this makes reviews a lot easier.

claudiu-cristea (approved):

pfrenssen (approved):

PR #492, opened 2019-05-08
Add slack to support channels

MPParsley:
There's now an #og channel on drupal.slack.com
See slack://channel?team=T06GX3JTS&id=CELR48EA0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "support": {
         "issues": "https://drupal.org/project/issues/og",
         "irc": "irc://irc.freenode.org/drupal-og",
-        "slack": "slack://channel?team=T06GX3JTS&id=CELR48EA0",
+        "slack": "https://drupal.slack.com/messages/CELR48EA0",
         "source": "https://git.drupalcode.org/project/og"
     },
     "require": {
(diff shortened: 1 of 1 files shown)


pfrenssen:
Why not use the HTTP link to the chat page? https://drupal.slack.com/messages/CELR48EA0/

This way it is accessible also by people who choose not to install the Slack application (which is not open source).

MPParsley:
> Why not use the HTTP link to the chat page? https://drupal.slack.com/messages/CELR48EA0/

Good point. The reason I used the slack protocol was:
- because we can (not so much though :-) )
- seems like we do the same for irc (where we could also add a link to freenode instead)

edit: you're right it's not open source, I added the http link.

composer.json
```diff
@@ -7,6 +7,7 @@
     "support": {
         "issues": "https://drupal.org/project/issues/og",
         "irc": "irc://irc.freenode.org/drupal-og",
+        "slack": "slack://channel?team=T06GX3JTS&id=CELR48EA0",
```
MPParsley: @pfrenssen feel free to change it:
```suggestion
        "slack": "https://drupal.slack.com/messages/CELR48EA0",
```

pfrenssen (approved):

PR #493, opened 2019-05-10
Use strict types in the OgMembership entity

pfrenssen:
Since we are requiring PHP 7.1 and higher we can start leveraging the new security features of PHP 7.

This PR proposes to use strict typing in OgMembership entities. This will ensure that our code will not be subject to typing bugs. Unfortunately this effort is limited to our internal protected methods and public methods that we define ourselves; the Drupal interfaces are still supporting PHP 5 and do not define return types and scalar type hints.

Still, this will still enhance security and it will be interesting to see if we get any test failures which point to bugs in our code or test coverage.

### DIFF
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -184,14 +184,10 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   /**
    * {@inheritdoc}
    */
-  public function getGroup(): ContentEntityInterface {
+  public function getGroup(): ?ContentEntityInterface {
     assert(!empty($this->get('entity_type')->value) || !empty($this->get('entity_id')->value), new \LogicException(__METHOD__ . '() should only be called on loaded memberships, or on newly created memberships that already have the group set.'));
     $entity_type = $this->get('entity_type')->value;
     $entity_id = $this->get('entity_id')->value;
-
-    if (empty($entity_type) || empty($entity_id)) {
-      return NULL;
-    }
 
     return \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
   }
--- a/src/Entity/OgMembership.php
+++ b/src/Entity/OgMembership.php
@@ -1,10 +1,12 @@
 <?php
 
+declare(strict_types = 1);
+
 namespace Drupal\og\Entity;
 
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\ContentEntityBase;
-use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityStorageInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
@@ -80,14 +82,14 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   /**
    * {@inheritdoc}
    */
-  public function getCreatedTime() {
+  public function getCreatedTime(): int {
     return $this->get('created')->value;
   }
 
   /**
    * {@inheritdoc}
    */
-  public function setCreatedTime($timestamp) {
+  public function setCreatedTime(int $timestamp): OgMembershipInterface {
     $this->set('created', $timestamp);
     return $this;
   }
@@ -125,7 +127,7 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   /**
    * {@inheritdoc}
    */
-  public function setGroup(EntityInterface $group) {
+  public function setGroup(ContentEntityInterface $group): OgMembershipInterface {
     $this->set('entity_type', $group->getEntityTypeId());
     $this->set('entity_bundle', $group->bundle());
     $this->set('entity_id', $group->id());
@@ -135,21 +137,21 @@ class OgMembership extends ContentEntityBase implements OgMembershipInterface {
   /**
    * {@inheritdoc}
    */
-  public function getGroupEntityType() {
+  public function getGroupEntityType(): string {
     return $this->get('entity_type')->value;
   }
 
   /**
    * {@inheritdoc}
    */
-  public function getGroupBundle() {
+  public function getGroupBundle(): string {
     return $this->get('entity_bundle')->value;
   }
 
   /**
    * {@inheritdoc}
    */
-  public function getGroupId() {
+  public function getGroupId(): string {
     return $this->get('entity_id')->value;
   }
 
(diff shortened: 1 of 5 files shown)


pfrenssen:
There are a couple of failures. Most are related to methods like `::getGroup()` and `::getGroupBundle()` returning `NULL` when this method is called on a newly created `OgMembership` entity which does not have a group set on it.

This is ambiguous at the moment. Our interfaces declare that the value will always be returned:

```
  /**
   * Gets the group entity bundle.
   *
   * @return string
   *   The bundle.
   */
  public function getGroupBundle();
```

But on the other hand we have tests that check that the method returns `NULL` when the group is not yet set on a newly created entity:

```
  /**
   * Tests getting the bundle of the group that is associated with a membership.
   *
   * @covers ::getGroupBundle
   */
  public function testGetGroupBundle() {
    $membership = OgMembership::create();

    // When no group has been set yet, the method should return NULL.
    $this->assertNull($membership->getGroupBundle());
  }
```

We also have some methods declared on the interface that are declaring that the return type is optional:

```
  /**
   * Gets the group associated with the membership.
   *
   * @return \Drupal\Core\Entity\ContentEntityInterface|null
   *   The group object which is referenced by the membership, or NULL if no
   *   group has been set yet.
   */
  public function getGroup();
```

We should solve this ambiguity. I think it makes most sense for this methods to be called only on saved entities. Most of the time when we are creating a new `OgMembership` entity we will know which group, state and roles needs to be set on it, otherwise we wouldn't even be able to save the entity.

On the other hand, if we would declare these return values to be optional we are required to check them every single time we want to retrieve the group, state or roles from a saved membership, which doesn't make much sense since these values are always present on a valid entity.

So with optional return values our code would have to look like this:

```
$membership = $this->membershipStorage->load($membership_id);

$group = $membership->getGroup();

if (empty($group)) {
  // Do nothing.
  // This code will never be executed since a saved membership always has a group.
  // However if we do not handle this case our code would not pass static analysis,
  // since the interface declares that the method possibly returns NULL.
}
```

By requiring the method to always return the value we would make the 90% use case much easier to use.

I would also propose  to add `assert()` statements to all these methods to give a helpful error message to developers.

src/Entity/OgMembership.php
```diff
@@ -125,7 +127,7 @@ public function setOwnerId($uid) {
   /**
    * {@inheritdoc}
    */
-  public function setGroup(EntityInterface $group) {
+  public function setGroup(ContentEntityInterface $group): OgMembershipInterface {
```
MPParsley, line 132: I assume config can never be a Group. Note that in core there's an initiative to support other entities than content and config.

See https://www.drupal.org/project/drupal/issues/3042467

src/Entity/OgMembership.php
```diff
@@ -96,13 +98,15 @@ public function setCreatedTime($timestamp) {
    * {@inheritdoc}
    */
   public function getOwner() {
+    assert(!empty($this->get('uid')->entity), new \LogicException(__METHOD__ . '() should only be called on loaded memberships, or on newly created memberships that already have the owner set.'));
```
MPParsley, line 101: We could refactor this assertion in a private or protected method that takes the method and the schema key as arguments?

pfrenssen:
OK I started addressing the failures and discovered that my reasoning is not applicable in the case of `OgMembership::getGroup()`. The group is stored as an entity reference on the `OgMembership` entity and is loaded from the entity storage before being returned. However it is always possible that the group has been correctly set at the time the membership was created, but in the meantime has been removed from the entity storage. `OgMembership` will not know about this.

So it is always possible that `OgMembership::getGroup()` returns `NULL` because it depends on the external entity storage. I will update the signature of this one to match this behaviour. All the other methods are in our control and we can always return the data.

MPParsley:
I like where this is going.

Note that strict typing will introduce some work in #231 (beta blocker) as it adds a method to `OgMembership`.

pfrenssen:
Thanks for pointing this out. I see that #231 is in full swing right now. I am in favour of holding this off until that important PR is finished, so that it doesn't interrupt the work.

src/Entity/OgMembership.php
```diff
@@ -170,50 +177,47 @@ public function getGroupId() {
...
   /**
    * {@inheritdoc}
    */
-  public function getGroup() {
+  public function getGroup(): ?ContentEntityInterface {
+    assert(!empty($this->get('entity_type')->value) || !empty($this->get('entity_id')->value), new \LogicException(__METHOD__ . '() should only be called on loaded memberships, or on newly created memberships that already have the group set.'));
     $entity_type = $this->get('entity_type')->value;
     $entity_id = $this->get('entity_id')->value;
 
-    if (empty($entity_type) || empty($entity_id)) {
```
MPParsley, line 184: Shouldn't we keep this?

pfrenssen:
Yeah sure I'm going to finish this, but we can wait with merging this until #231 is done. It will be easier to adapt this PR to match #231 than the other way around.

pfrenssen:
And it's green! :smile:

pfrenssen:
Hmm actually I had a look over at #231 and it seems to still require some effort, especially in the writing of tests. I would like to prepare a release for OG this weekend so that people who are using it can update to Drupal 8.7. We could still include this in that release.

Since strict types is a new development paradigm that is not very common in Drupal code I would like to have sign off from @amitaibu on this. What do you think, Amitai? :)

MPParsley (approved):

MPParsley:
> Hmm actually I had a look over at #231 and it seems to still require some effort, especially in the writing of tests. I would like to prepare a release for OG this weekend so that people who are using it can update to Drupal 8.7. We could still include this in that release.

See my remark on that issue wrt tests #cursing-in-church :-)

> Since strict types is a new development paradigm that is not very common in Drupal code I would like to have sign off from @amitaibu on this. What do you think, Amitai? :)

We're using strict typing for a while now, the only issue I might see here (more in general) would be if core interfaces will implement a different signature. For example EntityOwnerInterface might decide to return an object instead of an integer.

src/Entity/OgMembership.php
pfrenssen: No this case is never going to happen since the `assert()` is doing the same check and throwing an exception before this line is executed.

The only reason this should ever return `NULL` is if the entity storage deletes the group before the membership is deleted. This was happening in `OgMembershipManager::invalidateTagsOnDelete()`, which was one of the failing tests.

... And now that I am writing this I realize that this might actually be expected behaviour. An OgMembership entity _depends_ on the group. If the group is deleted then the membership _should_ be deleted alongside it.

It is possible this use case is not yet tested. I found a test case that checks that the memberships are deleted when a user is deleted (`\Drupal\Tests\og\Kernel\Entity\GetMembershipsTest::testOrphanedMembershipsDeletion()`) but could not find one for when the group is deleted.

I will investigate this case, write a test and see what happens. It might be that we can change this signature to be always returning a membership after all.

pfrenssen:
I looked into the deletion of orphaned memberships if the group is deleted. They are handled by the `OgDeleteOrphans` plugins. In case a site is configured do delete orphans in either a batch process or on cron jobs, any calls to `OgMembership::getGroup()` might return `NULL` until the batch / cron job is completed. So we need to keep it optional. I will update the documentation to clarify this.

amitaibu:
I'm in favor of extending the type hinting. 

Also , I'm not sure if we need to wait for #231 , as that might take a while, so better get this one in once it's ready.

pfrenssen:
Great!! I think this will benefit us and help make OG more stable!

pfrenssen:
I guess this is still good to go, it has only some documentation changes after being approved.

PR #243, opened 2017-03-07
Group roles and permissions UI

zerolab:
This PR is a continuation of amitaibu/og#196. It:

- Fixes the clashing routes for adding a role
- Fixes new group roles considered required, rather than standard
- Adds a role deletion form
- Adds a group permissions and a group role permissions form

To-Do:

- [x] clean-up the routing
- [x] use separate permissions for roles (`og_role.add`, `og_role.update`, `og_role.delete`)
- [x] allow permissions groups other that "Group" and "Group content"
- [ ] tests

Also needs decision as to when amitaibu/og#315 comes into force.

PR #497, opened 2019-05-11
Group roles ui

MPParsley:
Let's move the work from https://github.com/amitaibu/og/pull/196 here.


MPParsley:
See https://github.com/Gizra/og/pull/243

Issue #487, opened 2019-05-06
Performance hit in OgRoleCacheContext


pfrenssen:
`OgRoleCacheContext::getContext()` is slow for users that are a member of a large number of groups. This is critical since this code is running in order to calculate the cache IDs of items that vary by their group roles. This means that even on pages that are fully cached, this code needs to run in order to retrieve the correct items from the page cache or render cache. 

For users that are a member of hundreds or thousands of groups this is causing cached pages to load slowly.

Here is the code in question:

```
      foreach ($this->membershipManager->getMemberships($this->user) as $membership) {
        $role_names = array_map(function (OgRoleInterface $role) {
          return $role->getName();
        }, $membership->getRoles());
        if ($role_names) {
          $memberships[$membership->getGroupEntityType()][$membership->getGroupId()] = $role_names;
        }
      }
```

This code is doing multiple slow operations:

1. It loads the full `OgMembership` entities of all the user's memberships. This is a lot of data.
2. For each membership it calls `$membership->getRoles()` which returns the full `OgRole` entities for the membership. These are a lot of individual queries.

I investigated the code path and there are a number of improvements we can make to make this a lot faster, and also improve some parts that will make the OG API faster for our users:

- [ ] `OgMembership::getMemberships()` can be persistently cached. See #485.
- [x] `OgMembership::getRolesIds()` does not need to call `$this->getRoles()` in order to get the role IDs. These IDs are already present on the entity reference field. See #555.
- [x] Once the previous step is done, `OgRoleCacheContext::getContext()` should call `$membership->getRolesIds()` instead of `::getRoles()`. We can then get the role name from the ID, or even use the full ID. See #555.

This still leaves us with the problem that all the `OgMembership` entities are loaded. We could solve this by doing a `SELECT` query on the database that directly returns the data from the role field. This would be very fast, but would make OG incompatible with NoSQL databases like MongoDB / CouchDB / Virtuoso. This can still be solved by checking if the `OgMembership` database backend implements `SqlContentEntityStorage` and falling back to the inefficient loading of all memberships if needed.

amitaibu:
> but would make OG incompatible with NoSQL databases like MongoDB / CouchDB / Virtuoso
> This can still be solved by checking if the OgMembership database backend implements SqlContentEntityStorage and falling back to the inefficient loading of all memberships if needed.

[...]

PR #499, opened 2019-05-11
Test that an exception is thrown when a membership is saved using invalid roles

pfrenssen:
This is preparatory work for #487. In that issue it is proposed to refactor `OgMembership::getRolesIds()` but we only have basic test coverage for the relation of roles and memberships. Our current test coverage (`OgMembershipRoleReferenceTest`) hits all of the related methods at least once, but we are lacking systematic coverage with data providers that check all possible code paths. I would like to enhance our code coverage before attempting to refactor the current code.

This is a first PR in this effort, checking all the code paths in which saving a membership with invalid roles will result in an exception.

### DIFF
--- a/tests/src/Kernel/Entity/OgMembershipTest.php
+++ b/tests/src/Kernel/Entity/OgMembershipTest.php
@@ -3,6 +3,7 @@
 namespace Drupal\Tests\og\Kernel\Entity;
 
 use Drupal\entity_test\Entity\EntityTest;
+use Drupal\entity_test\Entity\EntityTestMul;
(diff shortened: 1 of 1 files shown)


pfrenssen:
This is failing because of #502.

pfrenssen:
It is green now on Travis. Codeclimate is still experimental.

MPParsley (approved):
The test looks good, thanks!

I flagged all default feedback from CodeClimate so the tick is now green.

Note: As long as we don't enable branch protection for CodeClimate you can still merge a PR even with feedback from CC.

PR #504, opened 2019-05-13
Fixed Uncaught Error: Cannot use object of type stdClass as array

MPParsley:
( ! ) Fatal error: Uncaught Error: Cannot use object of type stdClass as array in web/modules/contrib/og/src/GroupTypeManager.php on line 268
--

In #450 we refactored the use `Drupal\Core\State\StateInterface` to `Drupal\Core\Cache\CacheBackendInterface` but CacheBackendInterface::get() returns an object with a data property (not an array).

```
object(stdClass)[6055]
  public 'cid' => string 'og.group_manager.group_relation_map' (length=35)
  public 'data' => 
    array (size=1)
      'node' => 
        array (size=1)
          'subsite' => 
            array (size=2)
              ...
  public 'created' => string '1557745337.212' (length=14)
  public 'expire' => string '-1' (length=2)
  public 'serialized' => string '1' (length=1)
  public 'tags' => 
    array (size=0)
      empty
  public 'checksum' => string '0' (length=1)
  public 'valid' => boolean true
```

### DIFF
--- a/src/GroupTypeManager.php
+++ b/src/GroupTypeManager.php
@@ -371,7 +371,7 @@ class GroupTypeManager implements GroupTypeManagerInterface {
   protected function refreshGroupRelationMap() {
     // Retrieve a cached version of the map if it exists.
     if ($group_relation_map = $this->cache->get(self::GROUP_RELATION_MAP_CACHE_KEY)) {
-      $this->groupRelationMap = $group_relation_map;
+      $this->groupRelationMap = $group_relation_map->data;
       return;
     }
 
--- a/src/GroupTypeManager.php
+++ b/src/GroupTypeManager.php
@@ -370,8 +370,8 @@ class GroupTypeManager implements GroupTypeManagerInterface {
    */
   protected function refreshGroupRelationMap() {
     // Retrieve a cached version of the map if it exists.
-    if ($group_relation_map = $this->cache->get(self::GROUP_RELATION_MAP_CACHE_KEY)) {
-      $this->groupRelationMap = $group_relation_map->data;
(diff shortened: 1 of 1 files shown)


src/GroupTypeManager.php
```diff
@@ -371,7 +371,7 @@ protected function refreshGroupMap() {
   protected function refreshGroupRelationMap() {
     // Retrieve a cached version of the map if it exists.
     if ($group_relation_map = $this->cache->get(self::GROUP_RELATION_MAP_CACHE_KEY)) {
```
```suggestion
    if ($data_cached = $this->cache->get(self::GROUP_RELATION_MAP_CACHE_KEY)) {
```

src/GroupTypeManager.php
```diff
@@ -371,7 +371,7 @@ protected function refreshGroupMap() {
...
-      $this->groupRelationMap = $group_relation_map;
+      $this->groupRelationMap = $group_relation_map->data;
```
```suggestion
      $this->groupRelationMap = $data_cached->data;
```

pj-claeys (approved):

pfrenssen:
The branch was failing because of a coding standards violation, let's see if merging in the latest 8.x-1.x will make it green.

This looks like a valid fix, but if this problem is real it means we have no test coverage for this method. We should add it so this does not regress in the future.

pfrenssen:
I had a look at our test coverage for the `GroupTypeManager` and it indeed seems that we are missing coverage for a couple of methods. I added these missing tests as tasks to #372.

In scope of this issue it seems the easiest way to test this bug is to add a test to `GetBundleByBundleTest` that calls `::getGroupContentBundleIdsByGroupBundle()` and checks that it returns cached data.

pfrenssen:
I wrote a test that shows this bug. I published this in #523. Closing this in favor of #523.

PR #505, opened 2019-05-13
Add CodeClimate

MPParsley:
[Added code coverage report to code climate](https://docs.codeclimate.com/docs/travis-ci-test-coverage)

Reports are available on https://codeclimate.com/github/Gizra/og/issues
- [x] Add linters from Drupal core (e.g. javascript, css)
- [ ] Check that administrators can dismiss code climate suggestions

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,69 @@
+version: "2"
+
+prepare:
+  fetch:
+  - url: "https://git.drupalcode.org/project/drupal/raw/8.7.x/.csslintrc"
+    path: ".csslintrc"
+  - url: "https://git.drupalcode.org/project/drupal/raw/8.7.x/core/.eslintrc.json"
+    path: ".eslintrc.json"
+  - url: "https://git.drupalcode.org/project/drupal/raw/8.7.x/core/.eslintignore"
+    path: ".eslintignore"
+
+checks:
+  argument-count:
+    enabled: false
+  complex-logic:
+    enabled: false
+  file-lines:
+    enabled: false
+  method-complexity:
+    enabled: false
+  method-count:
+    enabled: false
+  method-lines:
+    enabled: false
+  nested-control-flow:
+    config:
+      threshold: 16
+  return-statements:
+    config:
+      threshold: 16
+  similar-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+  identical-code:
+    config:
+      threshold: # language-specific defaults. an override will affect all languages.
+
+plugins:
+  phpcodesniffer:
+    enabled: true
+    config:
+      standard: "Drupal,DrupalPractice"
+      file_extensions: "php,inc,module,install"
+  phpmd:
+    enabled: false
+  csslint:
+    enabled: true
+  eslint:
+    enabled: true
+  duplication:
+    enabled: true
+    config:
+      languages:
+        php:
+        javascript:
+  fixme:
+    enabled: true
+  markdownlint:
+    enabled: true
+    issue_override:
+      severity: minor
+
+exclude_patterns:
+- "config/"
+- "translations/"
+- "**/vendor/**/*"
+- "**/.idea/**/*"
+- "**/tests/**/*"
+- "**/*.settings.php"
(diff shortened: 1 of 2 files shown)


pfrenssen:
It looks like many of these suggestions are more intended for developing simple OO applications but this doesn't match the requirements of a real life CMS application.

I had a look through the first two pages of suggestions and none of them looked like they would make a difference in the code quality for us.

Since we are dealing with UI elements and complex decision trees like determining access, we are always going to need methods that have more than 25 lines or files that have more than 250 lines.

We are using PHP_CodeSniffer because it is also endorsed by Drupal core. I think once Drupal core adopts CodeClimate then we can maybe consider using it too, but I don't see this as a good fit for our project.

pfrenssen:
OK so my previous comment was based on an early version of the tool which was using mostly default values. The setup is being finetuned by @MPParsley in #506 and it is now returning a whole bunch of good suggestions. I am changing my mind on this, it will be a valuable addition to our workflow!

amitaibu:
@AronNovak fyi ^^

pfrenssen:
I had a look at the latest results and the configuration and it seems that all the things that are useful are actually being reported by the `DrupalPractice` standard for PHP_CodeSniffer. Everything else is just noise.

So I am wondering, wouldn't it be a lot simpler to just add `DrupalPractice` to the current coding standards checks? This is a 1 line change that would give the same benefits, and doesn't require us to depend on another proprietary vendor.

MPParsley:
> So I am wondering, wouldn't it be a lot simpler to just add DrupalPractice to the current coding standards checks? This is a 1 line change that would give the same benefits, and doesn't require us to depend on another proprietary vendor.

- We can always decide to remove .codeclimate.yml, there's no vendor lock-in and it's free (for open source);
- CodeClimate monitors statistics and gives better insight in things like [technical debt](https://docs.codeclimate.com/docs/maintainability) and [churn](https://docs.codeclimate.com/docs/churn);
- Some checks have been disabled in #506 but in time they may still prove to be useful; Among these:
-- [Cognitive Complexity](https://docs.codeclimate.com/docs/cognitive-complexity)
-- [Cyclomatic Complexity](https://docs.codeclimate.com/docs/cyclomatic-complexity)
- Besides Codesniffer there are [a few useful plugins](https://docs.codeclimate.com/docs/list-of-engines) available like [FIXME](https://docs.codeclimate.com/docs/fixme), [ShellCheck](https://docs.codeclimate.com/docs/shellcheck), [markdownlint](https://docs.codeclimate.com/docs/markdownlint) and [Duplication](https://docs.codeclimate.com/docs/duplication). Some of them are opensource and new plugins might be developed in the future;
- Codesniffer configuration might be the same as on Travis but the flow is a lot more accessible if you compare them:

Travis:
travis output

GitHub:
codeclimate feedback

Should have put the [comment](https://github.com/Gizra/og/pull/506#issuecomment-493546158) from #506 in here.

You can [review the results of the .codeclimate.yml file on my fork](https://codeclimate.com/github/MPParsley/og).

Please read/review and let me know if there are still any outstanding questions.

pfrenssen:
The suggestions given are very good now, there are no more false positives.

Is there a way that we can source the PHP_CodeSniffer configuration from our own ruleset that we declare in `phpcs-ruleset.xml.dist`? There should not be any disparity between these, and if we can source this file then we can rest assured that if we make any changes to our coding standards then these will be picked up by codeclimate too.

For me personally this has not much added value since I tend to run my coding standards check before I commit and when I do have any coding standards violations I fix it by rebasing and force pushing rather than committing them one by one in the GitHub UI. But if you like it feel free to add it! For me the most important thing is that it does not cause any additional work for us as maintainers.

MPParsley:
We could add DrupalPractice to codesniffer in phpcs-ruleset.xml.dist but the we'd need to fix all issues first. For now CodeClimate will allow us to "confirm" issues and fix them in new issues. See #511.

Looking at the [documentation](https://github.com/codeclimate/codeclimate-phpcodesniffer) it would seem that CodeClimate is using the rule sets from https://github.com/klausi/coder.

Its possible to define our own rulesets as follows:
```
engines:
  phpcodesniffer:
    enabled: true
    config:
      standard: "phpcs-ruleset.xml.dist"
```
But then we'd have to make sure all the rules are available in our repo.

MPParsley:
Continuing the work in #623.

PR #511, opened 2019-05-20
Add DrupalPractice to codesniffer

MPParsley:


pfrenssen:
What we could do is to add all failing sniffs to an exclude list, and creating novice issues for each one. These can then be worked on, and as soon as a sniff is fixed we remove it from the exclude list so we will be detecting violations in the future.

pfrenssen:
This is fulfilled by #551

PR #516, opened 2019-06-10
Do not require to pass full OgRole objects when we only need the ID.

pfrenssen:
I stumbled upon this when looking into #487.

We have recently added the methods `MembershipManager::getUserGroupsByRoles()` and `MembershipManager::getUserGroupIdsByRoles()` but in fact we do not need to have the full role objects for this to work. The first thing that is done is to retrieve the role ID and then the role objects are no longer required.

We can simplify things by just requiring developers to pass in the role IDs. This would also save developers the effort of having to load the full `OgRole` objects just to pass them into this method.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -127,11 +127,7 @@ class MembershipManager implements MembershipManagerInterface {
   /**
    * {@inheritdoc}
    */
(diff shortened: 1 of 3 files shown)


pfrenssen:
I'm merging this in so it can be included in the alpha3 release.

PR #523, opened 2019-06-11
Uncaught Error: Cannot use object of type stdClass as array

pfrenssen:
This PR replaces #504 with a test first approach.

Original report:

Fatal error: Uncaught Error: Cannot use object of type stdClass as array in web/modules/contrib/og/src/GroupTypeManager.php on line 268

In #450 we refactored the use `Drupal\Core\State\StateInterface` to `Drupal\Core\Cache\CacheBackendInterface` but CacheBackendInterface::get() returns an object with a data property (not an array).

```
object(stdClass)[6055]
  public 'cid' => string 'og.group_manager.group_relation_map' (length=35)
  public 'data' => 
    array (size=1)
      'node' => 
        array (size=1)
          'subsite' => 
            array (size=2)
              ...
  public 'created' => string '1557745337.212' (length=14)
  public 'expire' => string '-1' (length=2)
  public 'serialized' => string '1' (length=1)
  public 'tags' => 
    array (size=0)
      empty
  public 'checksum' => string '0' (length=1)
  public 'valid' => boolean true
```

### DIFF
--- a/src/GroupTypeManager.php
+++ b/src/GroupTypeManager.php
@@ -370,8 +370,8 @@ class GroupTypeManager implements GroupTypeManagerInterface {
    */
   protected function refreshGroupRelationMap() {
     // Retrieve a cached version of the map if it exists.
-    if ($data_cached = $this->cache->get(self::GROUP_RELATION_MAP_CACHE_KEY)) {
-      $this->groupRelationMap = $data_cached->data;
+    if ($cached_map = $this->getCachedGroupRelationMap()) {
+      $this->groupRelationMap = $cached_map;
       return;
     }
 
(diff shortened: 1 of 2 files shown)


pfrenssen:
The test fails on the reported error message:

```
There was 1 error:

1) Drupal\Tests\og\Kernel\Entity\GetBundleByBundleTest::testGetGroupContentBundleIdsByGroupBundleUsesCachedData
Error: Cannot use object of type stdClass as array

/home/travis/build/Gizra/og/src/GroupTypeManager.php:254
/home/travis/build/Gizra/og/tests/src/Kernel/Entity/GetBundleByBundleTest.php:602
```

Now let's fix it by cherry-picking the solution from #504.

pfrenssen:
I cherry-picked the fix from #504 and split off the retrieval of cached data to a separate method so we can leverage PHP 7's return type declaration to alert us if we are returning a cache object rather than the expected array.

MPParsley (approved):

pfrenssen:
Merging, the tests were green, codeclimate is still experimental and failing on false positives.

PR #529, opened 2019-06-13
Consolidate the permissions regarding roles and permissions.

### LINKED ISSUE #526: Consider to unify permissions for managing roles and permissions in `administer permissions`
In Drupal 7 we had two separate permissions for managing roles and permissions: `manage roles` and `manage permissions`.

However these are very closely related since editing a role also allows the editing of the role's permissions, and a user with `manage permissions` can edit the permissions of all roles.

Both UIs are also very closely related, the role UI showing the permissions in a single column of checkboxes, while the permissions UI shows the same thing but in a matrix that covers all roles.

I would propose to do this in the same way as it is handled in Drupal core: with a single `administer permissions` permission that grants access to both.

pfrenssen:
Fixes #526


zerolab:
That looks good, but should there be a hook_update to prevent existing projects from tripping up?

pfrenssen:
We're in alpha so we are not yet considered to be stable and do not have to write update hooks. These permissions are also not yet used anywhere.

zerolab:
Fair enough.

Our project made use of at least one of them, so at least the release notes should include something for it, imho

MPParsley:
Perhaps we can start to [keep a changelog](https://keepachangelog.com/en/1.0.0/)? #531

pfrenssen:
I don't think changelogs are so useful since the invention of version control systems :) We already mention all the changes in the release notes.

MPParsley:
> I don't think changelogs are so useful since the invention of version control systems :) We already mention all the changes in the release notes.

Toh, just opened #532 :-) And yes, I noticed that all changes have been properly addressed in the releases on github. Not sure how this will evolve when contrib is to be maintained on gitlab though (if we ever move there)?

pfrenssen:
We did not make plans yet to move to Gitlab but this is something that is very interesting, especially if we can keep running our custom CI pipelines.

MPParsley (approved):
Not sure if a hook_update would have been sufficient since we're removing the old permission.

Anyhow alpha & not yet used so if we want to do this we'd better do it now.

PR #534, opened 2019-06-21
Don't add self referencing dependency when exporting og permissions

jorrit:
The function `og_features_permission_features_export()` prevents adding the module that provides a permission at the start of the function, but doesn't do so later on.


includes/og_features_permission.features.inc
```diff
@@ -33,7 +33,9 @@ function og_features_permission_features_export($data, &$export, $module_name =
       if (isset($permissions[$group_type][$bundle][$perm]['module'])) {
         // Add the module which defines the permission as a dependency.
         $module = $permissions[$group_type][$bundle][$perm]['module'];
-        $export['dependencies'][$module] = $module;
+        if ($module != $module_name) {
```
MPParsley:
```suggestion
        if ($module !== $module_name) {
```

MPParsley (approved):

PR #541, opened 2019-07-22
Do not require to pass the full user object when only the ID is required

pfrenssen:
This is part of #487 

Most of the methods in the `MembershipManager` service that deal with users require the full user object to be passed in, but only require the user ID to operate. This requires the calling code to needlessly load the full user entity from the database while they could simply pass in the entity ID.

This PR provides a B/C layer so that existing code keeps working for now.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -63,7 +63,7 @@ class MembershipManager implements MembershipManagerInterface {
    * {@inheritdoc}
    */
   public function getUserGroupIds($user_id, array $states = [OgMembershipInterface::STATE_ACTIVE]) {
-    if (!is_int($user_id)) {
+    if ($user_id instanceof AccountInterface) {
       trigger_error('Passing an account object is deprecated in og:8.1.0-alpha4 and is removed from og:8.1.0-beta1. Instead pass the user ID as an integer value. See https://github.com/Gizra/og/issues/542', E_USER_DEPRECATED);
       $user_id = $user_id->id();
     }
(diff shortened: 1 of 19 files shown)


src/MembershipManager.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types = 1);
```
MPParsley, line 3: Curious to see if this will break any beta blockers.

MPParsley (approved):
This looks good.

I suppose that for the beta1 release we'll add the following BC breaking change?

from:
```
  public function isMemberPending(EntityInterface $group, $user_id) {
```
 
into:
```
  public function isMemberPending(EntityInterface $group, int $user) {
```

pfrenssen:
Thanks for the review! I agree that we can make the change to the `int` type hint in beta1. I have updated the deprecation check to see if anything other than an integer is passed in, so that people are alerted in time that they need to update their code to pass integers. I also created an issue to handle the removal of the B/C layer: #542 - I mentioned there that we need to update the type hint too.

src/MembershipManager.php
pfrenssen: Me too, that's why I added it, in the current code base all tests are passing with strict types! :+1:

pfrenssen:
Wow we have a lot of cases where a string is returned from `$user->id()`, this was quite unexpected. I see that indeed `UserInterface::id()` is allowed to return strings, even though the ID has been defined as an integer in the base field definitions, and also `EntityOwnerInterface::getOwnerId()` will always return integers.

I am not sure now about requiring integers, this will be too much of a hassle for developers to always cast to an integer. Let's for now keep the documentation as integer, but not enforce it until core is more strict about its data types. I will remove the strict check on the B/C layer.

PR #559, opened 2019-07-26
Speed up retrieval of membership data by using a SELECT query

pfrenssen:
This is the final PR for #487. This replaces the loading of the (cached) `OgMembership` entities with a direct `SELECT` query if a SQL database is used for storing the entities. The original approach is kept as a fallback for NoSQL databases as was suggested in #487.

On my system this shaves off another ~60ms of the wall time for a user with 1140 groups since the full entities no longer need to be retrieved from cache. This brings the total time down to ~11ms.

This is built on top of #555 so includes all commits from that PR. This is blocked until that PR is merged.


pfrenssen:
I'm getting failures when running this on my large project. I am investigating. It looks like the correct cache contexts are not returned if the user is a "normal" member, only when they have a dedicated role such as administrator.

pfrenssen:
All tests are green now, here and in my project :)

idimopoulos (changes requested):
I have checked both this and the https://github.com/Gizra/og/pull/555 and they both work fine. However, as I noted in the other PR, I do not see the `::getRolesIds` being tested anywhere.
In this test everything is prophesied (I am not very familiar with UnitTests but I see that you are programming what will be returned) and the only other case where `::getRolesIds` is called is a test where it can fail if wrong information are returned. Can we have a small addition in the `OgMembershipTest` Kernel test (or even a UnitTest if you are more comfortable with) that actually test the return of the method?
Or at least point me to where it is if I did not find it.

pfrenssen:
Yes I agree that a KernelTest for the SQL version would be better. There was a bug in the query and I adapted the test, but it is not very reliable in this state. The test only really assures that the query is doing certain conditions but does not really guarantee that it won't regress if for example the conditions change or a new condition is added.

src/Cache/Context/OgRoleCacheContext.php
```diff
@@ -67,16 +87,22 @@ public static function getLabel() {
...
+   * @param \Drupal\og\MembershipManagerInterface $membershipManager
    *   The membership manager service.
-   * @param \Drupal\Core\PrivateKey $private_key
+   * @param \Drupal\Core\Database\Connection $database
+   *   The active database connection.
+   * @param \Drupal\Core\PrivateKey $privateKey
    *   The private key service.
    */
-  public function __construct(AccountInterface $user, MembershipManagerInterface $membership_manager, PrivateKey $private_key) {
+  public function __construct(AccountInterface $user, EntityTypeManagerInterface $entityTypeManager, MembershipManagerInterface $membershipManager, Connection $database, PrivateKey $privateKey) {
```
MPParsley, line 99: I'm OK with using camel case vs snake case here but I think it would be the first time we use camel case for the formal parameters in a constructor.

MPParsley:
@pfrenssen, should this still be marked as postponed?

pfrenssen:
Yes this is still postponed on #555 which should go in first. But #555 is now near completion, the latest remarks were addressed. There were some failures but they seem to be infrastructure related.

For this PR I would like to convert the unit test for the query into a kernel test so that this is more reliably tested. I can work on this today, this is back in sprint for our project.

src/Cache/Context/OgRoleCacheContext.php
pfrenssen: Oh I was not even aware any more of this change. Both are allowed according to the coding standards, and there is no preferred variant. I don't think it is very important for OG to be consistent in this.
I have a personal preference for the camelcased variant since I think it is more modern and in line with other PHP frameworks, but this is just personal and I'm willing to use snake case if this is considered the standard for OG.

MPParsley: > I don't think it is very important for OG to be consistent in this.
Except when there are other PR's that also add/inject new items, you'll get awkward merge issues.

> I have a personal preference for the camelcased variant since I think it is more modern and in line with other PHP frameworks.
I agree with this.

Since there's probably no other PR changing this class I'm fine with this.

pfrenssen:
#555 is in so this is unblocked!

tests/src/Kernel/Cache/Context/OgRoleCacheContextTest.php
```diff
@@ -0,0 +1,248 @@
...
+    // test as a custom role in addition to the default roles 'member',
+    // 'administrator', etc.
+    foreach (['entity_test', 'node'] as $entity_type_id) {
+      /** @var \Drupal\og\OgRoleInterface $role */
+      $role = OgRole::create();
+      $role
+        ->setGroupType($entity_type_id)
+        ->setGroupBundle('group')
+        ->setName('moderator');
+      $role->save();
```
MPParsley: I think we can chain this:
```suggestion
        ->save();
```

tests/src/Kernel/Cache/Context/OgRoleCacheContextTest.php
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The group type manager.
```
MPParsley:
```suggestion
   * The OG group type manager service.
```
Just like "The OG membership manager service"

MPParsley (approved):

tests/src/Kernel/Cache/Context/OgRoleCacheContextTest.php
```diff
@@ -0,0 +1,248 @@
...
+    // Create the 'moderator' role for both group types. This is used in the
+    // test as a custom role in addition to the default roles 'member',
+    // 'administrator', etc.
+    foreach (['entity_test', 'node'] as $entity_type_id) {
+      /** @var \Drupal\og\OgRoleInterface $role */
+      $role = OgRole::create();
+      $role
+        ->setGroupType($entity_type_id)
+        ->setGroupBundle('group')
+        ->setName('moderator');
```
pfrenssen:
```suggestion
        ->setName('moderator')
```

pfrenssen:
Awesome, this concludes the work on #487. With this in the lookup of the cache context is around 14x faster. For more details on the speed improvements see this comment: https://github.com/Gizra/og/issues/487#issuecomment-515551309

PR #560, opened 2019-08-01
Create a test for the ::getRolesIds().

idimopoulos:
Sorry, I do not have access to directly push to Gizra but I discussed with @pfrenssen and he allowed me to create a test for `::getRolesIds`.

### DIFF
--- a/tests/src/Kernel/Entity/OgMembershipTest.php
+++ b/tests/src/Kernel/Entity/OgMembershipTest.php
@@ -778,6 +778,34 @@ class OgMembershipTest extends KernelTestBase {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


idimopoulos:
This PR is an addition to https://github.com/Gizra/og/pull/560

pfrenssen:
Thanks, this is a very useful addition. This method is very important so it is critical that we have test coverage for it!

However I would prefer this to be rerolled against the current develop branch instead of being tacked on to #560. #560 is not about retrieving role IDs so a test for this in not in scope of that PR.

pfrenssen:
I rebased this on 8.x-1.x and pushed it to a new PR: #561.

Let's continue this work there.

PR #562, opened 2019-08-08
Providing settings under 'handler_settings' is deprecated

pfrenssen:
This PR is intended to address one of the deprecations warnings we are receiving and are preventing us from moving to Drupal 8.8 in the future:

```
  15x: Providing settings under 'handler_settings' is deprecated and will be removed before 9.0.0. Move the settings in the root of the configuration array. See https://www.drupal.org/node/2870971.

    6x in SelectionHandlerTest::testSelectionHandlerResults from Drupal\Tests\og\Kernel\Entity

    4x in EntityCreateAccessTest::testViewPermissionDoesNotGrantCreateAccess from Drupal\Tests\og\Kernel\Entity

    3x in OgComplexWidgetTest::testFields from Drupal\Tests\og\Functional

    2x in SelectionHandlerTest::testSelectionHandler from Drupal\Tests\og\Kernel\Entity
```

The change record for this deprecation is [Entity reference selection handlers should extend a base class](https://www.drupal.org/node/2870971).

### DIFF
--- a/config/schema/og.schema.yml
+++ b/config/schema/og.schema.yml
@@ -13,9 +13,9 @@ field.field_settings.og_membership_reference:
     handler:
       type: string
       label: 'Reference method'
-    handler_settings:
-      type: entity_reference_selection.[%parent.handler]
-      label: 'Organic Groups reference selection plugin settings'
+    field_mode:
+      type: string
+      label: 'The field mode'
     access_override:
       type: boolean
       label: 'Access Override'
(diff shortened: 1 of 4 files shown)


MPParsley:
@pfrenssen, if we fix the 8.8 issue here we should also merge this in #181 as that one is marked as a beta blocker or how do you see this?

pfrenssen:
You're right. We should focus on #181 first since when that is in probably all the offending code is removed anyway and we don't need to refactor it any more.

I'll postpone this on #181.

claudiu-cristea:

PR #567, opened 2019-09-06
Remove memberships when a group gets deleted

matthijs-va:
When a group gets deleted, its memberships should be removed as well.

### DIFF
--- a/og.module
+++ b/og.module
@@ -88,7 +88,6 @@ function og_entity_predelete(EntityInterface $entity) {
     // @todo Delete user roles.
     // @see https://github.com/amitaibu/og/issues/175
     // og_delete_user_roles_by_group($entity_type, $entity);
-
     // Delete the group memberships.
     foreach ($membership_manager->getMembershipsByGroup($entity, []) as $membership) {
       $membership->delete();
--- a/og.module
+++ b/og.module
@@ -71,9 +71,11 @@ function og_entity_update(EntityInterface $entity) {
  * Implements hook_entity_predelete().
  */
 function og_entity_predelete(EntityInterface $entity) {
+  /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+  $membership_manager = \Drupal::service('og.membership_manager');
+
   if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle())) {
-    // Register orphaned group content and user memberships for deletion, if
-    // this option has been enabled.
+    // Register orphaned group content for deletion, if this option has been enabled.
     $config = \Drupal::config('og.settings');
     if ($config->get('delete_orphans')) {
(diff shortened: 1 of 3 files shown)


og.module
```diff
@@ -71,9 +71,11 @@ function og_entity_update(EntityInterface $entity) {
  * Implements hook_entity_predelete().
  */
 function og_entity_predelete(EntityInterface $entity) {
+  /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
+  $membership_manager = \Drupal::service('og.membership_manager');
+
   if (Og::isGroup($entity->getEntityTypeId(), $entity->bundle())) {
-    // Register orphaned group content and user memberships for deletion, if
-    // this option has been enabled.
+    // Register orphaned group content for deletion, if this option has been enabled.
```
```suggestion
    // Register orphaned group content for deletion, if this option has been
    // enabled.
```

og.module
```diff
@@ -71,9 +71,11 @@ function og_entity_update(EntityInterface $entity) {
  * Implements hook_entity_predelete().
  */
 function og_entity_predelete(EntityInterface $entity) {
+  /** @var \Drupal\og\MembershipManagerInterface $membership_manager */
```
MPParsley: I'd put this later on, just before the foreach.

matthijs-va: But then I would have to get it twice.

MPParsley (approved):

og.module
```diff
@@ -85,11 +88,15 @@ function og_entity_predelete(EntityInterface $entity) {
     // @todo Delete user roles.
     // @see https://github.com/amitaibu/og/issues/175
     // og_delete_user_roles_by_group($entity_type, $entity);
+
```
```suggestion
```
----------------------------------
 90 | WARNING | [x] There must be no blank line following an inline comment
 90 | WARNING | [ ] There must be no blank line following an inline comment

MPParsley: @matthijs-va can you remove this line? Codesniffer fails on it.
```suggestion
```

pfrenssen:
This is already implemented, you can see this being tested in `OgDeleteOrphansTest::testDeleteOrphans()` - at the start of the test there is 1 user membership, then the group is deleted, then the cleaning up of orphans is executed, and at the end of the test it is validated that the user membership is no longer there.

We can't clean up user memberships on `hook_entity_predelete()` like is proposed in this PR because it would cause timeouts when deleting groups with thousands of members.

By default user memberships are cleaned up on cron, but we also offer an option to delete them instantly, or to let an administrator clean them up manually through the UI by launching a batch operation. We offer a plugin type for developers who have custom requirements for removing the memberships after a group is deleted (ref `OgDeleteOrphansInterface`).

If you are seeing in your website that the user memberships are not being cleaned up, then make sure your cron jobs are correctly configured, or switch to the "Simple" plugin in the OG admin UI.

PR #573, opened 2019-12-04
MembershipManager should support UpdateBackend as cache backend during updates

### LINKED ISSUE #572: Error while running Drupal 8.8.x DB updates
```
>  [notice] Update started: system_update_8803
>  [error]  The cache backend must support cache tag invalidation.
>  [error]  Update failed: system_update_8803
...
 [error]  Update aborted by: system_update_8803
 [error]  Finished performing updates.
```

This happens because in `MembershipManager` this assertion is triggered:

```php
public function __construct(EntityTypeManagerInterface $entity_type_manager, OgGroupAudienceHelperInterface $group_audience_helper, CacheBackendInterface $cache) {
  assert($cache instanceof CacheTagsInvalidatorInterface, 'The cache backend must support cache tag invalidation.');
  ...
}
```

But during updates the cache backend, injected as `$cache` in the constructor, is an instance of the class `\Drupal\Core\Update\UpdateBackend` and that doesn't implement `CacheTagsInvalidatorInterface`.

claudiu-cristea:
Fixes #572

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -11,6 +11,7 @@ use Drupal\Core\Cache\CacheTagsInvalidatorInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Update\UpdateBackend;
 use Drupal\field\Entity\FieldStorageConfig;
 use Drupal\field\FieldStorageConfigInterface;
 use Drupal\og\Entity\OgMembership;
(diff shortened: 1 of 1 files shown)


pfrenssen (approved):

pfrenssen:
This is OK to me, the `UpdateBackend` extends `NullBackend` and doesn't cache anything at all, so we don't need to invalidate anything from it.

It would be even better if we could check for `NullBackend`?

pfrenssen (approved):

pfrenssen:
The assert should indeed be removed, as per the discussion in #572 
Thanks!

PR #577, opened 2019-12-07
The 'handler_settings' plugin config is deprecated

### LINKED ISSUE #519: Providing settings under 'handler_settings' is deprecated
When running tests on Drupal 8.8 we are getting the following deprecation notice:

```
  15x: Providing settings under 'handler_settings' is deprecated and will be removed before 9.0.0. Move the settings in the root of the configuration array. See https://www.drupal.org/node/2870971.
    6x in SelectionHandlerTest::testSelectionHandlerResults from Drupal\Tests\og\Kernel\Entity
    4x in EntityCreateAccessTest::testViewPermissionDoesNotGrantCreateAccess from Drupal\Tests\og\Kernel\Entity
    3x in OgComplexWidgetTest::testFields from Drupal\Tests\og\Functional
    2x in SelectionHandlerTest::testSelectionHandler from Drupal\Tests\og\Kernel\Entity
```

Change record: [Entity reference selection handlers should extend a base class](https://www.drupal.org/node/2870971)

claudiu-cristea:
Fixes #519

### DIFF
--- a/src/Og.php
+++ b/src/Og.php
@@ -391,16 +391,14 @@ class Og {
       throw new \Exception("The field $field_name is not an audience field.");
     }
 
-    $options = NestedArray::mergeDeep([
+    $default_options = [
       'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
       'handler' => $field_definition->getSetting('handler'),
-      'handler_settings' => [
-        'field_mode' => 'default',
-      ],
-    ], $options);
+      'field_mode' => 'default',
+    ] + $field_definition->getSetting('handler_settings');
 
-    // Deep merge the handler settings.
-    $options['handler_settings'] = NestedArray::mergeDeep($field_definition->getSetting('handler_settings'), $options['handler_settings']);
+    // Override with passed $options.
+    $options = NestedArray::mergeDeep($default_options, $options);
 
     return \Drupal::service('plugin.manager.entity_reference_selection')->createInstance('og:default', $options);
   }
--- a/src/Plugin/EntityReferenceSelection/OgSelection.php
+++ b/src/Plugin/EntityReferenceSelection/OgSelection.php
@@ -19,7 +19,7 @@ use Drupal\og\Og;
  *   id = "og:default",
  *   label = @Translation("OG selection"),
  *   group = "og",
- *   weight = 1
+ *   weight = 1,
  * )
  */
 class OgSelection extends DefaultSelection {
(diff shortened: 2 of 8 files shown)


tests/src/Kernel/SelectionHandlerSettingsSchemaTest.php
```diff
@@ -39,6 +39,8 @@ public function testSelectionHandlerSettingsSchema() {
     $node_type->og_target_type = 'node';
     $node_type->og_target_bundles = [$bundle];
     $node_type->save();
+    // Provide at least one assertion.
+    $this->assertTrue(TRUE);
```
pfrenssen: Let's explain here why this is done, people might not be familiar with the way PHPUnit marks tests without asserts as being risky.

src/Plugin/EntityReferenceSelection/OgSelection.php
```diff
@@ -31,13 +31,16 @@ class OgSelection extends DefaultSelection {
...
-    ];
+    $options = $this->getConfiguration();
+    // The 'handler' key intentionally absent as we want the selection manager
+    // to choose the best option.
+    // @see \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager::getInstance()
+    unset($options['handler']);
+    // Remove also de backwards compatibility layer because that will be passed
+    // to the chosen selection handler setter and, as an effect, will trigger a
+    // deprecation notice.
+    // @see \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginBase::resolveBackwardCompatibilityConfiguration()
```
pfrenssen: Typo: "de" -> "the"

src/Og.php
```diff
@@ -391,16 +391,14 @@ public static function getSelectionHandler(FieldDefinitionInterface $field_defin
...
-    $options = NestedArray::mergeDeep([
+    $default_options = [
       'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
       'handler' => $field_definition->getSetting('handler'),
-      'handler_settings' => [
-        'field_mode' => 'default',
-      ],
-    ], $options);
+      'field_mode' => 'default',
+    ] + $field_definition->getSetting('handler_settings');
```
pfrenssen, line 398: This retrieval of the `handler_settings` is still here to account for the fact that field definitions can be passed in that still use the deprecated handler settings. It's OK, the default implementation in core also still has this.

On a side note, it looks to me that this function is probably not really needed any more. The only thing that seems to make it different now from the standard implementation in `SelectionPluginManager::getSelectionHandler()` is that is allows to override the handler options, but it is only called once (in `og_entity_create_access()` and here the `$options` are not even passed. I will make a followup issue to deprecate and remove this.

Edit: created followup #580

pfrenssen (changes requested):
Very nice work, just a small nitpick.

pfrenssen (approved):
Fixed the nitpicks, this is good to go! Thanks a lot!!

PR #583, opened 2019-12-08
Use currently supported versions of PHP and Drupal to test against

pfrenssen:
In the meantime PHP 7.4 and Drupal 8.8 have been released, and PHP 7.1 has been deprecated. There is also a new Drupal 8.9 development branch that is interesting for forwards compatibility.

This PR updates the test matrix to the currently supported version ranges.

Drupal 8.7 is also deprecated but we should keep it for a little while longer until OG is stable against Drupal 8.8. It can be removed in a future PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ sudo: true
 dist: xenial
 
 php:
-  - 7.1
   - 7.2
   - 7.3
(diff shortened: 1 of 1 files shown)


claudiu-cristea:
Fixes #585

PR #593, opened 2019-12-08
Require tests to pass on Drupal 8.8.x

### LINKED ISSUE #585: Require tests on Drupal 8.8.x to pass in the test matrix
Before we make the new release we should ensure that all our tests will pass on Drupal 8.8: we should remove the `allow_failures` setting from the Travis CI config from the test matrix.

pfrenssen:
Fixes #585.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ matrix:
     - php: 7.4
       env: TEST_SUITE=PHP_CodeSniffer
   allow_failures:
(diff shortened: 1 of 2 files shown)


claudiu-cristea:
@pfrenssen let's require also for Drupal 8.9.x as now they are green so it would easier now to catch and fix a failure on a per-PR basis rather than when the branches will diverge too much.

claudiu-cristea (approved):

PR #570, opened 2019-09-10
160 remove complex widget

MPParsley:
Move amitaibu:160-remove-complex-widget & damienmckenna:160-remove-complex-widget to Gizra:og

See #181 & #558.

Fixes #160

PR #596, opened 2019-12-10
Removed redundant getGroupBundlesByEntityType() method.

damienmckenna:
This is for #570.

### DIFF
--- a/src/GroupTypeManager.php
+++ b/src/GroupTypeManager.php
@@ -191,7 +191,7 @@ class GroupTypeManager implements GroupTypeManagerInterface {
    * {@inheritdoc}
    */
   public function getAllGroupBundles($entity_type = NULL) {
-    // @todo Should be remove since this method don't do any thing.
+    // @todo Should be removed since this method don't do any thing.
     return $this->getGroupMap();
   }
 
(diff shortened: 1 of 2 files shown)


src/GroupTypeManager.php
```diff
@@ -191,24 +191,10 @@ public function getGroupBundleIdsByEntityType($entity_type_id) {
    * {@inheritdoc}
    */
   public function getAllGroupBundles($entity_type = NULL) {
-    // Todo - should be remove since this method don't do any thing.
+    // @todo Should be remove since this method don't do any thing.
```
```suggestion
    // @todo Should be removed since this method don't do any thing.
```

MPParsley (approved):

src/GroupTypeManager.php
damienmckenna: Doh.

MPParsley:
@damienmckenna, I reverted this as it seems this introduced several failures in #570.

Have a look at this remark: https://github.com/Gizra/og/pull/181/files#r199885163

> Let's undo this change, it's out of scope for this issue and it causes a fatal error because now the signature is different from the one defined in GroupTypeManagerInterface.
> 
> There is already an issue to remove this method: #372, let's do it in there.

Issue #546, opened 2019-07-23
`User::load()` calls should be avoided in classes


pfrenssen:
- [x] Remove the following line from `phpcs-ruleset.xml.dist`:
`<exclude name="DrupalPractice.Objects.GlobalClass.GlobalClass" />`
- [x] Run `./vendor/bin/phpcs`
- [x] Fix all reported warnings.

MPParsley:
Solved in #601

PR #601, opened 2019-12-23
`User::load()` calls should be avoided in classes #546 fix

ghost:
Updated CS sniffer ruleset, run it, replaced User::load with dependency injection. #546

### DIFF
--- a/phpcs-ruleset.xml.dist
+++ b/phpcs-ruleset.xml.dist
@@ -22,10 +22,6 @@
     <rule ref="./vendor/drupal/coder/coder_sniffer/DrupalPractice">
         <!-- https://github.com/Gizra/og/issues/549 -->
         <exclude name="DrupalPractice.CodeAnalysis.VariableAnalysis.UndefinedVariable" />
-        <!-- https://github.com/Gizra/og/issues/545 -->
-        <exclude name="DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable" />
-        <!-- https://github.com/Gizra/og/issues/548 -->
-        <exclude name="DrupalPractice.Commenting.ExpectedException.TagFound" />
         <!-- https://github.com/Gizra/og/issues/550 -->
         <exclude name="DrupalPractice.Constants.GlobalDefine.GlobalConstant" />
         <!-- https://github.com/Gizra/og/issues/544 -->
--- a/phpcs-ruleset.xml.dist
+++ b/phpcs-ruleset.xml.dist
@@ -26,8 +26,6 @@
         <exclude name="DrupalPractice.Constants.GlobalDefine.GlobalConstant" />
         <!-- https://github.com/Gizra/og/issues/544 -->
         <exclude name="DrupalPractice.InfoFiles.NamespacedDependency.NonNamespaced" />
-        <!-- https://github.com/Gizra/og/issues/546 -->
-        <exclude name="DrupalPractice.Objects.GlobalClass.GlobalClass" />
         <!-- https://github.com/Gizra/og/issues/543 -->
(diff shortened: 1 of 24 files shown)


src/Controller/SubscriptionController.php
```diff
@@ -36,17 +36,31 @@ class SubscriptionController extends ControllerBase {
...
    * @param \Drupal\Core\Messenger\MessengerInterface $messenger
    *   The messenger service.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager service.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  public function __construct(OgAccessInterface $og_access, MessengerInterface $messenger) {
+  public function __construct(OgAccessInterface $og_access, MessengerInterface $messenger, EntityTypeManagerInterface $entity_type_manager) {
```
pfrenssen: In the `SubscriptionController` we don't need to inject the `EntityTypeManager` since we already inherit the `$this->getEntityTypeManager()` method from the base class (`ControllerBase`).

src/Controller/SubscriptionController.php
```diff
@@ -84,7 +99,7 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       throw new AccessDeniedHttpException();
     }
 
-    $user = User::load($this->currentUser()->id());
+    $user = $this->entityTypeManager->load($this->currentUser()->id());
```
pfrenssen: Here instead we can call `$this->getEntityTypeManager()->getStorage('user')->load()`.

pfrenssen (changes requested):
Looking good, but in one case it is not necessary to inject the entity type manager ourselves, since the base class already has a convenient method that allows us to access it.

MPParsley (approved):
Looks good

MPParsley (approved):

MPParsley (approved):

phpcs-ruleset.xml.dist
```diff
@@ -28,6 +28,8 @@
         <exclude name="DrupalPractice.InfoFiles.NamespacedDependency.NonNamespaced" />
         <!-- https://github.com/Gizra/og/issues/546 -->
         <exclude name="DrupalPractice.Objects.GlobalClass.GlobalClass" />
+        <!-- https://github.com/Gizra/og/issues/547 -->
+        <exclude name="DrupalPractice.Objects.GlobalDrupal.GlobalDrupal" />
```
pfrenssen: These two lines are out of scope for this PR. I think they were probably introduced by accident through a merge conflict. Let's remove them.

pfrenssen (dismissed):

phpcs-ruleset.xml.dist
```diff
@@ -28,6 +28,8 @@
         <exclude name="DrupalPractice.InfoFiles.NamespacedDependency.NonNamespaced" />
         <!-- https://github.com/Gizra/og/issues/546 -->
         <exclude name="DrupalPractice.Objects.GlobalClass.GlobalClass" />
```
MPParsley: We should delete these though.

phpcs-ruleset.xml.dist
MPParsley: We should delete these lines (they've probably been reintroduced in the merge conflict):
```
        <!-- https://github.com/Gizra/og/issues/546 -->
        <exclude name="DrupalPractice.Objects.GlobalClass.GlobalClass" />

```

MPParsley (approved):

pfrenssen (approved):

Issue #547, opened 2019-07-23
`\Drupal` calls should be avoided in classes


pfrenssen:
- [ ] Remove the following line from `phpcs-ruleset.xml.dist`:
`<exclude name="DrupalPractice.Objects.GlobalDrupal.GlobalDrupal" />`
- [ ] Run `./vendor/bin/phpcs`
- [ ] Fix all reported warnings.

PR #602, opened 2019-12-24
Added dependency injection for EntityTypeManager instead \Drupal calls

kbrodej:
Fix for #547 $entity_type_manager was already in construct but never used.


pfrenssen:
Hah this is a very good find, we were already injecting it but we forgot to store the dependency locally and were accessing it through the legacy static `\Drupal::entityTypeManager()`!

Great job, thanks @kbrodej !

pfrenssen (approved):

Issue #545, opened 2019-07-23
Remove unused variables


pfrenssen:
- [x] Remove the following line from `phpcs-ruleset.xml.dist`:
`<exclude name="DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable" />`
- [x] Run `./vendor/bin/phpcs`
- [x] Fix all reported warnings.

pfrenssen:
Fixed in #604

PR #604, opened 2019-12-24
Removed unused variables

kbrodej:
Fix for #545

### DIFF
--- a/src/ContextProvider/OgContext.php
+++ b/src/ContextProvider/OgContext.php
@@ -126,7 +126,6 @@ class OgContext implements OgContextInterface, ContextProviderInterface {
    */
   protected function getBestCandidate() {
     $collection = new OgResolvedGroupCollection();
-    $plugins = [];
 
     // Retrieve the list of group resolvers. These are stored in config, and are
     // ordered by priority.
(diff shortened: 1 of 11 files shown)


src/ContextProvider/OgContext.php
```diff
@@ -135,8 +134,7 @@ protected function getBestCandidate() {
     foreach ($group_resolvers as $plugin_id) {
       /** @var \Drupal\og\OgGroupResolverInterface $plugin */
       if ($plugin = $this->pluginManager->createInstance($plugin_id)) {
-        $plugins[$plugin_id] = $plugin;
-
+        // $plugins[$plugin_id] = $plugin;
```
MPParsley: this doesn't seem right?

kbrodej: Fixed in the next commit.

tests/src/Unit/PermissionEventTest.php
```diff
@@ -445,6 +445,8 @@ public function testOffsetSet(array $permissions, $entity_type_id, $bundle_id, a
   public function testOffsetSetInvalidPermission($key, $permission) {
     $event = new PermissionEvent($this->randomMachineName(), $this->randomMachineName(), []);
     $event[$key] = $permission;
+
+    return $event;
```
pfrenssen: This should not be here, this test should not return any values. PHP_CodeSniffer cannot understand here that the variable is actually used in that it triggers an exception.

Let's add the following comments around it to disable this check:

```
    // Avoid false positive in coding standards checking. The variable is actually used for the test.
    // phpcs:disable DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable
    $event = new PermissionEvent($this->randomMachineName(), $this->randomMachineName(), []);
    $event[$key] = $permission;
    // phpcs:enable DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable
```

pfrenssen (changes requested):
Thanks for working on this! This is looking good, except for one case: PHP_CodeSniffer cannot understand the way that our test is using a variable assignment to trigger an exception. Instead of adding a fake line that returns the value, we should just disable this coding standards check for this case.

pfrenssen (approved):

Issue #548, opened 2019-07-23
`@expectedException` tags should not be used


pfrenssen:
- [x] Remove the following line from `phpcs-ruleset.xml.dist`:
`<exclude name="DrupalPractice.Commenting.ExpectedException.TagFound" />`
- [x] Run `./vendor/bin/phpcs`
- [x] Fix all reported warnings.

pfrenssen:
Fixed in #605

PR #605, opened 2020-01-07
Removed @expectedException tags

kbrodej:
Fix for #548


phpcs-ruleset.xml.dist
```diff
@@ -25,7 +25,6 @@
         <!-- https://github.com/Gizra/og/issues/545 -->
         <exclude name="DrupalPractice.CodeAnalysis.VariableAnalysis.UnusedVariable" />
         <!-- https://github.com/Gizra/og/issues/548 -->
-        <exclude name="DrupalPractice.Commenting.ExpectedException.TagFound" />
```
pfrenssen: Let's also remove the comment line for this, it is no longer needed after this is accepted.

pfrenssen (approved):
Looking great, thanks so much for this PR!

I like how the `$this->expectException()` calls are always placed right before the line that triggers the exception, it makes the code "exceptionally" easy to follow :grin: :+1:

PR #610, opened 2020-02-06
580 Deprecate `Og::getSelectionHandler()`

### LINKED ISSUE #580: Deprecate `Og::getSelectionHandler()`
The static function `Og::getSelectionHandler()` seems to be no longer needed and can be removed. The only thing that seems to make it different now from the standard implementation in `SelectionPluginManager::getSelectionHandler()` is that is allows to override the handler options, but it is only called once (in `og_entity_create_access()`) and here the `$options` are not even passed.

So we can remove this and call into the `SelectionPluginManager` instead.

It is possible that some of our users are still calling this in their custom code so we should handle this in two PRs: first deprecate it, and then in a later release it can be removed completely.

kbrodej:
Fix for #580. I hope i got the versions correct :)

### DIFF
--- a/og.module
+++ b/og.module
@@ -218,7 +218,7 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
     /** @var \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager $handler */
     $handler = \Drupal::service('plugin.manager.entity_reference_selection');
 
-    if ($handler->createInstance('og:default',$options)) {
+    if ($handler->createInstance('og:default', $options)) {
       return AccessResult::neutral();
     }
     // Allow users to create content outside of groups, if none of the
--- a/og.module
+++ b/og.module
@@ -210,12 +210,17 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
       continue;
     }
 
-    $handler = Og::getSelectionHandler($field_definition);
+    $options = [
+      'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
+      'handler' => $field_definition->getSetting('handler'),
+      'field_mode' => 'admin',
+    ];
+    /** @var \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager $handler */
+    $handler = \Drupal::service('plugin.manager.entity_reference_selection');
 
-    if ($handler->getReferenceableEntities()) {
+    if ($handler->createInstance('og:default',$options)) {
       return AccessResult::neutral();
     }
-
     // Allow users to create content outside of groups, if none of the
     // audience fields is required.
     $required = $field_definition->isRequired();
--- a/src/Og.php
+++ b/src/Og.php
@@ -389,7 +389,7 @@ class Og {
    *
    * @deprecated in og:8.x-1.0-alpha4 and is removed from og:8.x-1.0-alpha5.
    *   Use
-   *   \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager::getSelectionHandler()
+   *   \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager::getInstance()
    *   Instead.
    * @codingStandardsIgnoreStart
    * @see https://github.com/Gizra/og/issues/580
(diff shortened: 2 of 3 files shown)


MPParsley:
Thanks @kbrodej. Seems like the deprecated method is still being used:
```
Remaining deprecation notices (5)
  5x: Og:getSelectionHandler() is deprecated in og 8.x-1.0-alpha4
      and is removed in og 8.x-1.0-alpha5.
      Use \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager::getInstance()
      instead. See https://github.com/Gizra/og/issues/580
    2x in EntityCreateAccessTest::testViewPermissionDoesNotGrantCreateAccess from Drupal\Tests\og\Kernel\Entity
    2x in SelectionHandlerTest::testSelectionHandlerResults from Drupal\Tests\og\Kernel\Entity
    1x in SelectionHandlerTest::testSelectionHandler from Drupal\Tests\og\Kernel\Entity
```

[Drupal core expects a unit test proving the deprecation notice will be triggered](https://www.drupal.org/core/deprecation) when the deprecated code is called. @pfrenssen, what's our policy here?

kbrodej:
Hi. Yes.  I pulled the push trigger too soon. totally forgot about the tests. I am working on it atm :)

kbrodej:
The tests are fixed now, no more deprecation messages but why that build failed i don't know. Passed locally with flying colours :). Maybe re-trigger it @MPParsley ?

MPParsley:
> The tests are fixed now, no more deprecation messages but why that build failed i don't know. Passed locally with flying colours :). Maybe re-trigger it @MPParsley ?

curl: (7) Failed to connect to storage.googleapis.com port 443: Connection timed out

Triggered new build

pfrenssen:
Thanks a lot for picking this up!

This looks good to me but indeed one job failed due to a connection issue. When we can get a green then this is ready to merge.

pfrenssen:
Retest is green! Merging. Thanks a lot for the contribution!

og.module
```diff
@@ -210,12 +210,17 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
...
-    if ($handler->getReferenceableEntities()) {
+    $options = [
+      'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
+      'handler' => $field_definition->getSetting('handler'),
+      'field_mode' => 'admin',
+    ];
+    /** @var \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager $handler */
+    $handler = \Drupal::service('plugin.manager.entity_reference_selection');
+
+    if ($handler->createInstance('og:default', $options)) {
```
MPParsley, line 221: @kbrodej, the deprecation says we should use SelectionPluginManager::getInstance. Why are we using createInstance() here, can't we use $handler->getInstance($options)? See #622.

Also, why are these two considered equivalent?
    if ($handler->createInstance('og:default', $options)) {
    if ($handler->getReferenceableEntities()) {

@pfrenssen, this may require a comment. This is causing a merge conflict in #570 (it's using $handler->**countReferenceableEntities()**), I'd like to know why we tested for **getReferenceableEntities()** and now on **createinstance()** so I can resolve the conflict.

og.module
```diff
@@ -210,12 +210,17 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
       continue;
     }
 
-    $handler = Og::getSelectionHandler($field_definition);
-
-    if ($handler->getReferenceableEntities()) {
+    $options = [
+      'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
+      'handler' => $field_definition->getSetting('handler'),
+      'field_mode' => 'admin',
```
MPParsley, line 216: Did we introduce a bug here, before this PR there wasn't a 'field_mode' => 'admin' in the config?

cc @pfrenssen, @kbrodej

og.module
kbrodej: Oversight on my side. We should use SelectionPluginManager::getInstance() in all cases and we shouldn't consider createInstance() and getReferenceableEntities() as equal.

Issue #642, opened 2020-04-30
Other groups field ignoring bundles settings


alfaguru:
To reproduce this error, configure more than one content type (CT1, CT2) to be a group. Add an audience field to another content type( CT3), and configure it so only CT1 is valid.

**Problem 1**. On editing a CT3 the "Other groups" field will return nodes of type CT2 as well as CT1. 

**Problem 2**. Attempting to submit with a CT2 node selected will cause an exception to be thrown as the validation of the "Other groups" field is incorrect in some way not yet identified.

I have a fix for problem 1 and will add a PR for it shortly.

pfrenssen:
Fixed in #643

PR #643, opened 2020-04-30
Other Groups field should respect bundles config

alfaguru:
See issue #642


pfrenssen:
Thanks for this. I have not tested it but it seems to make sense to me.

We are in the process though of removing the `OgComplex` field (ref #570) so this is fixing code that is going to be removed. However since the removal of `OgComplex` seems to have stalled because it needs to be redone from scratch, maybe it is a good idea to get this in to help people in the short term.

@MPParsley what is your opinion?

MPParsley:
Indeed, this issue should be fixed in the meanwhile.

OgComplex.php is removed in https://github.com/Gizra/og/pull/570/files#diff-9704db18679e46e953b1ea9a98c59078

I haven't tested it yet but I believe that removing the 'complex' field and using a 'simple' entity reference should prevent this issue from occurring.

pfrenssen:
Alright let's get this in.

Thanks @alfaguru for the contribution!

PR #647, opened 2020-07-02
Drupal core permissions are leaking into OG group content permissions

### LINKED ISSUE #646: Core permissions on group entities are leaking to group content
Use case:

* I have a "Community" group type which is intended to host posts from the group members, similar to a Facebook group.
* I have two group content types:
  1. "Post" nodes which are posted by the group members.
  2. "Page" nodes which contain basic information about the community.
* I have an "Editor" role which is allowed to edit and delete the "Page" nodes (which is an OG permission), and also the "Community" node itself (which is a core permission), but is not allowed to edit and delete the "Post" nodes.

Expected behavior:

* When I am logged in as an Editor and visit a "Post" node, then I should not have access to edit or delete the node.

Actual behavior:

* I have access to delete the node.

This is caused in this section of `OgAccess::userAccessEntity()`:

```
// Check if the operation matches a group level operation such as
// 'subscribe without approval'.
$user_access = $this->userAccess($group, $operation, $user);
if ($user_access->isAllowed()) {
  return $user_access->addCacheTags($cache_tags);
}

$forbidden->inheritCacheability($user_access);
```

This doesn't check that the operation passed on to `$this->userAccess()` is actually a group level operation (such as `subscribe without approval`) and not a core entity operation (such as 'view', 'update', 'delete'). At the moment all permissions that a user has on the group will fall through to group content.

Also it seems unnecessary to call `$this->userAccess()` to check for group level operations on group content. These permissions are typically things like `manage members`, `subscribe` etc which don't apply to group content. The call to `$this->userAccess()` earlier in the method should be sufficient.

This is probably a security issue since it gives unintended access to do basic entity operations on all group content entities if the user has this permission on the group entity. This will also break existing code which is relying on this behavior but it is clearly incorrect. If a role should have a certain access on group content they should be assigned the appropriate `GroupContentOperationPermission`.

pfrenssen:
Fixes #646 

I am starting with a test case. This doesn't yet exhibit the bug I am seeing. I will investigate further how this bug can be replicated.

### DIFF
--- a/src/OgAccess.php
+++ b/src/OgAccess.php
@@ -151,6 +151,8 @@ class OgAccess implements OgAccessInterface {
     }
 
     // User ID 1 has all privileges.
+    // @todo If we drop the 'user' cache context then at this point we should
+    //   add the "user.is_super_user" cache context.
     if ($user->id() == 1) {
       return AccessResult::allowed()->addCacheableDependency($cacheable_metadata);
     }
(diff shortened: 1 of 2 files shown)


pfrenssen:
Postponing this on the planned fixes of access / permission handling: #657 #658 #659

pfrenssen:
This is no longer relevant thanks to the recent series of PRs to fix and refactor `OgAccess`.

PR #648, opened 2020-07-02
Remove dead code in OgAccessHookTest

pfrenssen:
There is some dead code in OgAccessHookTest which has been identified before and has been marked with a `@todo`. It appears that this is just some relic from an earlier test case that no longer exists, and can simply be removed.

### DIFF
--- a/tests/src/Unit/OgAccessHookTest.php
+++ b/tests/src/Unit/OgAccessHookTest.php
@@ -44,17 +44,7 @@ class OgAccessHookTest extends OgAccessEntityTestBase {
   public function testGetEntityGroups($operation) {
     $this->user->hasPermission(OgAccess::ADMINISTER_GROUP_PERMISSION)->willReturn(TRUE);
     $user_entity_access = og_entity_access($this->groupContentEntity->reveal(), $operation, $this->user->reveal());
-
-    // @todo This is strange, 'view' is not part of the operations supplied by
-    //   ::permissionsProvider(). And why would a group administrator be allowed
-    //   access to all operations, except 'view'? Shouldn't this also return
-    //   'allowed'?
-    if ($operation == 'view') {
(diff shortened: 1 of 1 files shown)


pfrenssen:
Test has passed but is not reported back since we are still on the legacy webhooks integration which is deprecated.

Here is the build: https://travis-ci.org/github/Gizra/og/builds/704446848

MPParsley (approved):

pfrenssen:
Hmm I cannot merge this, github tells me:

```
remote: error: Required status check "continuous-integration/travis-ci" is expected.
```

I will try to trigger a rebuild.

pfrenssen:
Changed commit message and force pushed, let's see if this does the trick.

MPParsley:
Seems to have worked: https://travis-ci.org/github/Gizra/og/builds/704654475

pfrenssen:
Yes the builds are usually started without a problem, but often with the legacy webhooks integration the result of the build is not reported back. It's hit and miss, let's see if it works this time.

I will open an issue to switch to the newfangled Github Apps integration.

pfrenssen:
Same problem. I am restarting one of the two builds on Travis CI.

pfrenssen:
Progress! There is one passed build and one still pending, let's try once more.

MPParsley:
"Waiting to be queued" after 4hrs...

Perhaps it's easier to start a new PR?

pfrenssen:
I have done some reading on the Travis CI forum and this seems to be caused by Github enforcing rate limiting on the requests done by Travis CI. A new PR would not work for this, it seems that we just need to be lucky that our request is happening on a moment when the rate limit has not been reached. No idea how this works, if it is a rate limit per second, per hour, per day...?

In the meantime I have restarted one of the builds, and now we have green on all fronts! Let's merge quickly :D

PR #655, opened 2020-07-07
Remove the static cache in OgAccess::userAccess().

### LINKED ISSUE #654: Remove static cache in `OgAccess::userAccess()`
We currently have a static cache in `OgAccess::userAccess()` which adds some considerable complexity but does not seem to be needed. All of the external methods that we are calling that do database queries already have static caches. The only part that is not covered is the alter hook, but it is not our responsibility to provide a caching layer for expensive logic that happens inside an alter hook.

I did some digging and it turns out this was added back in 2009 (ref commit [9e8e1e5590](https://github.com/Gizra/og/commit/9e8e1e5590af129cff12d5e4ff99ef25b9153b23#diff-dbdf03a54bc4970e445b8de4fae2ae6aR1317)) to cache the results from `user_role_permissions()`. Even back then it was not needed, since `user_role_permissions()` has its own static cache, right from when it was initially added back in 2008 (ref [issue #73874](https://www.drupal.org/node/73874)).

So it seems this static cache has never served any useful purpose in the whole history of OG. Let's remove it.

pfrenssen:
Fixes #654

### DIFF
--- a/phpcs-ruleset.xml.dist
+++ b/phpcs-ruleset.xml.dist
@@ -16,6 +16,7 @@
     <rule ref="./vendor/drupal/coder/coder_sniffer/Drupal">
         <!-- URLs in deprecation messages point to Github rather than drupal.org. -->
         <exclude name="Drupal.Semantics.FunctionTriggerError.TriggerErrorSeeUrlFormat" />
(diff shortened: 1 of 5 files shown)


MPParsley (approved):

pfrenssen:
Thanks for reviewing! I am planning to clean up and improve `OgAccess` over the coming days. It needs some tender loving care :)

PR #661, opened 2020-07-13
Rename the global permission to administer all groups to 'administer organic groups'

### LINKED ISSUE #657: Rename 'administer group' permission to 'administer organic groups'
A user can be given a global permission to give them access to manage all groups in the website. This permission is called `administer group`, but there are a few problems with this permission:

1. It is not properly namespaced, and is conflicting with the [same permission from the Group module](https://git.drupalcode.org/project/group/-/blob/8.x-1.x/group.group.permissions.yml#L2).
2. It is confusing in the context of Organic Groups since it seems to imply the permission is only valid for a single group entity.

I had a look at why this is called as such and it is for historical reasons. OG originally had a bunch of permissions such as `administer group users`, `administer group permissions` and at some point a new global permission was added called `administer organic groups` (ref commit [44d6a32](https://github.com/Gizra/og/commit/44d6a325ec1a3734fa84363e57b323ff8d32bec9#diff-dbdf03a54bc4970e445b8de4fae2ae6aR1830)). Then it was renamed to `administer group` in a gigantic commit which had the intention back in 2010 to rename Organic Groups to Group (ref commit [ec296b4](https://github.com/Gizra/og/commit/ec296b4d42948e9a1810cda978307f18edfb0fa1)). This plan was abandoned later on and the name "Organic Groups" was kept, but the permission wasn't changed back.

Let's call it "administer organic groups" again so we don't cause any conflicts with the Group module, but keep a B/C layer in place for people upgrading their sites from D7.

pfrenssen:
Fixes #657

### DIFF
--- a/og.permissions.yml
+++ b/og.permissions.yml
@@ -1,4 +1,4 @@
 administer organic groups:
-  title: 'Administer Organic groups permissions'
+  title: 'Administer Organic groups'
   description: 'Administer all groups and permissions.'
   restrict access: true
--- a/og.permissions.yml
+++ b/og.permissions.yml
@@ -1,4 +1,4 @@
-administer group:
+administer organic groups:
   title: 'Administer Organic groups permissions'
   description: 'Administer all groups and permissions.'
   restrict access: true
--- a/og.module
+++ b/og.module
@@ -147,8 +147,9 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface 
     return AccessResult::neutral();
   }
 
-  // If the user has permission to administer all groups, allow access.
-  if ($account->hasPermission('administer group')) {
+  // If the user has the global permission to administer all groups, allow
+  // access.
+  if ($account->hasPermission('administer organic groups')) {
     return AccessResult::allowed();
   }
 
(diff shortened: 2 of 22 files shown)


pfrenssen:
Drupal 9.1.x is failing due to a new deprecation notice. This is fixed in #662 and merged back here.

pfrenssen:
A wild deprecation appears! https://www.drupal.org/node/3154407

pfrenssen:
This deprecation is not solvable in a way that works on both D8 and D9. I think we should probably disable deprecation checking in D9 builds until D8 goes EOL in November 2021. I opened #663 to handle this.

MPParsley:
This PR looks good.

@pfrenssen, note that atm we're only seeing deprecations when they appear during phpunit testing so all code should be covered. We could easily add/swap to https://github.com/mglaman/drupal-check (or https://github.com/phpstan/phpstan) to make sure we catch all deprecations for D8 only.

I suppose we should we have a new major version for Drupal 9 (where we break Drupal 8 compatibility)?

og.permissions.yml
```diff
@@ -1,4 +1,4 @@
-administer group:
+administer organic groups:
   title: 'Administer Organic groups permissions'
```
MPParsley: This permission allows much more than managing permissions, no?
```suggestion
  title: 'Administer Organic groups'
```

pfrenssen:
PHPStan is a good idea, or we can uncomment the section in phpunit.xml.dist that disables the deprecation check on D9 builds. But let's discuss that in the right issue: #663

MPParsley (approved):
With the exception of the deprecation this looks good.

MPParsley:
We should also require php 7.2, travis only tests 7.2, no longer 7.1
        "php": "^7.2",

pfrenssen:
The PHP 7.2 requirement has been handled in #665. So this is good to go, thanks!

PR #664, opened 2020-07-17
Do not check for deprecated code on D9

### LINKED ISSUE #663: Do not fail 9.x.x builds on deprecation warnings
We have introduced a check for deprecation warnings in our CI pipelines when we were preparing the D9 compatibility. This is causing problems now that D9 is going full steam ahead and has started to deprecate things in preparation for Drupal 10.

Some of these deprecations cannot be resolved in a way that would work both on Drupal 8 and Drupal 9 (for example https://www.drupal.org/node/3154407 - this requires <code>symfony/event-dispatcher</code> version 4.x which is not compatible with Drupal 8).

There are two possibilities: either we start to maintain separate D8 and D9 versions, or we ignore the deprecation warnings for now on D9 builds). I would propose to do the latter.

pfrenssen:
Fixes #663

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "drupal/core": "^8.8 || ^9.0"
     },
     "require-dev": {
-        "drupal/coder": "^8.2"
+        "drupal/coder": "^8.2",
+        "phpunit/phpunit": "^7 || ^8"
     },
     "minimum-stability": "RC"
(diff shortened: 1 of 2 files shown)


pfrenssen:
Yay green! =)

MPParsley (approved):

PR #665, opened 2020-07-17
Drop support for php 7.1

MPParsley:
Php 7.1 is no longer supported
https://www.php.net/supported-versions.php
Travis no longer runs test for php 7.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "source": "https://git.drupalcode.org/project/og"
     },
     "require": {
-        "php": "^7.1.0",
+        "php": "^7.2",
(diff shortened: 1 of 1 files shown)


pfrenssen:
Drupal still officially supports PHP 7.0+ and we put this on PHP 7.1 since we are using the `void` return type which is PHP 7.1+.

But yes I am in favor for this since it is no longer supported and untested. At my job we still have a number of projects running on PHP 7.1 which are in the process of being updated, but they can stay on the current version of OG or fork it.

The test failures should be fixed once #664  is in.

pfrenssen (approved):
Can be merged once the tests are green, after merging #664

Issue #658, opened 2020-07-10
Clean up confusion between 'edit', 'update' and 'update group' group level permissions


pfrenssen:
In `OgAccess` there is a mysterious section of code:

```
    // Update group special permission. At this point, the operation should have
    // already been handled by Og. If the operation is simply 'edit'
    // (or 'update' for content entities), it is referring to the current group,
    // so we have to map it to the special permission.
    if (in_array($operation, ['update', 'edit'])) {
      $operation = OgAccess::UPDATE_GROUP_PERMISSION;
    }
```

So what is happening here is that we are conflating two equivalent entity operations `edit` and `update` and mapping them on an OG specific group level permission `OgAccess::UPDATE_GROUP_PERMISSION` which is `update group`.

This can be traced back to the early days of the D8 port when the access control was initially implemented (ref https://github.com/amitaibu/og/pull/250). In that PR there were already some concerns raised about this section, but there was test coverage that seemingly proved that this was necessary.

I did some testing and it turns out that the `edit` operation in fact does not exist. The tests were trying to prove this by visiting an edit form, but stepping through this code it turns that edit forms also use the `update` operation. The operation is derived from the `_entity_access` route requirement which is `{$entity_type_id}.update` for edit forms (ref. [DefaultHtmlRouteProvider::getEditFormRoute()](https://git.drupalcode.org/project/drupal/-/blob/2ee59bef8f25f529f25d9c6ae3fd856fac3c6cb9/core/lib/Drupal/Core/Entity/Routing/DefaultHtmlRouteProvider.php#L277)).

So we can drop `edit` here, since core will always pass in the `update` operation for the entity CRUD update operation.

Then there remains the translation to `update group`. This makes sense in that if we are checking if a user has access to the `update` entity operation then we need to check if the user has the `update group` permission in the group. But the implementation is wrong, it is mixing up entity operations and permissions. Instead of turning the `update` entity operation into the `update group` permission, we should instead do something like the following:

```
// Grant access to the update operation if the user has permission to update the group.
if ($operation === 'update') {
  $membership = $this->membershipManager->getMembership($group, $user->id());
  if ($membership->hasPermission(OgAccess::UPDATE_GROUP_PERMISSION)) {
    return AccessResult::allowed()->addCacheableDependency($cacheable_metadata);
  }
}
```

[...]

PR #671, opened 2020-07-28
User access check should no longer accept operations, only permissions

### LINKED ISSUE #659: `OgAccess::userAccess()` doesn't differentiate between entity operations and group level permissions
There is a problem with the architecture of `OgAcces::userAccess()` in that it doesn't make a distinction between two concepts in core that are depending on eachother but are fundamentally different:

## Entity operations

Entity operations are basic actions that can be taken on entity level. They are mostly used for CRUD level operations like `create`, `view`, `update` and `delete`, but there is also a widespread operation `view label` and entity types are free to check for custom operations. For example the core Workspaces module has an additional operation `publish`.

## Permissions

Permissions are module defined privileges that can be assigned to roles, and granted to users. They allow access to functionalities provided by the module. They might be related to entities (e.g. `delete own article content`) but they can also be giving access to some functionality which is completely unrelated to entities (e.g. `access site in maintenance mode`).

## How are they related?

When a piece of code wants to perform an entity operation on behalf of a user they will need to check if the user has the rights to do this. For example when access is checked for an entity edit form controller we will execute code similar to the following:

```
if ($entity->access('update', $user)) {
  return AccessResult::allowed();
}
```

The call to `$entity->access()` will invoke the correct `AccessControlHandler` through the entity type manager, and inside the access control handler (or in access hooks that are fired by the handler) there will be code that checks the user's permissions that are relevant for the `update` entity operation. For example:

```
  $type = $node->bundle();
  switch ($operation) {
    case 'update':
      $access = AccessResult::allowedIfHasPermission($account, 'edit any ' . $type . ' content');
      break;
  }
```

This demonstrates how entity operations and permissions are related. Note that they are still distinct concepts: the operation `update` is related to `edit any article content` but at no point will we check if the user has an `update` permission, because this permission doesn't exist. Also we will not check for an `edit any article content` entity operation.

## How about OgAccess?

operation-vs-permission

According to the current documentation and method signature `OgAccess::userAccess()` is designed to determine access to entity operations. But it also accepts permissions instead of operations and handles both in the same way.

Example call to check access on an entity operation:

```
  public function userAccessEntity($operation, EntityInterface $entity, AccountInterface $user = NULL): AccessResultInterface {
    // ...
    $user_access = $this->userAccess($entity, $operation, $user);
```

Example call to check access on a user permission (from `GroupSubscribeForm`):

```
  public function isStateActive() {
    // ...
    $skip_approval = $this->ogAccess->userAccess($group, 'subscribe without approval', $user)->isAllowed();
```

## How did the mixup happen?

The code is based on the D7 function `og_user_access()` which does not suffer from this problem. `og_user_access()` is only for checking user permissions, and this is clearly documented. There is nothing in the code that deals with entity operations in any way. When this function was ported to D8, it was done as a proof of concept in scope of checking entity access (inside `og_entity_access()`) and the mixup was present in the initial commit of this class (ref commit [d24cd80](https://github.com/Gizra/og/commit/d24cd808c390a0a67748ce87363c3ade1a4ff754)).

What is interesting is that in this initial work, the only problem is in the argument naming and documentation: the method takes an argument `$operation` (instead of `$permission`) and it is documented as `The entity operation being checked for.`. The code itself is still 100% fine at this point: it is dealing only with permissions.

But as the port progressed, we started to call this method to check entity operations (after all, it says that this is its purpose!) and gradually the code was morphed into some kind of Frankenmethod that delivers "good enough" results to pass tests for both entity operations and permissions!

## How to solve it

I propose to revert to the original intention as it was designed in D7: we should use this method exclusively for permissions. Let's fix the documentation and variable naming, and strip out the workarounds to make this work with entity operations. All code calling this to check for entity operations should be moved to entity hooks, and call in here with a relevant permission such as "edit group".

pfrenssen:
Fixes #659


src/OgAccess.php
```diff
@@ -151,14 +153,6 @@ public function userAccess(EntityInterface $group, $operation, AccountInterface
...
     }
 
-    // Update group special permission. At this point, the operation should have
-    // already been handled by Og. If the operation is simply 'edit'
-    // (or 'update' for content entities), it is referring to the current group,
-    // so we have to map it to the special permission.
-    if (in_array($operation, ['update', 'edit'])) {
-      $operation = OgAccess::UPDATE_GROUP_PERMISSION;
-    }
-
```
pfrenssen, line 161: This is contamination that has been added to handle entity operations and translate them to a permission. Since in this PR we will no longer handle operations this can be removed. The test for this is `GroupUpdateTest` which is also being removed in this PR.

This change is also fixing #658

tests/src/Kernel/Access/OgEntityAccessTest.php
```diff
@@ -315,10 +317,6 @@ public function testAccess() {
 
     $this->assertTrue($og_access->userAccess($this->group1, 'some_perm', $this->user3)->isAllowed());
 
-    // A member with permission to update the group. The operation edit is
-    // passed to the userAccess method.
-    $this->assertTrue($og_access->userAccess($this->group1, 'edit', $this->user4)->isAllowed());
-
```
pfrenssen, line 321: This one line was the only test that covers handling an entity operation through this method. Everything else is about permissions. This is a good indication that we are on the right path with this cleanup.

MPParsley (approved):
@pfrenssen could you have a look at the impact this will have on #231?

If I remember correctly that PR also used operations instead of permissions.

pfrenssen:
> @pfrenssen could you have a look at the impact this will have on #231?
> 
> If I remember correctly that PR also user operations instead of permissions.

It looks like that PR is not impacted. There are 4 calls to `OgAccess::userAccess()` but all check for a user permission (e.g. `manage members`) and not an entity operation.

PR #673, opened 2020-07-29
Split up OgAccess::userAccessEntity()

### LINKED ISSUE #672: Split up OgAccess::userAccessEntity() into two methods
This is part of the `OgAccess` revamp.

As detailed in earlier issues some inconsistencies were introduced in the documentation of the `OgAccess` methods during the early days of the D7 > D8 port and over time we have deviated from the original functionality as designed in D7.

In short, the original D7 functions `og_user_access()` and `og_user_access_entity()` were designed to check user permissions on groups and group content, but during the early port the methods were accidentally documented as dealing with entity operations rather than user permissions. Later on in the D8 port we have been confused by this wrong documentation and have introduced code that deals with entity operations. The current state is that these methods deal with both concepts but this is incorrect and problematic. For more details see #659.

**Current state**

`OgAccess::userAccessEntity()` accepts both user permissions (e.g. `subscribe without approval`) and entity operations (e.g. `create` or `update`). The code has been mangled to work with both. This has resulted in some strange code acrobatics, like calling into `::userAccess()` twice.

There is test coverage for both scenarios, and somehow tests pass for both. The coverage for the entity operations case is very extensive, but for user permissions is basic.

**Scope**

- [x] Split off the code that deals with entity operations to a new method `OgAccess::userAccessEntityOperation()`.
- [x] Reduce the original method to align closely to the D7 functionality, only dealing with user permissions.
- [x] Fix the documentation and ensure the difference between both is completely clear.
- [x] Ensure existing tests are calling into the respective functions and all tests are passing.

pfrenssen:
Fixes #672

### DIFF
--- a/src/OgAccessInterface.php
+++ b/src/OgAccessInterface.php
@@ -62,7 +62,7 @@ interface OgAccessInterface {
    * @return \Drupal\Core\Access\AccessResultInterface
    *   An access result object.
    */
-  public function userAccessEntity($operation, EntityInterface $entity, AccountInterface $user = NULL): AccessResultInterface;
+  public function userAccessEntity(string $operation, EntityInterface $entity, AccountInterface $user = NULL): AccessResultInterface;
 
   /**
    * Checks access for entity operations on group content entities.
@@ -89,7 +89,7 @@ interface OgAccessInterface {
    *
    * @see \Drupal\og\PermissionManager::getEntityOperationPermissions()
    */
-  public function userAccessGroupContentEntityOperation($operation, EntityInterface $group_entity, EntityInterface $group_content_entity, AccountInterface $user = NULL): AccessResultInterface;
+  public function userAccessGroupContentEntityOperation(string $operation, EntityInterface $group_entity, EntityInterface $group_content_entity, AccountInterface $user = NULL): AccessResultInterface;
 
   /**
    * Resets the static cache.
--- a/src/OgAccessInterface.php
+++ b/src/OgAccessInterface.php
@@ -50,7 +50,41 @@ interface OgAccessInterface {
   public function userAccess(EntityInterface $group, string $permission, AccountInterface $user = NULL, bool $skip_alter = FALSE): AccessResultInterface;
 
   /**
-   * Check if a user has access to a permission on a certain entity context.
+   * Determines whether a user has a group permission in a given entity.
+   *
+   * This does an exhaustive, but slow, check to discover whether access can be
+   * granted and works both on groups and group content. It will iterate over
+   * all groups that are associated with the entity and do a permission check on
+   * each group. If a passed in entity is both a group and group content, it
+   * will return a positive result if the user has the requested permission in
+   * either the entity itself or its parent group(s).
+   *
+   * In case you know the specific group you want to check access for then it is
+   * recommended to use the faster ::userAccess().
+   *
+   * @param string $permission
+   *   The name of the OG permission being checked. This includes both group
+   *   level permissions such as 'subscribe without approval' and group content
+   *   entity operation permissions such as 'edit own article content'.
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity object. This can be either a group or group content entity.
+   * @param \Drupal\Core\Session\AccountInterface $user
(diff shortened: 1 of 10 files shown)


pfrenssen:
This is ready but since this is built on top of #671 I will keep this in draft until the other PR is accepted.

pfrenssen:
Setting back to draft since this needs more work. I am running tests on my "big" project and there is a regression. `::userAccessEntity()` was being called from `og_entity_access()` with an entity operation, both on group content and group entities. It should only do this for group content from now on.

We have currently no full support for entity operation permissions on groups, only for the `update` operation (through the `UPDATE_GROUP_PERMISSION` `GroupPermission`). I will make a followup to add also a delete permission.

pfrenssen:
Created followup #674 but I had another look and we can still retain the functionality in scope of this, we simply need to check if we are handling a group entity and if this is the case we can translate `update` to `UPDATE_GROUP_PERMISSION` and delegate to `::userAccess()`. It will be solved in a couple lines and there will not be a regression for now.

I propose to handle the test coverage for this in #674. We don't have the coverage right now and this would be out of scope.

pfrenssen:
I have added the code which solves the regression and I think this is now complete.

I will keep this in draft until I have finished work on #674 - that issue will complete the entity operations for group entities by adding the delete operation. Most importantly it will also add the necessary test coverage which I think is out of scope for this PR.

I also want to run the extensive test suite on my work project that has discovered this regression.

pfrenssen:
I am near finishing work on #674. Here is the coverage for the regression on `hook_entity_access()`: https://github.com/Gizra/og/pull/681/files#diff-103d019edfa64fc25b94e05082e8bba8R332

This is now ready for review.

src/OgAccessInterface.php
```diff
@@ -50,31 +52,68 @@ interface OgAccessInterface {
...
+   * Determines whether a user has a group permission in a given entity.
+   *
+   * This does an exhaustive, but slow, check to discover whether access can be
+   * granted and works both on groups and group content. It will iterate over
+   * all groups that are associated with the entity and do a permission check on
+   * each group. If a passed in entity is both a group and group content, it
+   * will return a positive result if the user has the requested permission in
+   * either the entity itself or its parent group(s).
+   *
+   * In case you know the specific group you want to check access for then it is
```
MPParsley, line 64: Should we add an @see here?

MPParsley (approved):

src/OgAccessInterface.php
pfrenssen: Sure! Good idea since this will be clickable in the API documentation on drupal.org. I will add this for both methods.

pfrenssen:
@MPParsley Thanks for reviewing! I addressed the remark. I will wait for test results to come back green before merging. I can then "unlock" the next PR.

This `OgAccess` refactoring is coming to an end soon, I have finished work on the final 2 PRs (#681 and #684) and I am now working on integrating them into my project so I can run our extensive test suite.

I can also still as part of this work write the release notes for the next release since this is breaking backwards compatibility. We can still break B/C as long as we're in alpha but I would like to inform our users so that they will know what to expect when updating.

MPParsley:
Thanks for the update!

I'm mostly worried about the impact on the ongoing PRs for [beta blockers](https://github.com/Gizra/og/labels/Beta%20blocker) (e.g. #231, #242, #570, ...).

pfrenssen:
The impact is due to having split the entity access check into separate checks for user permissions and entity operations. So existing code that is checking entity operations will need to call the other method now.

Here is an example how this now looks in my project code: (ref. https://github.com/ec-europa/joinup-dev/pull/2184/files#diff-f26d2164853bef0da7d08750148704f5L86)

```
-    return $this->ogAccess->userAccessEntity('create', $this->createNewAssetRelease($rdf_entity), $account);
+    return $this->ogAccess->userAccessEntityOperation('create', $this->createNewAssetRelease($rdf_entity), $account);
```

Also if people are implementing `hook_og_access_alter()` and are altering the results expecting entity operations to be passed in, they will need to use `hook_og_access_entity_operations_alter()` now. Example: https://github.com/ec-europa/joinup-dev/pull/2184/files#diff-0c9ed0667150da0d6130597e6bb8789bR179

PR #681, opened 2020-07-31
Provide "delete" entity operation access check on group entities

### LINKED ISSUE #674: Provide "delete" entity operation access checks on group entities
Currently we only have full entity operation access checks for group content (see for example `OgEventSubscriber::provideDefaultNodePermissions()` which supplies these for node group content).

We have a group permission called `update group` but ~this is not currently enforced and~ (edit: is enforced now, ref. #673) test coverage for this is incomplete (ref. a user with this permission (`user4`) is created in `OgEntityAccessTest` but the test doesn't actually check that this user has edit access and other users don't.

We are also lacking a `delete group` permission. I think a `create group` permission is not needed, since this is controlled by Drupal permissions (a newly created group doesn't have any members yet so we cannot control this using an OG permission). We are ignoring view permissions too.

In order to fulfill this I propose:

- [x] Create a new `GroupPermission` to delete groups, apply it to the group admin by default, and treat it in the same way as the `UPDATE_GROUP_PERMISSION`.
- [x] Extend the tests, not only on the return value of the access checks like is being done in `OgEntityAccessTest` but also in scenarios where groups are actually being updated and deleted, i.e. through `hook_entity_access()`.

pfrenssen:
Fixes #674

This is building on top of #673 as part of the OgAccess refactoring.


pfrenssen:
#673 has been merged, this is now ready for review.

src/EventSubscriber/OgEventSubscriber.php
```diff
@@ -91,6 +91,12 @@ public function provideDefaultOgPermissions(PermissionEventInterface $event) {
         'description' => $this->t('Edit the group entity.'),
         'default roles' => [OgRoleInterface::ADMINISTRATOR],
       ]),
+      new GroupPermission([
+        'name' => OgAccess::DELETE_GROUP_PERMISSION,
+        'title' => $this->t('Delete group'),
+        'description' => $this->t('Delete the group.'),
```
MPParsley: Either we add group here:

```suggestion
        'description' => $this->t('Delete the group entity.'),
```

Or we remove group from
`        'description' => $this->t('Edit the group entity.'),
`

src/EventSubscriber/OgEventSubscriber.php
```diff
@@ -91,6 +91,12 @@ public function provideDefaultOgPermissions(PermissionEventInterface $event) {
         'description' => $this->t('Edit the group entity.'),
         'default roles' => [OgRoleInterface::ADMINISTRATOR],
       ]),
+      new GroupPermission([
+        'name' => OgAccess::DELETE_GROUP_PERMISSION,
```
MPParsley, line 96: Do we care about the order? "D"elete comes before "U"pdate in most other instances.

tests/src/Kernel/Access/GroupLevelAccessTest.php
```diff
@@ -0,0 +1,450 @@
...
+
+    // Create a group and associate with the group owner.
+    $this->group = EntityTest::create([
+      'type' => $this->groupBundle,
+      'name' => $this->randomString(),
+      'user_id' => $this->ownerUser->id(),
+    ]);
+    $this->group->save();
+
+    // A non-member.
```
MPParsley:
```suggestion
    // Create a non-member.
```

tests/src/Kernel/Access/GroupLevelAccessTest.php
```diff
@@ -0,0 +1,450 @@
...
+    $this->nonMemberUser = User::create(['name' => $this->randomString()]);
+    $this->nonMemberUser->save();
+
+    // Admin user.
```
MPParsley:
```suggestion
    // Create an admin user.
```

tests/src/Kernel/Access/GroupLevelAccessTest.php
```diff
@@ -0,0 +1,450 @@
...
+    $this->adminUser = User::create(['name' => $this->randomString()]);
+    $this->adminUser->save();
+
+    // The administrator role is added automatically when the group is created.
```
MPParsley:
```suggestion
    // Get the default administrator role that is created automatically when the
    // group is created and assign it to the test user.
```

tests/src/Kernel/Access/GroupLevelAccessTest.php
```diff
@@ -0,0 +1,450 @@
...
+    $admin_role = OgRole::loadByGroupAndName($this->group, OgRoleInterface::ADMINISTRATOR);
+
+    $membership = Og::createMembership($this->group, $this->adminUser);
+    $membership
+      ->addRole($admin_role)
+      ->save();
+
+    // Create a second administration role and assign it to a test user. This is
+    // a supported use case: it is possible to have multiple administration
+    // roles.
```
MPParsley:
```suggestion
    // Create another administrator role and assign it to a second test user.
    // This is a supported use case: it is possible to have multiple
    // administration roles.
```

tests/src/Kernel/Access/GroupLevelAccessTest.php
```diff
@@ -0,0 +1,450 @@
...
+      ->setName($this->randomMachineName())
+      ->setLabel($this->randomString())
+      ->setGroupType($this->group->getEntityTypeId())
+      ->setGroupBundle($this->groupBundle)
+      ->save();
+
+    $membership = Og::createMembership($this->group, $user);
+    $membership
+      ->addRole($role_without_permissions)
+      ->save();
```
MPParsley: I'd create a helper function to group this:

```
  function createTestUser($role) {
    // Create a user which is a member and has a role without any permissions.
    $user = User::create(['name' => $this->randomString()]);
    $user->save();

    $membership = Og::createMembership($this->group, $user);
    $membership
      ->addRole($role)
      ->save();
    
    return $user;
  }

```

tests/src/Kernel/PermissionEventTest.php
```diff
@@ -130,12 +130,13 @@ public function permissionEventDataProvider() {
...
       'add user',
       OgAccess::ADMINISTER_GROUP_PERMISSION,
+      OgAccess::DELETE_GROUP_PERMISSION,
+      OgAccess::UPDATE_GROUP_PERMISSION,
       'approve and deny subscription',
       'manage members',
       'administer permissions',
       'subscribe without approval',
       'subscribe',
-      'update group',
```
MPParsley, line 138: Nice catch

MPParsley (approved):

src/EventSubscriber/OgEventSubscriber.php
pfrenssen: OK I will put them alphabetically. I was torn between alphabetical ordering and "CRUD ordering".

tests/src/Kernel/Access/GroupLevelAccessTest.php
pfrenssen: Great idea, this will make the setup more compact and easy to follow. I'll also make a helper method to create OG roles.

pfrenssen:
Solved all the remarks, I assume this is good to go now since it has been approved.

PR #684, opened 2020-08-02
Group content entity operations ~~hook~~ event

### LINKED ISSUE #675: Provide a hook for altering access on group content entity operations
We have an alter hook in `OgAccess::userAccess()` that allows to alter user permission based access. In a similar vein we should allow to alter based on group content entity operations.

At the moment this is not impossible, since the group content entity operation access check also passes through `::userAccess()`, but this puts the burden on the hook implementation to do the reverse mapping of the user permission (e.g. `edit own article content`) back to the operation (e.g. `update`) which can become difficult if group content exists for many different entity types.

I propose that for now we also introduce a `hook_og_user_access_entity_operation_alter()` hook so we are consistent with `hook_og_user_access_alter()`, even though event subscribers are superior. Switching to event subscribers is out of scope for this issue.

pfrenssen:
Fixes #675.

This is building on top of #681.

### DIFF
--- a/og.api.php
+++ b/og.api.php
@@ -7,6 +7,7 @@
 
 declare(strict_types = 1);
 
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Access\AccessResultInterface;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Entity\EntityPublishedInterface;
@@ -23,14 +24,14 @@ use Drupal\og\OgAccess;
  * @param array $permissions
  *   The list of group level permissions, passed by reference.
  * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
- *   The cache metadata, passed by reference.
+ *   The cache metadata.
  * @param array $context
  *   An associative array containing contextual information, with keys:
  *   - 'permission': The group level permission being checked, as a string.
  *   - 'group': The group entity on which the permission applies.
  *   - 'user': The user account for which access is being determined.
  */
-function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cacheable_metadata, array $context): void {
+function hook_og_user_access_alter(array &$permissions, CacheableMetadata $cacheable_metadata, array $context): void {
   // This example implements a use case where a custom module allows site
   // builders to toggle a configuration setting that will prevent groups to be
   // deleted if they are published.
@@ -38,7 +39,7 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cach
   $config = \Drupal::config('mymodule.settings');
 
   // Check if the site is configured to allow deletion of published groups.
-  $published_groups_can_be_deleted = \Drupal::config('mymodule')->get('delete_published_groups');
+  $published_groups_can_be_deleted = $config->get('delete_published_groups');
 
   // If deletion is not allowed and the group is published, revoke the
   // permission.
@@ -61,7 +62,7 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cach
  * @param \Drupal\Core\Access\AccessResultInterface $access_result
  *   The access result being altered.
  * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
- *   The cache metadata, passed by reference.
+ *   The cache metadata.
  * @param array $context
  *   An associative array containing contextual information, with keys:
  *   - 'operation': The entity operation being performed on the group content.
@@ -70,7 +71,7 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cach
  *      performed.
  *   - 'user': The user account for which access is being determined.
  */
-function hook_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata &$cacheable_metadata, array $context): void {
+function hook_og_user_access_entity_operation_alter(AccessResultInterface $access_result, CacheableMetadata $cacheable_metadata, array $context): void {
   // This example implements a use case where a custom module allows site
   // builders to toggle a configuration setting that will allow users with the
   // site wide 'edit and delete comments in all groups' permission to edit and
@@ -86,10 +87,10 @@ function hook_og_user_access_entity_operation_alter(AccessResultInterface &$acce
   // access to the 'update' and 'delete' operations on comment entities.
   $is_comment = $group_content->getEntityTypeId() === 'comment';
   $user_can_moderate_comments = $user->hasPermission('edit and delete comments in all groups');
-  $comment_moderation_is_enabled = \Drupal::config('mymodule')->get('comment_moderation_enabled');
+  $comment_moderation_is_enabled = $config->get('comment_moderation_enabled');
 
   if ($is_comment && $user_can_moderate_comments && $comment_moderation_is_enabled) {
-    $access_result = new AccessResultAllowed();
+    $access_result = AccessResult::allowed();
   }
 
   // Since our access result depends on our custom module configuration, we need
--- a/og.api.php
+++ b/og.api.php
@@ -1,0 +1,102 @@
+<?php
+
+/**
+ * @file
+ * Hooks provided by the Organic Groups module.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Core\Access\AccessResultInterface;
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Entity\EntityPublishedInterface;
+use Drupal\og\OgAccess;
+
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
+/**
+ * Allows modules to alter group level permissions.
+ *
+ * @param array $permissions
+ *   The list of group level permissions, passed by reference.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
+ *   The cache metadata, passed by reference.
+ * @param array $context
+ *   An associative array containing contextual information, with keys:
+ *   - 'permission': The group level permission being checked, as a string.
+ *   - 'group': The group entity on which the permission applies.
+ *   - 'user': The user account for which access is being determined.
+ */
+function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cacheable_metadata, array $context): void {
+  // This example implements a use case where a custom module allows site
+  // builders to toggle a configuration setting that will prevent groups to be
+  // deleted if they are published.
+  // Retrieve the module configuration.
+  $config = \Drupal::config('mymodule.settings');
+
+  // Check if the site is configured to allow deletion of published groups.
+  $published_groups_can_be_deleted = \Drupal::config('mymodule')->get('delete_published_groups');
+
+  // If deletion is not allowed and the group is published, revoke the
+  // permission.
+  $group = $context['group'];
+  if ($group instanceof EntityPublishedInterface && !$group->isPublished() && !$published_groups_can_be_deleted) {
+    $key = array_search(OgAccess::DELETE_GROUP_PERMISSION, $permissions);
+    if ($key !== FALSE) {
+      unset($permissions[$key]);
+    }
+  }
+
+  // Since our access result depends on our custom module configuration, we need
+  // to add it to the cache metadata.
+  $cacheable_metadata->addCacheableDependency($config);
+}
+
+/**
+ * Allows to alter access to entity operations performed on group content.
+ *
+ * @param \Drupal\Core\Access\AccessResultInterface $access_result
+ *   The access result being altered.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
+ *   The cache metadata, passed by reference.
+ * @param array $context
+ *   An associative array containing contextual information, with keys:
+ *   - 'operation': The entity operation being performed on the group content.
+ *   - 'group': The group entity to which the group content belongs.
+ *   - 'group_content': The group content entity upon which the operation is
+ *      performed.
+ *   - 'user': The user account for which access is being determined.
+ */
+function hook_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata &$cacheable_metadata, array $context): void {
+  // This example implements a use case where a custom module allows site
+  // builders to toggle a configuration setting that will allow users with the
+  // site wide 'edit and delete comments in all groups' permission to edit and
+  // delete all comments in all groups, even if they are not a group member.
+  /** @var \Drupal\Core\Session\AccountProxyInterface $user */
+  $user = $context['user'];
+  $group_content = $context['group_content'];
+
+  // Retrieve the module configuration.
+  $config = \Drupal::config('mymodule.settings');
+
+  // If comment moderation is allowed and the user has the permission, grant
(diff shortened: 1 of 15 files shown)


pfrenssen:
This is the final PR in the `OgAccess` refactoring. All previous ones have been merged so this is now ready for review.

src/OgAccess.php
```diff
@@ -366,7 +368,16 @@ public function userAccessGroupContentEntityOperation(string $operation, EntityI
...
 
-    return AccessResult::neutral()->addCacheableDependency($cacheable_metadata);
+    // Let modules alter the access result.
+    $context = [
+      'operation' => $operation,
+      'group' => $group_entity,
+      'group_content' => $group_content_entity,
+      'user' => $user,
+    ];
+    $this->moduleHandler->alter('og_user_access_entity_operation', $access_result, $cacheable_metadata, $context);
```
MPParsley: Should we document this on og.api.php?

tests/modules/og_test/og_test.module
```diff
@@ -39,3 +44,23 @@ function og_test_entity_insert(EntityInterface $entity) {
     $membership->save();
   }
 }
+
+/**
+ * Implements hook_og_user_access_entity_operation_alter().
+ */
+function og_test_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata &$cacheable_metadata, $context): void {
+  if (\Drupal::state()->get('og_test_group_content_entity_operation_access_alter', FALSE)) {
```
MPParsley, line 52: Perhaps we can cleanup the state key in a hook_uninstall()?

MPParsley (approved):

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+ * @{
+ */
+
+/**
+ * Allows modules to alter group level permissions.
+ *
+ * @param array $permissions
+ *   The list of group level permissions, passed by reference.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
+ *   The cache metadata, passed by reference.
```
MPParsley: Does it make sense to pass an object by reference?
```suggestion
 *   The cache metadata.
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+ * @param array $context
+ *   An associative array containing contextual information, with keys:
+ *   - 'operation': The entity operation being performed on the group content.
+ *   - 'group': The group entity to which the group content belongs.
+ *   - 'group_content': The group content entity upon which the operation is
+ *      performed.
+ *   - 'user': The user account for which access is being determined.
+ */
+function hook_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata &$cacheable_metadata, array $context): void {
```
MPParsley: Objects are passed by reference by default, not sure if it makes sense to explicitly do so here.
```suggestion
function hook_og_user_access_entity_operation_alter(AccessResultInterface $access_result, CacheableMetadata $cacheable_metadata, array $context): void {
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cacheable_metadata, array $context): void {
+  // This example implements a use case where a custom module allows site
+  // builders to toggle a configuration setting that will prevent groups to be
+  // deleted if they are published.
+  // Retrieve the module configuration.
+  $config = \Drupal::config('mymodule.settings');
+
+  // Check if the site is configured to allow deletion of published groups.
+  $published_groups_can_be_deleted = \Drupal::config('mymodule')->get('delete_published_groups');
```
MPParsley: Assuming the key lives in mymodule.settings I'd reuse the $config object. Same remark below.
```suggestion
  $published_groups_can_be_deleted = $config->get('delete_published_groups');
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+  $group_content = $context['group_content'];
+
+  // Retrieve the module configuration.
+  $config = \Drupal::config('mymodule.settings');
+
+  // If comment moderation is allowed and the user has the permission, grant
+  // access to the 'update' and 'delete' operations on comment entities.
+  $is_comment = $group_content->getEntityTypeId() === 'comment';
+  $user_can_moderate_comments = $user->hasPermission('edit and delete comments in all groups');
+  $comment_moderation_is_enabled = \Drupal::config('mymodule')->get('comment_moderation_enabled');
```
MPParsley: Use $config?
```suggestion
  $comment_moderation_is_enabled = $config->get('comment_moderation_enabled');
```

tests/modules/og_test/og_test.module
```diff
@@ -39,3 +44,23 @@ function og_test_entity_insert(EntityInterface $entity) {
     $membership->save();
   }
 }
+
+/**
+ * Implements hook_og_user_access_entity_operation_alter().
+ */
+function og_test_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata &$cacheable_metadata, $context): void {
```
MPParsley: Seems like I missed this the first time, do we care here if we pass exactely the same pointer? I assume it's sufficient if the argument points to the same instance.
```suggestion
function og_test_og_user_access_entity_operation_alter(AccessResultInterface $access_result, CacheableMetadata $cacheable_metadata, $context): void {
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+ *   The list of group level permissions, passed by reference.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
+ *   The cache metadata, passed by reference.
+ * @param array $context
+ *   An associative array containing contextual information, with keys:
+ *   - 'permission': The group level permission being checked, as a string.
+ *   - 'group': The group entity on which the permission applies.
+ *   - 'user': The user account for which access is being determined.
+ */
+function hook_og_user_access_alter(array &$permissions, CacheableMetadata &$cacheable_metadata, array $context): void {
```
MPParsley:
```suggestion
function hook_og_user_access_alter(array &$permissions, CacheableMetadata $cacheable_metadata, array $context): void {
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+  $cacheable_metadata->addCacheableDependency($config);
+}
+
+/**
+ * Allows to alter access to entity operations performed on group content.
+ *
+ * @param \Drupal\Core\Access\AccessResultInterface $access_result
+ *   The access result being altered.
+ * @param \Drupal\Core\Cache\CacheableMetadata $cacheable_metadata
+ *   The cache metadata, passed by reference.
```
MPParsley:
```suggestion
 *   The cache metadata.
```

og.api.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+/**
+ * @file
+ * Hooks provided by the Organic Groups module.
+ */
+
+declare(strict_types = 1);
+
+use Drupal\Core\Access\AccessResultInterface;
```
MPParsley: We're using AccessResultAllowed later on:
```suggestion
use Drupal\Core\Access\AccessResultAllowed;
use Drupal\Core\Access\AccessResultInterface;
```

But I 'd suggest using AccessResult:allowed()
```suggestion
use Drupal\Core\Access\AccessResult;
use Drupal\Core\Access\AccessResultInterface;
```

og.api.php
```diff
@@ -0,0 +1,102 @@
...
+  $config = \Drupal::config('mymodule.settings');
+
+  // If comment moderation is allowed and the user has the permission, grant
+  // access to the 'update' and 'delete' operations on comment entities.
+  $is_comment = $group_content->getEntityTypeId() === 'comment';
+  $user_can_moderate_comments = $user->hasPermission('edit and delete comments in all groups');
+  $comment_moderation_is_enabled = \Drupal::config('mymodule')->get('comment_moderation_enabled');
+
+  if ($is_comment && $user_can_moderate_comments && $comment_moderation_is_enabled) {
+    $access_result = new AccessResultAllowed();
```
MPParsley: Probably just a matter of taste:
```suggestion
    $access_result = AccessResult::allowed();
```

og.api.php
pfrenssen: I had the understanding that in PHP objects are not passed by reference, but rather you get a copy of the object instance, so you are working with a new object reference that points to the same memory location. If you then "shadow" the variable by instantiating it again then you lose the link with the original object. If you use a reference then your local variable will modify the original object instance.

So I think for the cache metadata we can remove the reference and force the user to work with the original object rather than replace it, but I think we need to keep the reference for the access result, since we need to be able to replace it with a new object.

I can be wrong though, I am nowadays more working with low level languages than with PHP and I'm possibly mixing things up. I will try this by removing the reference from the test and see if it fails.

MPParsley:
In fact the argument is passed by value as a variable that contains the reference to an object (much like pointers). But when passing objects it shouldn't really matter if you have one or two variables as long as they point to the same object.

edit: unless if you want to change the original object reference (e.g. have it reference another object).

og.api.php
MPParsley: `I had the understanding that in PHP objects are not passed by reference`
Indeed. They're passed by value.

`but rather you get a copy of the object instance`
Not anymore, this used to be true in PHP 4 but as of PHP 5 you get a copy of the reference.

`So I think for the cache metadata we can remove the reference and force the user to work with the original object rather than replace it, but I think we need to keep the reference for the access result, since we need to be able to replace it with a new object.`

Indeed. I thought you only wanted to modify the object, if you want to replace the original one you do need the original reference to it.

> One of the key-points of PHP 5 OOP that is often mentioned is that "objects are passed by references by default". This is not completely true. This section rectifies that general thought using some examples.
> 
> A PHP reference is an alias, which allows two different variables to write to the same value. **As of PHP 5, an object variable doesn't contain the object itself as value anymore. It only contains an object identifier which allows object accessors to find the actual object. When an object is sent by argument, returned or assigned to another variable, the different variables are not aliases: they hold a copy of the identifier, which points to the same object.**

See https://www.php.net/manual/en/language.oop5.references.php

That said, I wonder if we're doing this the right way? I'm fine with passing the AccessResult and changing it as it's done now but perhaps we could return the AccessResult.
Then the invoker needs to handle it, much like core's EntityAccessControlHandler does: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityAccessControlHandler.php/function/EntityAccessControlHandler%3A%3AprocessAccessHookResults/8.2.x

tests/modules/og_test/og_test.module
```diff
@@ -41,3 +44,23 @@ function og_test_entity_insert(EntityInterface $entity) {
...
+    // groups.
+    /** @var \Drupal\Core\Session\AccountProxyInterface $user */
+    $user = $context['user'];
+    $group_content = $context['group_content'];
+
+    $is_comment = $group_content->getEntityTypeId() === 'comment';
+    $user_can_moderate_comments = $user->hasPermission('edit and delete comments in all groups');
+
+    if ($is_comment && $user_can_moderate_comments) {
+      $access_result = new AccessResultAllowed();
```
MPParsley, line 63: This would probably work too but it looks ugly:
$access_result->orIf(new AccessResultAllowed());

tests/modules/og_test/og_test.module
```diff
@@ -41,3 +44,23 @@ function og_test_entity_insert(EntityInterface $entity) {
     $membership->save();
   }
 }
+
+/**
+ * Implements hook_og_user_access_entity_operation_alter().
+ */
+function og_test_og_user_access_entity_operation_alter(AccessResultInterface $access_result, CacheableMetadata $cacheable_metadata, $context): void {
```
MPParsley: Let's revert the pass by reference
```suggestion
function og_test_og_user_access_entity_operation_alter(AccessResultInterface &$access_result, CacheableMetadata $cacheable_metadata, $context): void {
```

og.api.php
pfrenssen: I like this idea a lot, but then I think it is better to go with an event dispatcher right away. If we go with a normal hook then we have to provide a default implementation (like `og_og_user_access_entity_operation()`) otherwise we would have no way for modules to override a denied access originating from our code. With an event it is much easier to make OG's code run first and all other modules run after it.

pfrenssen: I gave this some thought and I will implement the following:

**AccessEventInterface**

An interface that can be used for all OG access events, so that it can be reused easily in the future for #686. This interface needs to supply common methods:

* Extend `RefineableCacheableDependencyInterface` to be able to supply cache metadata.
* Provide methods to deny and grant access. We will follow core's example that a single deny will overrule any grants.
* Provide methods to retrieve the group and user for which to check access. We can assume that in scope of OG there will always be a group in play when determining access for something.

**GroupContentAccessEventInterface**

This will extend `AccessEventInterface` and will provide additional methods to retrieve the operation and the group content entity.

**AccessEventBase**

Base class with shared methods for this PR + #686 .

**GroupContentAccessEvent**

Implements the remaining methods of `GroupContentAccessEventInterface`.

**OgEventSubscriber::checkGroupContentEntityOperationAccess()**

Will contain the logic that is now inside `OgAccess`.

MPParsley: Sounds like a lot of ~~work~~ fun :-)

tests/modules/og_test/og_test.module
pfrenssen: Resolving this, it is no longer applicable now that the hook has been dropped in favor of an event subscriber.

pfrenssen:
@MPParsley I think this might be good to go but before merging I want to see a green build on my big project to see if there are possible regressions. I need to integrate the latest changes into my project (mainly #402) but I plan to tackle this right now so I will probably have news in ~2 hours.

pfrenssen:
Got a green test: https://github.com/ec-europa/joinup-dev/pull/2184, no regressions, let's get this in. I have one final small improvement waiting in #692.

src/Event/AccessEventBase.php
```diff
@@ -0,0 +1,98 @@
...
+    return $this->user;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAccessResult(): AccessResultInterface {
+    $access = $this->access;
+
+    if ($access instanceof RefinableCacheableDependencyInterface) {
```
amitaibu, line 91: Maybe worth adding a comment - I'm not clear on this part?

pfrenssen: This ensures that our code adheres correctly to our interface and avoids a potential fatal error.

We are returning an `AccessResultInterface` object, and this doesn't offer `::addCacheableDependency()` so we are not supposed to call this method. But in reality all access objects in Drupal core are extending `AccessResult`, which _does_ have this method because it also implements `RefinableCacheableDependencyInterface`. So in typical Drupal code cache metadata is supported and we can call this method. Drupal is full of these kind of discrepancies, and often developers ignore these cases where a method is part of a different interface.

It is not reliable though. It is always possible someone will implement their own `AccessResultInterface` object which does not implement `RefinableCacheableDependencyInterface` and then we get a fatal error. Just putting this simple `if` statement here makes sure we are never going to throw a fatal error if somebody is using a custom implementation.

We in fact are offering extra functionality (cache metadata support) in case the object we are dealing with supports it. This is commonly called "type widening" in PHP. Modern IDE's like PHPStorm have become very good at pointing out these kind of errors.

I created a PR to document this section: https://github.com/Gizra/og/pull/697

PR #692, opened 2020-08-12
Permission based access checks should return neutral if a permission is not granted

### LINKED ISSUE #691: OgAccess should return a neutral access result if a permission is not granted
Currently most methods in `OgAccess` that deal with determining permission based access return `AccessResult::allowed()` if the user has permission, but also return a hard `AccessResult::forbidden()` if the user doesn't have access. While this (mostly) achieves the expected result this comes with some problems:

* It does not work in the same way as access checks in Drupal core. Core returns a neutral result if the user has no permission (see `AccessResult::allowedIfHasPermission()`) so this doesn't meet developer expectations.
* It makes it more difficult to determine the final access result if we are allowing modules to alter the results. We cannot reuse the Drupal core code because in Drupal a `forbidden` is considered to be a hard deny - you can never override a `forbidden` result with an `allowed` result in core so we need to write and maintain our own access resolution code.
* We are effectively losing the ability for a module to do a "hard deny". If a module returns an access deny this can always be turned into an allowed again by a hook implementation that runs at a later point.
* We are also returning access on the very first match, effectively losing the ability for hooks that run at a later moment to deny an earlier allow.

Scope of this is to align OG access result handling with the way it is done in the core `EntityAccessControlHandler`.

pfrenssen:
Fixes #691 

Very happy with how this turned out.

It is remarkable how this small change cleans up our access code. By leveraging the core functionality we no longer need to handle the access checks and merging of cacheability metadata ourselves and the end result is very lean and readable code.

This is building on top of #684 so marking this as a draft until that PR is in.

### DIFF
--- a/src/Event/AccessEventBase.php
+++ b/src/Event/AccessEventBase.php
@@ -57,15 +57,17 @@ class AccessEventBase extends Event implements AccessEventInterface {
   /**
    * {@inheritdoc}
    */
-  public function grantAccess(): void {
+  public function grantAccess(): AccessResultInterface {
     $this->access = $this->access->orIf(AccessResult::allowed());
+    return $this->access;
   }
 
   /**
    * {@inheritdoc}
    */
-  public function denyAccess(): void {
+  public function denyAccess(): AccessResultInterface {
     $this->access = $this->access->orIf(AccessResult::forbidden());
+    return $this->access;
   }
 
   /**
--- a/src/Event/AccessEventBase.php
+++ b/src/Event/AccessEventBase.php
@@ -73,6 +73,14 @@ class AccessEventBase extends Event implements AccessEventInterface {
   /**
    * {@inheritdoc}
    */
+  public function mergeAccessResult(AccessResultInterface $access_result): AccessResultInterface {
+    $this->access = $this->access->orIf($access_result);
+    return $this->access;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function getGroup(): ContentEntityInterface {
     return $this->group;
   }
--- a/src/Event/AccessEventInterface.php
+++ b/src/Event/AccessEventInterface.php
@@ -19,8 +19,11 @@ interface AccessEventInterface extends RefinableCacheableDependencyInterface {
    *
    * Calling this method will cause access to be granted for the action that is
    * being checked, unless another event listener denies access.
+   *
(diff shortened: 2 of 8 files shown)


claudiu-cristea (commented):
Opening the access check to 3rd party is a good decision. However, the proposal here aligns with the _entity access check_, where the a result of _neutral_ reads as "allow access". I'm just curious why you chose entity over route access. Drupal core handles route access different than entity access. On routes, only an _allowed_ result reads as "allow access", _neutral_ means deny.

Relevant:

* https://www.drupal.org/docs/8/api/routing-system/access-checking-on-routes/route-access-checking-basics#s-multiple-access-checks
* \Drupal\Core\Access\AccessManager

pfrenssen:
The reason for this is that we are indeed doing entity access checks. This code can also be called for route access but then the caller will need to translate a neutral result to forbidden. I haven't given much thought yet to route access. I see that indeed in `AccessManager::checkRequest()` the `andIf()` access merging strategy is used, while `EntityAccessControlHandler` uses `orIf()`.

The code path that is being followed to reach these methods is: `$entity->access($operation)` -> `EntityAccessControlHandler::check()` -> `og_entity_access()` hook implementation -> `OgAccess::userAccessEntityOperation()`

So I am confident that this is using the right approach for this use case, but I need to research how route access checking relates to entity access checking. From a first glance they seem to be completely independent implementations. We might need to support this also in OG.

pfrenssen:
Assigning to me, I will update the documentation for the affected methods to make it clear which access result will be returned under which conditions.

MPParsley:
Glad to see this cleaned up some of the access code.

Just one question here: I know we're in alpha but changing the default from deny to neutral might break for some users, shouldn't we release this in a new major version (or document this)?

amitaibu:
IMO A new alpha would be the right move.

pfrenssen:
We are still in alpha and as such it is OK to make B/C breaking changes. We are also very clear about this, it is mentioned on the project page, and in the sticky post with the 8.0 roadmap.

claudiu-cristea (approved):
Nice work. I've anticipated somehow the reply to my question https://github.com/Gizra/og/pull/692#pullrequestreview-466051992, but I wanted to make sure we're doing right.

pfrenssen:
Thanks! Let's finish up this track so we can start preparing the next release.

PR #695, opened 2020-08-14
694 Fix access condition check in SubscriptionController

### LINKED ISSUE #694: Users can still access 'subscribe' link directly for closed groups
## Overview
When a group is closed, a user can still go to the 'subscribe' link for that group and still register to that group. This is a bug and needs fixing.

## Details
This happens on a site which uses `hook_og_user_access_alter` to deny permissions for `subscribe` and `subscribe without approval` for the group in question.

varshith:
#694 

## Technical Details
The issue here is that in `SubscriptionController` there is an access check where the return value of `$this->ogAccess->userAccess()` is checked directly. This will always return `true` even for Forbidden access.
We should use `userAccess()->isAllowed()` or 'userAccess()->isForbidden()' inside the if conditions.
This PR fixes this issue. Tested locally for my case and it works as expected after this fix.

### DIFF
--- a/src/Controller/SubscriptionController.php
+++ b/src/Controller/SubscriptionController.php
@@ -136,7 +136,8 @@ class SubscriptionController extends ControllerBase {
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    if (!$this->ogAccess->userAccess($group, 'subscribe', $user) && !$this->ogAccess->userAccess($group, 'subscribe without approval', $user)) {
+    if ((($access = $this->ogAccess->userAccess($group, 'subscribe', $user)) && $access->isForbidden()) &&
+      (($access = $this->ogAccess->userAccess($group, 'subscribe without approval', $user)) && $access->isForbidden())) {
       throw new AccessDeniedHttpException();
     }
 
--- a/src/Controller/SubscriptionController.php
+++ b/src/Controller/SubscriptionController.php
@@ -86,7 +86,7 @@ class SubscriptionController extends ControllerBase {
       throw new AccessDeniedHttpException();
     }
 
-    $user = $this->currentUser();
+    $user = $this->entityTypeManager()->getStorage('user')->load($this->currentUser()->id());
 
     if ($user->isAnonymous()) {
       // Anonymous user can't request membership.
--- a/src/Controller/SubscriptionController.php
+++ b/src/Controller/SubscriptionController.php
@@ -137,7 +137,7 @@ class SubscriptionController extends ControllerBase {
     }
 
     $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
-    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval');
+    $subscribe_without_approval = $this->ogAccess->userAccess($group, 'subscribe without approval');
     if ($subscribe->isForbidden() && $subscribeWithoutApproval->isForbidden()) {
       throw new AccessDeniedHttpException();
     }
--- a/src/Controller/SubscriptionController.php
+++ b/src/Controller/SubscriptionController.php
@@ -138,7 +138,7 @@ class SubscriptionController extends ControllerBase {
 
     $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
     $subscribe_without_approval = $this->ogAccess->userAccess($group, 'subscribe without approval');
-    if ($subscribe->isForbidden() && $subscribe_without_approval->isForbidden()) {
+    if (!$subscribe->isAllowed() && !$subscribe_without_approval->isAllowed()) {
       throw new AccessDeniedHttpException();
     }
 
--- a/src/Controller/SubscriptionController.php
+++ b/src/Controller/SubscriptionController.php
@@ -136,8 +136,8 @@ class SubscriptionController extends ControllerBase {
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    $subscribe = $this->ogAccess->userAccess($group, 'subscribe', $user);
-    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval', $user);
+    $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
+    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval');
     if ($subscribe->isForbidden() && $subscribeWithoutApproval->isForbidden()) {
       throw new AccessDeniedHttpException();
     }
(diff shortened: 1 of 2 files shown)


src/Controller/SubscriptionController.php
```diff
@@ -136,7 +136,8 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    if (!$this->ogAccess->userAccess($group, 'subscribe', $user) && !$this->ogAccess->userAccess($group, 'subscribe without approval', $user)) {
+    if ((($access = $this->ogAccess->userAccess($group, 'subscribe', $user)) && $access->isForbidden()) &&
+      (($access = $this->ogAccess->userAccess($group, 'subscribe without approval', $user)) && $access->isForbidden())) {
```
MPParsley: Making this a bit more readable:
```suggestion
    $subscribe = $this->ogAccess->userAccess($group, 'subscribe', $user);
    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval', $user);
    if ($subscribe->isForbidden() && $subscribeWithoutApproval->isForbidden()) {
```

MPParsley (changes requested):

varshith:
@MPParsley 
done.

src/Controller/SubscriptionController.php
```diff
@@ -136,7 +136,9 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    if (!$this->ogAccess->userAccess($group, 'subscribe', $user) && !$this->ogAccess->userAccess($group, 'subscribe without approval', $user)) {
+    $subscribe = $this->ogAccess->userAccess($group, 'subscribe', $user);
+    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval', $user);
```
MPParsley: We don't really need to pass the user if it's the current user either:
```suggestion
    $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval');
```

src/Controller/SubscriptionController.php
```diff
@@ -136,7 +136,9 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    if (!$this->ogAccess->userAccess($group, 'subscribe', $user) && !$this->ogAccess->userAccess($group, 'subscribe without approval', $user)) {
+    $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
+    $subscribeWithoutApproval = $this->ogAccess->userAccess($group, 'subscribe without approval');
```
amitaibu:
```suggestion
    $subscribe_without_approval = $this->ogAccess->userAccess($group, 'subscribe without approval');
```

amitaibu:
@varshith when you'll have time, it would be great to have a test for this as-well

src/Controller/SubscriptionController.php
```diff
@@ -136,7 +136,9 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       return new RedirectResponse($group->toUrl()->setAbsolute(TRUE)->toString());
     }
 
-    if (!$this->ogAccess->userAccess($group, 'subscribe', $user) && !$this->ogAccess->userAccess($group, 'subscribe without approval', $user)) {
+    $subscribe = $this->ogAccess->userAccess($group, 'subscribe');
+    $subscribe_without_approval = $this->ogAccess->userAccess($group, 'subscribe without approval');
+    if ($subscribe->isForbidden() && $subscribeWithoutApproval->isForbidden()) {
```
MPParsley:
```suggestion
    if ($subscribe->isForbidden() && $subscribe_without_approval->isForbidden()) {
```

MPParsley:
@amitaibu, fixed the snake case.

MPParsley (approved):

src/Controller/SubscriptionController.php
MPParsley: @amitaibu, changed the variable to snake case in the if as well. fyi, you can now also give code suggestions on multiple lines at once by clicking and dragging over the lines you want.

src/Controller/SubscriptionController.php
```diff
@@ -86,7 +86,7 @@ public function subscribe($entity_type_id, EntityInterface $group, OgMembershipT
       throw new AccessDeniedHttpException();
     }
 
-    $user = $this->entityTypeManager()->getStorage('user')->load($this->currentUser()->id());
+    $user = $this->currentUser();
```
MPParsley: Exception: TypeError: Argument 2 passed to Drupal\og\MembershipManager::createMembership() must implement interface Drupal\user\UserInterface, instance of Drupal\Core\Session\AccountProxy given, called in /home/travis/build/Gizra/og/src/Og.php on line 190

Seems like this was the reason the user was loaded as such:

```suggestion
    $user = $this->entityTypeManager()->getStorage('user')->load($this->currentUser()->id());
```

pfrenssen:
It would be better to check for `::isAllowed()` rather than for `::isForbidden()` since this would open up a security issue in case a neutral result is returned. If the access result is neutral then the user shouldn't have access either. Only when either of them returns allowed.

pfrenssen (changes requested):
We need to check for `::isAllowed()` rather than for `::isForbidden()`.

pfrenssen:
The previous implementation is not type safe (it is casting an `AccessResult` object to a boolean) so it should have triggered a test failure when we turned on strict typing. This didn't happen so it appears we have no test coverage for this, and this is a security issue since it allows users to join groups without permission.

It would be a good idea to add a test for this so that we can prevent this security issue regressing in the future. We should also be more careful in the future not to accept feature requests without sufficient coverage.

pfrenssen:
Assigning to write a test.

pfrenssen:
I have pushed the test in a separate PR so that we can check that it fails correctly: #698. This can be merged into this PR once the test results are in.

pfrenssen:
Test is failing correctly, so I merged it in here. Now hopefully the test should turn green.

pfrenssen:
Test is green! I have also changed the check to `::isAllowed()` as I suggested earlier. Unassigning.

This is good to go for me, but I won't merge right now since I added the test.

@varshith thanks a lot for the bug report and the fix!

pfrenssen (approved):

MPParsley (approved):

PR #700, opened 2020-09-10
Fix PHP CS issues

MPParsley:
Seems like there's been an update to phpcs;

```
PHP CodeSniffer Config installed_paths set to ../../drupal/coder/coder_sniffer,../../sirbrillig/phpcs-variable-analysis,../../slevomat/coding-standard
........EEE..................E..E.......E....E.....E.E......  60 / 201 (30%)
......EE.E....E.....E...E.......EW.......................... 120 / 201 (60%)
...............................W.EEWW.............E......E.. 180 / 201 (90%)
E.........W.........W                                        201 / 201 (100%)
FILE: /home/travis/build/Gizra/og/tests/src/Functional/GroupTabTest.php
--------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 68 | ERROR | The array declaration line has 89 characters (the limit is 80).
    |       | The array content should be split up over multiple lines
 71 | ERROR | The array declaration line has 89 characters (the limit is 80).
    |       | The array content should be split up over multiple lines
--------------------------------------------------------------------------------
FILE: ...travis/build/Gizra/og/tests/src/Functional/GroupSubscribeFormatterTest.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 69 | ERROR | The array declaration line has 96 characters (the limit is 80).
    |       | The array content should be split up over multiple lines
--------------------------------------------------------------------------------
FILE: /home/travis/build/Gizra/og/tests/src/Unit/GroupTypeManagerTest.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 229 | ERROR | The array declaration line has 95 characters (the limit is 80).
     |       | The array content should be split up over multiple lines
--------------------------------------------------------------------------------
FILE: ...ome/travis/build/Gizra/og/tests/src/Unit/OgResolvedGroupCollectionTest.php
--------------------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
--------------------------------------------------------------------------------
  33 | ERROR | The array declaration line has 89 characters (the limit is 80).
     |       | The array content should be split up over multiple lines
  80 | ERROR | The array declaration line has 83 characters (the limit is 80).
     |       | The array content should be split up over multiple lines
  95 | ERROR | The array declaration line has 91 characters (the limit is 80).
[...]

PR #701, opened 2020-09-22
Add 'auto_add_group_owner_membership' setting

bboro:
Case study:

Sometimes website may be creating entities programmatically, or just plain do not desire automatic membership creation for the group creator. In such cases, it's quite difficult to prevent this from happening due to the creation being in `og_entity_insert`.

Adding new global setting: 'Automatically add creators to the group' 

image

* When the setting is ON (default) membership is created
  automatically for group owners/creators.
* When the setting is OFF membership is not created automatically
  for group owners/creators.

Changelist:

* Added config schema changes
* Added default installation config changes
* Added update hook to update existing configurations
* Added the setting in the AdminSettingsForm
* Updated og_entity_insert to be aware of the new config setting
* Updated tests


bboro:
This requirement was originally stemming from https://github.com/Gizra/jep-portal/issues/663 where we don't want to create memberships automatically for group owners.

og_ui/src/Form/AdminSettingsForm.php
```diff
@@ -133,6 +132,13 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       }
     }
 
+    $form['auto_add_owners'] = [
```
amitaibu: How about a more explicit name, like

```suggestion
    $form['auto_add_membership_to_group_owner'] = [
```

bboro: Renamed to `auto_add_group_owner_membership`

bboro:
Looks like some PHPCS failures, but looks like fixes are coming in another PR. I had a quick look at my test failures and seems like it's nothing introduced from my side, but I may have easily missed it.

pfrenssen:
I did a code review and everything looks good to me. Thanks for adding test coverage!

I did not test it out in the UI though, if someone can confirm this works as expected when setting the option in the UI this is good to go.

amitaibu:
I've confirmed manually the feature works

## Enabled

image


## Disabled

image


I think as a follow up we should show "No members in the group yet" message, as we've never had up until now empty groups. @bboro care to tackle this?

pfrenssen (approved):
Thanks for the test and the screenshots!

Approving, but we will have to wait for #700 to be in before merging this. We're affected by a test failure which is being fixed in that PR.

bboro:
@pfrenssen thanks for the review!

@amitaibu sure thing I'll make some time to work on this tomorrow. Question about the implementation: Since the members listing is a view, how should we handle the view configuration on existing installations? Do we just leave them be since they may have a custom empty message set already?

amitaibu:
> how should we handle the view configuration on existing installations?

You can change the existing one. As it was not (easily) possibly to have no-members, I don't think it's a problem.

og.install
```diff
@@ -93,3 +94,48 @@ function og_update_8002() {
...
+    'table' => 'views',
+    'field' => 'area',
+    'relationship' => 'none',
+    'group_type' => 'group',
+    'admin_label' => 'No results text',
+    'empty' => TRUE,
+    'tokenize' => FALSE,
+    'content' => [
+      'value' => 'No members in the group yet.',
+      'format' => 'basic_html',
```
bboro: Is it ok to assume basic_html is an existing format here? (Also same in the optional installation view config) 

If not, does anyone have any pointers on how this is usually handled? Perhaps I should use a different plugin/field?

amitaibu: I don't think we need an upgrade path for Views config.

bboro: > I don't think we need an upgrade path for Views config.

@amitaibu so we don't want to update existing installations and won't need this update hook? I'm not sure I understand exactly what you mean.

amitaibu: Correct, no upgrade hook for Views

bboro: Thanks! I'll remove

bboro: FYI I'm also setting the empty text to unfiltered text in the installationg config yml instead of trying to force text formats

MPParsley:
#700 is fixed, updating branch.

og.install
```diff
@@ -9,6 +9,7 @@ declare(strict_types = 1);
 
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\views\Views;
```
MPParsley: Unused use statement:

```suggestion
```

tests/src/Kernel/GroupManagerSubscriptionTest.php
```diff
@@ -117,6 +117,26 @@ public function testGroupManagerSubscription($group_has_owner, $membership_is_ov
...
+    $config->set('auto_add_group_owner_membership', $auto_add_group_owner_membership);
+    $config->save();
+    $group = Node::create([
+      'title' => 'membership is not overridden',
+      'type' => 'group',
+      'uid' => $this->owner->id(),
+    ]);
+    $group->save();
+    $membership = $this->membershipManager->getMembership($group, $this->owner->id());
+    // Membership should be created only when auto_add_group_owner_membership is enabled.
```
MPParsley: Line exceeds 80 characters; contains 89 characters 

```suggestion
    // Membership should be created only when auto_add_group_owner_membership is
    // enabled.
```

MPParsley:
Fixed CS issues:

```
FILE: /home/travis/build/Gizra/og/og.install
----------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 12 | WARNING | [x] Unused use statement
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
FILE: ...me/travis/build/Gizra/og/tests/src/Kernel/GroupManagerSubscriptionTest.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------
 136 | WARNING | Line exceeds 80 characters; contains 89 characters
--------------------------------------------------------------------------------
```

tests/src/Kernel/GroupManagerSubscriptionTest.php
```diff
@@ -176,4 +197,24 @@ public static function groupManagerSubscriptionProvider() {
...
+   * Data provider for ::testAutoAddOwnersSetting().
+   *
+   * @return array[]
+   *   Array of data with values for the 'auto_add_group_owner_membership'
+   *   config key.
+   */
+  public static function autoAddOwnersSettingProvider() {
+    return [
+      [
+        // Auto Add owners.
```
MPParsley:
```suggestion
        // Auto add owners.
```

config/optional/views.view.og_members_overview.yml
```diff
@@ -430,7 +428,18 @@ display:
...
+        area_text_custom:
+          id: area_text_custom
+          table: views
+          field: area_text_custom
+          relationship: none
+          group_type: group
+          admin_label: 'No results text'
+          empty: true
+          tokenize: false
+          content: 'No members in the group yet.'
```
MPParsley:
```suggestion
          content: 'There are no members in this group.'
```

```suggestion
          content: 'There are no members that belong to this group.'
```

MPParsley (commented):
Test are green, just added a small grammar suggestion.

Should we make the default role configurable?

bboro:
Rerolled against latest 8.x-1.x

MPParsley (approved):

PR #703, opened 2020-09-26
Fix up formatting of the Readme.md.

gheydon:
Fix up formatting of the sub numbering in section 9 of the Installation.

The Markdown had not been formatted correctly which meant when it gets rendered it would just displays it as a long paragraph and is hard to read.


pfrenssen (approved):

pfrenssen:
Nice catch, thanks!

This is failing because of an unrelated issue (#700), but since this is a documentation fix this is OK to merge without requiring a passing test suite.

PR #704, opened 2020-10-16
Remove any null values when building the list of target groups.

gheydon:
During the building of the OgComplex widget (actually with any widget) a null item is appended to the og_audience field. This then causes a undefined index warning.

This fixes the warning and then strips out the NULL target ids.

### DIFF
--- a/src/MembershipManager.php
+++ b/src/MembershipManager.php
@@ -306,7 +306,7 @@ class MembershipManager implements MembershipManagerInterface {
 
       // Compile a list of group target IDs.
       $target_ids = array_filter(array_map(function ($value) {
-        return isset($value['target_id']) ? $value['target_id'] : NULL;
+        return $value['target_id'] ?? NULL;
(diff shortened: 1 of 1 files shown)


src/MembershipManager.php
```diff
@@ -305,9 +305,9 @@ public function getGroupIds(EntityInterface $entity, $group_type_id = NULL, $gro
       }
 
       // Compile a list of group target IDs.
-      $target_ids = array_map(function ($value) {
-        return $value['target_id'];
-      }, $entity->get($field->getName())->getValue());
+      $target_ids = array_filter(array_map(function ($value) {
+        return isset($value['target_id']) ? $value['target_id'] : NULL;
```
MPParsley: See https://www.php.net/manual/en/migration70.new-features.php
```suggestion
        return $value['target_id'] ?? NULL;
```

MPParsley (approved):

PR #711, opened 2020-12-19
Downgrade composer to version 1 for Drupal 8.8

MPParsley:
Travis is now using composer 2 which requires some dependencies to be updated.
```
$ composer self-update
Updating to version 2.0.8 (stable channel).
   Downloading (100%)
```

```
You are using Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report a plugin-issue to ask them to support Composer 2.
Script @composer update phpunit/phpunit symfony/phpunit-bridge phpspec/prophecy symfony/yaml --with-dependencies --no-progress handling the drupal-phpunit-upgrade event returned with error code 2
```

PHP Fatal error:  Trait 'Prophecy\PhpUnit\ProphecyTrait' not found in /home/travis/build/Gizra/drupal/core/tests/Drupal/TestTools/PhpUnitCompatibility/PhpUnit9/TestCompatibilityTrait.php on line 12

See https://www.drupal.org/node/3176567

We should probably drop support for Drupal 8.8.x, it doesn't support composer 2.

```
    - drupal/core-vendor-hardening 8.8.x-dev requires composer-plugin-api ^1.1 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,8 @@ before_script:
 
   # Update PHPUnit.
   - test ${TEST_SUITE} == "PHP_CodeSniffer" || travis_retry composer update phpunit/phpunit symfony/phpunit-bridge phpspec/prophecy symfony/yaml --with-dependencies --working-dir=$DRUPAL_DIR
-
+  - test ${TEST_SUITE} == "PHP_CodeSniffer" || travis_retry composer drupal-phpunit-upgrade
+  
   # Start a web server on port 8888 in the background.
   - test ${TEST_SUITE} == "PHP_CodeSniffer" || nohup php -S localhost:8888 --docroot $DRUPAL_DIR > /dev/null 2>&1 &
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,9 @@ before_script:
   # Make sure Composer is up to date.
   - composer self-update
 
+  # Downgrade to composer version 1 to support Drupal 8.8.
+  - test ${TEST_SUITE} == "8.8.x" && composer self-update --1
+
   # Remember the current directory for later use in the Drupal installation.
   - MODULE_DIR=$(pwd)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ before_script:
   - composer self-update
 
   # Downgrade to composer version 1 to support Drupal 8.8.
-  - test ${TEST_SUITE} == "8.8.x" && composer self-update --1
+  - test ${TEST_SUITE} != "8.8.x" || composer self-update --1
 
   # Remember the current directory for later use in the Drupal installation.
   - MODULE_DIR=$(pwd)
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -17,6 +17,7 @@
     "require-dev": {
         "drupal/coder": "^8.2",
         "phpunit/phpunit": "^7 || ^8",
+        "phpspec/prophecy-phpunit: ^2",
```
MPParsley:
```suggestion
        "phpspec/prophecy-phpunit": "^2",
```

composer.json
```diff
@@ -17,6 +17,7 @@
     "require-dev": {
         "drupal/coder": "^8.2",
         "phpunit/phpunit": "^7 || ^8",
+        "phpspec/prophecy-phpunit": "^2",
```
MPParsley:
```suggestion
```

MPParsley:
@pfrenssen, @amitaibu, ok to drop support for Drupal 8.8?

amitaibu:
@MPParsley why is that needed? I'm not even sure what's the diff between 8.8 and 8.9.

MPParsley:
composer 2 requires composer/installers ^1.9 but composer/installers is locked to version v1.7.0

This was fixed for Drupal 8.9 but as far as I can tell not for 8.8. See https://www.drupal.org/project/drupal/issues/3159730

amitaibu:
Got it, thanks. IMO we should keep support for 8.8 as it's still an active, and has a fairly new release.

I wonder if another way to fix it could be to downgrade composer to 1.x?

MPParsley:
We could probably find a patch for it somewhere, similar to https://www.drupal.org/files/issues/2020-07-17/3159730-2.patch
Or we could downgrade composer to v1: `composer self-update --1` in travis.

```
  # Downgrade to composer version 1 to support Drupal 8.8.
  - test ${TEST_SUITE} != "8.8.x" || composer self-update --1
```

.travis.yml
```diff
@@ -50,6 +50,9 @@ before_script:
   # Make sure Composer is up to date.
   - composer self-update
 
+  # Downgrade to composer version 1 to support Drupal 8.8.
+  - test ${TEST_SUITE} == "8.8.x" && composer self-update --1
```
MPParsley:
```suggestion
  - test ${TEST_SUITE} != "8.8.x" || composer self-update --1
```

.travis.yml
```diff
@@ -50,6 +50,9 @@ before_script:
   # Make sure Composer is up to date.
   - composer self-update
 
+  # Downgrade to composer version 1 to support Drupal 8.8.
+  - test ${TEST_SUITE} != "8.8.x" || composer self-update --1
```
MPParsley, line 54: This may be more readable:
```suggestion
  - if [ ${TEST_SUITE} == "8.8.x" ]; then composer self-update --1; fi
```

MPParsley:
@pfrenssen, @amitaibu, all tests are green but we still have soms phpcs issues from #700.

Merge this one first?

amitaibu (approved):
lgtm

PR #713, opened 2021-01-18
Use dependency injection for loadMultiple

MPParsley:
FieldStorageConfig::loadMultiple calls should be avoided in classes, use dependency injection instead.


MPParsley:
Remaining php cs issues to be fixed in #700.

MPParsley:
Fixed in #700.

Issue #160, opened 2016-09-04
Remove the "OG complex" widget in favor of a single input


amitaibu:
@RoySegall had an idea on how to simplify OG even more - by no longer splitting the "My groups"/"Other groups". 

Instead - show a single input field (e.g. autocomplete or select list), and have all the groups there. This will also simplify our ER selection handler - we'll just need to show all groups for admin user, or "my groups" for non-admin user.

I'm really starting to like this idea 😄

RoySegall:
@amitaibu Should I start a POC for this one?

amitaibu:
No, would be great if you could tackle the other issues first

aDarkling:
I would love that as well!
That would (should?) allow Inline Entity widgets to be used to add or edit groups on the fly.
And it may also allow entity reference views to be used - which would also greatly enhance UX.

damienmckenna:
I created a new PR in my own fork for this change which works against this repo's 8.x-1.x branch: https://github.com/damienmckenna/og/pull/1

MPParsley:
Closing in favor of #570.

PR #726, opened 2021-09-28
#160 Remove complex widget

MPParsley:


tests/src/Kernel/EntityReference/OgSelectionConfigurationFormTest.php
```diff
@@ -0,0 +1,98 @@
...
+class OgSelectionConfigurationFormTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static $modules = [
+    'field',
+    'field_ui',
+    'entity_test',
+    'og',
```
MPParsley, line 29:
```suggestion
    'og',
    'options',
```

damienmckenna:
Is this intended to be a cleaned-up copy of the original PR? I've lost track of what's what.

MPParsley:
@damienmckenna, indeed. I had closed #570 and fixed some merge conflicts and a broken test in this PR.

Perhaps you can have a look at the deprecation error?

src/Plugin/EntityReferenceSelection/OgSelection.php
```diff
@@ -81,37 +155,45 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
...
+    $entity = $this->configuration['entity'];
+    $bundle = $entity->bundle();
 
     $ids = [];
-    if (!empty($this->getConfiguration()['field_mode']) && $this->getConfiguration()['field_mode'] === 'admin') {
-      // Don't include the groups, the user doesn't have create permission.
-      foreach ($user_groups as $group) {
+    foreach ($this->getUserGroups() as $group) {
+      // Check user has "create" permission on this entity.
+      $permission = "create $bundle content";
```
zero2one, line 184: This should be:
```php
$permission = sprintf('create %s %s', $bundle, $entity->getEntityTypeId());
```

zero2one (changes requested):

src/Plugin/EntityReferenceSelection/OgSelection.php
MPParsley, line 184: @zero2one, we should use the permission layer instead of using string patterns:
```suggestion
      // Check if the current user can perform the "create" operation on this entity.
      if ($this->ogAccess->userAccessGroupContentEntityOperation('create', $group, $entity, $this->currentUser)) {
```

zero2one: Then we need to check if we have a new or existing entity ($entity->isNew()) and do something like this:

* Entity is new => check "create" operation for that group.
* Entity is existing and linked to the group being checked: check "update" operation for that group.
* Entity is existing and not linked to the group being checked : check "create" operation for that group.

pfrenssen:
I started reviewing this, but right at the start of the code review I see the same bizarre and incorrect logic that I already reported back in 2019 on #570.

Can we try not to continue on this flawed approach? By continuing on this path and not addressing the fundamental flaws in it we are just wasting time. Maybe if the approach is generally considered to be correct and there are only some rough edges that I happen to stumble over, would it be possible to at least remove all code changes that are either undocumented, untested or have any other bizarre intentions / code smells?

og.module
```diff
@@ -164,6 +164,20 @@ function og_entity_access(EntityInterface $entity, $operation, AccountInterface
...
+      /** @var \Drupal\Core\Field\FieldDefinitionInterface $field_definition */
+      if (!in_array($field_definition->getName(), array_keys($audience_fields))) {
+        continue;
+      }
+
+      // Check if the field needs to be required.
+      $field_definition->setRequired($access);
+    }
+  }
+
```
pfrenssen, line 175: This is logic is undocumented, untested and seemingly unneeded. Altering field definitions to set them required inside `hook_entity_access()`? This is completely bizarre and can not be the correct solution for whatever this is trying to fix.

This section MUST be removed.

og.module
```diff
@@ -201,6 +215,11 @@ function og_og_membership_type_access(OgMembershipTypeInterface $entity, $operat
 function og_entity_create_access(AccountInterface $account, array $context, $bundle) {
   $entity_type_id = $context['entity_type_id'];
 
+  if (!empty($context['skip_og_permission'])) {
```
pfrenssen, line 213: This workaround `skip_og_permission` indicates a flaw in the logic of this PR. We should not introduce a bypass in access checking. This will just lead to more complexity down the road and potential security vulnerabilities. There must be a loop introduced somewhere, where an access check is done inside of an access check. We should fix this in a different way.

og.module
```diff
@@ -226,23 +245,41 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
...
     }
 
+    if (count($audience_fields) === 1) {
+      // Set the field to required only if we have a single audience field.
+      // If we have multiple fields then the require logic will perform at the
+      // entity constraint level.
+      $field_definition->setRequired(!\Drupal::entityTypeManager()
+        ->getAccessControlHandler($entity_type_id)
+        ->createAccess($bundle, $account, ['skip_og_permission' => TRUE]));
+    }
```
pfrenssen, line 260: What is going on here? Again altering field definitions inside access hooks???

This change MUST be removed. This is NOT the place for messing with anything else that is not strictly about access.

og.module
```diff
@@ -226,23 +245,41 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
...
       'target_type' => $field_definition->getFieldStorageDefinition()->getSetting('target_type'),
       'handler' => $field_definition->getSetting('handler'),
-      'field_mode' => 'admin',
     ];
     /** @var \Drupal\Core\Entity\EntityReferenceSelection\SelectionPluginManager $handler */
     $handler = \Drupal::service('plugin.manager.entity_reference_selection');
+    $handler = $handler->getInstance($options);
+
+    // @todo Remove deprecated call to Og::getSelectionHandler.
+    $handler = Og::getSelectionHandler($field_definition);
```
pfrenssen, line 271: Introducing deprecated calls in new code?

og.module
```diff
@@ -226,23 +245,41 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
...
 
-    if ($handler->getInstance($options)) {
-      return AccessResult::neutral();
+    if ($handler && $handler->countReferenceableEntities()) {
+      // At least one of the fields has referenceable groups. That mean that
+      // the user can create group content when referencing the group content in
+      // at least one of the fields.
+      return AccessResult::allowed();
```
pfrenssen, line 277: This is not clear to me, but it raises a red flag. An additional check is done, but why is the access level raised from neutral to allowed?

pfrenssen (changes requested):
I have reviewed 10% of the PR. There are some serious problems with what I have seen so far.

og.module
```diff
@@ -226,23 +245,41 @@ function og_entity_create_access(AccountInterface $account, array $context, $bun
   // @see \Drupal\og\Plugin\EntityReferenceSelection\OgSelection::buildEntityQuery()
   $required = FALSE;
 
+  /** @var \Drupal\og\OgGroupAudienceHelperInterface $audience_helper */
+  $audience_helper = \Drupal::service('og.group_audience_helper');
+  $audience_fields = $audience_helper->getAllGroupAudienceFields($entity_type_id, $bundle);
   $field_definitions = \Drupal::service('entity_field.manager')->getFieldDefinitions($entity_type_id, $bundle);
```
pfrenssen, line 246: Both these calls to `::getAllGroupAudienceField()` and `::getFieldDefinitions()` duplicate the same actions. The field definition list is retrieved twice and filtered twice now.

amitaibu:
@pfrenssen after so many years that this PR is laying around, it's possible that what's best is to close it, and start from scratch (if you have the will/ capacity for it :wink: ).

tests/src/Functional/OgMultipleSelectionWidgetAutoCompleteTest.php
```diff
@@ -0,0 +1,199 @@
...
+
+    // Create a role.
+    $role_storage = $this->container->get('entity_type.manager')->getStorage('user_role');
+    $role_storage->create([
+      'id' => 'dummy_role',
+      'permissions' => [
+        'create group_content content',
+        'edit own group_content content',
+        'edit any group_content content',
+      ],
```
MPParsley, line 118: Drupal\Tests\og\Functional\OgMultipleSelectionWidgetAutoCompleteTest::testAutoCompleteForNonGroupMember
Role label is expected to be a string. in /home/travis/build/Gizra/drupal/core/modules/user/user.module:851

```suggestion
      ],
      'label' => 'Dummy role',
```

claudiu-cristea:

PR #731, opened 2021-12-15
Revert the fix from #709

pfrenssen:
In local testing I have discovered that the fix that was proposed in #709 does not seem to be needed. The test for that PR still passes if the fix is reverted.

It might be possible I am getting a false negative due to some environment specific issues (for example, I have only tested on a Drupal 8 installation). Also it is possible that the test for the issue was not capturing the exact steps needed to replicate the bug.

Let's see what the test bot says.


pfrenssen:
OK I found out why this is passing on my installation. This is actually a duplicate of a core bug: [Issue #2730631: Upcast node and node_revision parameters of node revision routes](https://www.drupal.org/project/drupal/issues/2730631). I had this patch applied.

In the meantime this has been fixed in core and released in Drupal 9.3.

Shall we remove the fix from OG since this is not the right place to fix it? If users are affected they can update to Drupal 9.3, or on older versions they can apply the patch from the core issue.

amitaibu:
> Shall we remove the fix from OG since this is not the right place to fix it?

Yes, sounds right

idimopoulos (approved):
RTBC +1.

PR #732, opened 2021-12-17
Support Drupal 9.3

pfrenssen:
Drupal 9.3 has been released, and versions 9.0 and 9.1 are no longer supported. Let's update our composer.json and test matrix.

I've also added experimental support for PHP 8, but if this causes trouble we can handle it in a separate PR.


pfrenssen:
This is failing on Drupal 9.3 because the fix that was added in #709 was in fact caused by a bug in Drupal core (ref. [Issue #2730631: Upcast node and node_revision parameters of node revision routes](https://www.drupal.org/project/drupal/issues/2730631)). It seems that the fix in core and the one in OG are causing some kind of conflict.

Let's block this until #731 is in, which reverts the fix from #709.

idimopoulos (approved):
RTBC +1 :)

pfrenssen:
Thanks for reviewing!

PR #733, opened 2021-12-17
Signature of EventDispatcherInterface::dispatch() has changed

pfrenssen:
In Drupal 9.1 the order of the arguments to this method has been changed.

Ref. [Change record #3154407: Signature of EventDispatcherInterface::dispatch() has changed](https://www.drupal.org/node/3154407)

Fixes #725.

This also updates the touched classes with modern practices, such as introducing missing type hints, and some small unrelated fixes.

PR #734, opened 2021-12-18
Refactor \Drupal\og_ui\BundleFormAlter

pfrenssen:
The `\Drupal\og_ui\BundleFormAlter` class is in need of refactoring and bug fixing.

In #733 new scalar type hints are being introduced, and the testing of this has uncovered that the `BundleFormAlter` class is calling into `Og::isGroup()` with the `$bundle_id` set to `NULL`, while this method is only accepting strings.

Apart from this bug the class is also showing its age and is not up to date with current best practices. Let's:

* Look into the architecture of the class. Now it is using a very workaround way of instantiating it with an entity with is derived from the form state, then later on the form and form state are passed and a protected `::prepare()` method then acts as a second constructor which populates further properties. Internally then the form and form state are still passed on between methods. It looks like this can be simplified a lot, by just passing on the form and form state and using some basic helper methods. It looks like the whole class can operate without keeping any state, with purely static methods.
* Introduce strict typing and type hints wherever possible.
* Rename variables so that their purpose is clear and non-ambiguous.
* Update tests, especially the AJAX functionality is not being tested in a proper way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ mysql:
 services:
   - mysql
 
+addons:
+  chrome: stable
+
 cache:
   directories:
     - $HOME/.composer/cache/files
@@ -77,5 +80,8 @@ before_script:
 
   # Export database variable for kernel tests.
   - export SIMPLETEST_DB=mysql://root:@127.0.0.1/og
+
+  # Start chromedriver for JS tests.
+  - chromedriver --port=4444 &
 
 script: DRUPAL_DIR=$DRUPAL_DIR MODULE_DIR=$MODULE_DIR $MODULE_DIR/scripts/travis-ci/run-test.sh $TEST_SUITE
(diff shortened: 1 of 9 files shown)


pfrenssen:
After spending some quality time with the class I am realizing it would better be converted into a service, since it depends on a bunch of other services such as the `GroupTypeManager`.

pfrenssen:
This is ready but blocked by #732. Assigning to me to keep track.

pfrenssen:
Now that #732 has been merged, this is unblocked and ready for review. Unassigning.

amitaibu:
@bboro if you can re-roll as-well please

bboro (approved):

amitaibu:
@bboro Thanks. I see Travis is failing - let's fix that first?

bboro:
@amitaibu did phpcs ever work on this project?

amitaibu:
Yes - for example [this PR](https://github.com/Gizra/og/pull/763)

PR #738, opened 2022-02-23
Avoid PHP 8.1 deprecation notices

claudiu-cristea:
## Description

With PHP 8.1, we get this message:

> Return type of Drupal\og\Event\PermissionEvent::offsetGet() should either be compatible with IteratorAggregate::offsetGet(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

This happens for all methods defined in `IteratorAggregate`.

## The fix

We add proper return strict typing for the methods except for `\Drupal\og\Event\PermissionEvent::offsetGet()` because we cannot type-hint to `mixed` as this return type has been introduced only in PHP 8.0 but we still keep the module compatible with PHP 7.4. In that case we use `#[\ReturnTypeWillChange]`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ dist: bionic
 language: php
 
 php:
-  - 7.4
-  - 8.0
-  - 8.1
+  - '7.4'
+  - '8.0'
+  - nightly
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: php
 php:
   - 7.4
   - 8.0
+  - 8.1
 
 env:
   global:
@@ -18,6 +19,8 @@ env:
 matrix:
   exclude:
     - php: 7.4
+      env: TEST_SUITE=PHP_CodeSniffer
+    - php: 8.0
       env: TEST_SUITE=PHP_CodeSniffer
 
 mysql:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,8 +3,9 @@ dist: bionic
 language: php
 
 php:
-  - 7.4
-  - 8.0
+  - '7.4'
+  - '8.0'
```
MPParsley, line 7:
```suggestion
  - '8.0'
  - '8.1.0'
```

claudiu-cristea: @MPParsley I would prefer `nightly` because <8.1.3 have security vulnerabilities (PHP 8.1.3 was a security release). So, either `>=8.1.3` or `nightly`. I would say let's keep `nightly` until PHP >=8.1.3 is available

composer.json
```diff
@@ -26,12 +26,22 @@
...
         "php": "^7.4 || ^8",
+        "cweagans/composer-patches": "^1.7",
         "drupal/core": "^9.2"
     },
     "require-dev": {
         "drupal/coder": "^8.2",
         "phpunit/phpunit": "^7 || ^8",
         "slevomat/coding-standard": "~6.0"
     },
+     "extra": {
```
MPParsley, line 37: Looks like https://www.drupal.org/project/coder/issues/3250986 has been fixed by @klausi. Perhaps @klausi or @pfrenssen could create a new release for coder?

claudiu-cristea: Until then, I think, we can live with the patch

MPParsley (approved):

Issue #749, opened 2022-09-08
Update Role Change forms to be aware of group & member without using tempstore.private


dipakmdhrm:
## Problem/Motivation
Og has two forms (`OgAddMultipleRolesForm` & `OgRemoveMultipleRolesForm`) to add/remove group roles for a group memeber.

Both these form extend the `OgChangeMultipleRolesFormBase` class which stores the group & member data in `tempstore.private`.

This makes it a little complicated to alter these forms and do any kind of logic that needs to be aware of current group and the member for which the roles are being updated.

## Proposed resolution
There are two options:

1. Pass the group & member data in `$form_state`
2. Update the current paths for role add & remove from `/group/roles/add` & `/group/roles/remove` to `/group/{group}/roles/add/{member}` & `/group/{group}/roles/remove/{member}` respectively. This will allow developers to get the group & member data from path.

I'd actually recommend that we do both.

PR #750, opened 2022-10-13
Pass membership data in form_state for roles add/remove form.

dipakmdhrm:
#749

### DIFF
--- a/og.api.php
+++ b/og.api.php
@@ -59,7 +59,7 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata $cache
  * Implements hook_form_FORM_ID_alter().
  */
 function hook_form_og_membership_remove_multiple_roles_action_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
-  // Get access to current group and selected memberships when on role remove
+  // Get access to current group and selected memberships when we're on role remove
   // form.
   $memberships = $form_state->getTemporaryValue('selected_memberships');
   $group = $form_state->getTemporaryValue('current_group');
--- a/og.api.php
+++ b/og.api.php
@@ -59,12 +59,11 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata $cache
  * Implements hook_form_FORM_ID_alter().
  */
 function hook_form_og_membership_remove_multiple_roles_action_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
-  // Get access to current group and selected memberships when we're on role remove
-  // form.
+  // Get access to current group and selected memberships when we're on the role
+  // remove form.
   $memberships = $form_state->getTemporaryValue('selected_memberships');
-  $group = $form_state->getTemporaryValue('current_group');
 
-  $form['roles']['#options'] = array_filter($form['roles']['#options'], function ($key) use($memberships, $group) {
+  $form['roles']['#options'] = array_filter($form['roles']['#options'], function ($key) use($memberships) {
     // Code to filter out options based on data from memberships and group.
   }, ARRAY_FILTER_USE_KEY);
 }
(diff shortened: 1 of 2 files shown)


amitaibu:
I think a better solution is what you proposed in https://github.com/Gizra/og/issues/749

> Update the current paths for role add & remove from /group/roles/add & /group/roles/remove to /group/{group}/roles/add/{member} & /group/{group}/roles/remove/{member} respectively. This will allow developers to get the group & member data from path.

dipakmdhrm:
> I think a better solution is what you proposed in #749
> 
> > Update the current paths for role add & remove from /group/roles/add & /group/roles/remove to /group/{group}/roles/add/{member} & /group/{group}/roles/remove/{member} respectively. This will allow developers to get the group & member data from path.

Implementing this wouldn't be possible because of use of actions to add/remove roles. 

og-actions

This is a problem for our proposed implementation because of 2 reasons:
1. My proposed implementation (`/group/{group}/roles/add/{member}`) only considers addition/removal of roles for just 1 member at a time. However, actions allows user to select multiple members and bulk update roles for them.
2. Actions (example, publish node, set flag etc.) are supposed to be self contained and are not usually expected to pass data in a next step to some form. Actions are in no way connected to a new form on a new page (like our proposed form at /group/{group}/add/roles) and we need some way to pass the list of selected members to the form. That's why we need tempstore. 

I tried to find other modules where something similar was set up in hope I might find a way. But even in other modules (views_bulk_edit for example) tempstore is used.

That unfortunately leaves us with the fallback of passing membership data from form build to form alter for people to be able to use it.

In conclusion: I ended up spending way too much time only to end up with disappointment :sweat_smile:

amitaibu:
Having it in the form_state then would be fine :+1:

amitaibu:
@dipakmdhrm this needs more work

dipakmdhrm:
Added examples in `og.api.php`

og.api.php
```diff
@@ -54,6 +55,34 @@ function hook_og_user_access_alter(array &$permissions, CacheableMetadata $cache
   $cacheable_metadata->addCacheableDependency($config);
 }
 
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function hook_form_og_membership_remove_multiple_roles_action_form_alter(array &$form, FormStateInterface $form_state, string $form_id) {
+  // Get access to current group and selected memberships when on role remove
```
amitaibu:
```suggestion
  // Get access to current group and selected memberships when we're on role remove
```

src/Form/OgChangeMultipleRolesFormBase.php
```diff
@@ -93,6 +93,17 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#value' => $this->t('Submit'),
     ];
 
+    // Set selected memberships & group in form state. This can be used by
+    // developers to alter the roles form.
```
amitaibu:
```suggestion
    // Set selected memberships in the form state. This can be used by
    // to alter the roles form.
```

src/Form/OgChangeMultipleRolesFormBase.php
```diff
@@ -93,6 +93,17 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+    $memberships = $this->getMemberships();
+    if (empty($memberships)) {
+      return $form;
+    }
+    $membership = current($memberships);
+    $group = $membership->getGroup();
```
amitaibu: Can this be removed? We can extract the group from the Membership

amitaibu (commented):
Still WIP?

src/Form/OgChangeMultipleRolesFormBase.php
dipakmdhrm: Removed

amitaibu:
@dipakmdhrm can you please work on a PR to fix the Travis errors (against the `8.x-1.x` branch)

dipakmdhrm:
> @dipakmdhrm can you please work on a PR to fix the Travis errors (against the `8.x-1.x` branch)

WIP: https://github.com/Gizra/og/pull/763
@amitaibu

claudiu-cristea:

Issue #744, opened 2022-07-29
Routing error when viewing Membership Types overview page


cofihouse:
I'm getting a WSOD post install when visiting the membership types administration page. Here's the error:

`Symfony\Component\Routing\Exception\MissingMandatoryParametersException: Some mandatory parameters are missing ("membership_type") to generate a URL for route "entity.og_membership_type.auto_label". in Drupal\Core\Routing\UrlGenerator->doGenerate() (line 181 of C:\wamp64\www\list\core\lib\Drupal\Core\Routing\UrlGenerator.php).`

I can visit the edit/add pages if I input the direct route.

jludwig:
See: https://github.com/Gizra/og/pull/751

cofihouse:
@jludwig pull request fixes my issue

PR #751, opened 2022-10-13
Fix machine name mismatch in OgMembershipType link template vs route

jludwig:
The link template in OgMembershipType specifies {membership} as a placeholder, but the parameter name in og.routing.yml is {og_membership_type}. This cause issues when another module tries to add additional links, like in #744 and https://gitlab.com/drupalspoons/devel/-/issues/443

I prefer the more specific og_membership_type over membership_type, but an alternative fix would be to change the parameter names in og.routing.yml.

### DIFF
--- a/src/Entity/OgMembershipType.php
+++ b/src/Entity/OgMembershipType.php
@@ -46,8 +46,8 @@ use Drupal\og\OgMembershipTypeInterface;
  *     "description"
  *   },
  *   links = {
- *     "edit-form" = "/admin/structure/membership-types/manage/{membership_type}",
- *     "delete-form" = "/admin/structure/membership-types/manage/{membership_type}/delete",
+ *     "edit-form" = "/admin/structure/membership-types/manage/{og_membership_type}",
+ *     "delete-form" = "/admin/structure/membership-types/manage/{og_membership_type}/delete",
  *     "collection" = "/admin/structure/membership-types",
  *   }
  * )


cofihouse (approved):
Corrects the issue for me.

Lancelight:
This patch doesnt do anything to fix the problem.  As far as I can tell, all it does is modify a PHP comment?  That certainly wont affect how PHP reacts to anything unless comments are now executable (which I doubt).  It didnt solve the problem of Devel module complaining with the following:

Uncaught PHP Exception Symfony\\Component\\Routing\\Exception\\MissingMandatoryParametersException: "Some mandatory parameters are missing ("membership_type") to generate a URL for route "entity.og_membership_type.devel_load"." at core/lib/Drupal/Core/Routing/UrlGenerator.php line 181

jludwig:
@Lancelight

The problem is a mismatch between the plugin definition and the route. The plugin definition is defined PHP comment you saw. See annotations for an explanation of how this comment affects functionality: https://www.drupal.org/docs/drupal-apis/plugin-api/annotations-based-plugins

Edit:  Make sure to clear caches after applying this patch.

Lancelight:
I read that annotations link and from what I can gather, the benefit is less memory for the PHP compiler since compiler ignores comments.  Sneaky lol.  IDK if it actually makes any noticeable difference but in theory it sounds like it might.  :D

Anyway,
I actually did try to add the comments to the file (preserving the correct indentations),  but it didnt change the behavior nor change the error message.  I tried uninstall/reinstall of devel,  cleared caches,  nothing seemed to fix the error (other than uninstalling devel which isnt what I want to do) I was running into which feels eerily similar to the issue this patch is supposed to fix.  Is there another file that needs to be modified?  I did not upgrade OG at all,  just added this patch.  Moshe from devel module is the one that pointed me at this issue item.

This is what I have for og in composer.json

"type": "package",
            "package": {
                "name": "gizra/og",
                "version": "dev-8.x-1.x",
                "type": "drupal-module",
                "dist": {
                    "url": "https://github.com/Gizra/og/archive/8.x-1.x.zip",
                    "type": "zip"

MPParsley (approved):

PR #754, opened 2022-11-19
D10 compatibility

claudiu-cristea:
## Review notes

* I've added `::accessCheck()` to all entity queries (see https://www.drupal.org/node/3201242) without any argument, meaning the _access is checked_. I did this in order to preserve the current code behavior. However, I'm not sure that this is correct for all the cases. It's not in the scope of this PR to fix potential cases where the entity query should bypass the current user access constraint. The module maintainer should carefully review each case and fix potential cases in a followup.
* See the hack and the comment I've added in `OgRole::calculateDependencies()`. The module maintainer should weight on the `@todo` and, potentially, handle that in a followup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ matrix:
       env: TEST_SUITE=PHP_CodeSniffer
     - php: 8.0
       env: TEST_SUITE=10.0.x
+    - php: 8.1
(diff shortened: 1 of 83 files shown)


MPParsley (approved):
Thank you @claudiu-cristea!

Issue #758, opened 2023-04-06
OgMembershipAccessControlHandler has a fatal error when trying to use JSON:API to create memberships


pwolanin:
`\Drupal\og\OgMembershipAccessControlHandler::createAccess()` has a fatal error when it's called in any normal way via REST or JSON:API.  It only works from the OG UI when called as a route access check from `\Drupal\og\Access\OgMembershipAddAccessCheck::access()`

The problem resolves around the use of `$context['group']` to try to generate a cid when we probably don't even get much advantage from caching this access result (you'd have to be creating multiple in the same page load).

Should probably be resolved by not bothering to try to try to cache.  Also possibly get the current group from the OG Context if it's not passed in.  Finally,  seems like the general goal of this is that the group context and the group the membership is being created for should match - that might need some kind of additional validation constraitnt?

PR #759, opened 2023-04-07
Prevent fatal error in membership create access check

pwolanin:
fix for #758

### DIFF
--- a/src/OgMembershipAccessControlHandler.php
+++ b/src/OgMembershipAccessControlHandler.php
@@ -37,6 +37,13 @@ class OgMembershipAccessControlHandler extends EntityAccessControlHandler implem
   protected $membershipManager;
 
   /**
+   * Instance of the OgContext service.
+   *
+   * @var \Drupal\og\ContextProvider\OgContext
+   */
+  protected $ogContext;
+
+  /**
    * Constructs a OgMembershipAccessControllHandler object.
    *
    * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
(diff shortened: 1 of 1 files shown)


src/OgMembershipAccessControlHandler.php
```diff
@@ -103,16 +114,10 @@ public function createAccess($entity_bundle = NULL, AccountInterface $account =
...
-    $cid = 'create:' . $context['group']->getEntityTypeId() . ':' . $context['group']->id();
-    if ($entity_bundle) {
-      $cid .= ':' . $entity_bundle;
-    }
-
-    if (($access = $this->getCache($cid, 'create', $context['langcode'], $account)) !== NULL) {
-      // Cache hit, no work necessary.
-      return $return_as_object ? $access : $access->isAllowed();
+    if (empty($context['group'])) {
+      $context['group'] = $this->ogContext->getGroup();
```
amitaibu, line 120: I think that if we don't have `$this->ogContext->getGroup();` then we should return early.

btw, How are you passing the group info via RESTful?

pwolanin: @amitaibu in this case we are using JSON:API and the group context is populated via the query string.  

We have our own version we use instead, but basically very similar to `\Drupal\og\Plugin\OgGroupResolver\RequestQueryArgumentResolver`

Ours does not require a second plugin to confirm since we need to have group context for JSON:API requests of various kinds.

pwolanin: We make sure to order the plugins so if you already found a group context from e.g. a node or other entity in the web or JSON:API request you cannot change it.

amitaibu: Thanks. I still think we would need to add a check for an existing `$this->ogContext->getGroup();` as some faulty 3rd party `RequestQueryArgumentResolver` plugin might not pass the context correctly, and we'd end up without a group here.

Does it make sense, or am I'm missing your point?

pwolanin: I just ran into this problem again today. Ending up without a group is fine - it falls back to global permissions.

amitaibu (approved):
Looks good to me. @MPParsley can be merged?

amitaibu:
Thanks!

PR #764, opened 2023-12-22
[Drupal 10.2] Using a translatable string as a category for field type is deprecated

claudiu-cristea:
See https://www.drupal.org/node/3375748

### DIFF
--- a/og.field_type_categories.yml
+++ b/og.field_type_categories.yml
@@ -1,3 +1,3 @@
 og:
   label: Organic groups
-  description: Organic group fields
+  description: Organic groups fields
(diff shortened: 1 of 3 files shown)


joelpittet:
@claudiu-cristea Are there any negatives for pre 10.2 with this change? Looks good I just am not sure where that is used in 10.1 off hand

claudiu-cristea:
The only thing I can see is that the category will be rendered as lowercase and untranslated in the admin UI.

There's no way to fix this in a definitions alter hook because that acts after the depreciation error message has been broadcasted.

amitaibu:
@bboro ^^

claudiu-cristea:
Since https://www.drupal.org/project/og/issues/3467694 was merged into the `8.x-1.x` branch, the development has been moved to https://www.drupal.org/project/issues/og. If this change is still actual, please open an issue at https://www.drupal.org/project/issues/og and provide there a _merge request_ similar with this pull request.

